PR #900, opened 2020-03-13
Add GitHub labeler

se-bastiaan:
### Summary

Automatically label Pull Requests based on the changed files


se-bastiaan:
@joren485 No I need to do more changes for that and those are unrelated. I cannot change the name of the label with a PR that changes it where necessary (issue templates etc)

PR #1040, opened 2020-03-18
Checkout when setting up review environment

joren485:
### Summary
Checkout when setting up review environment.


se-bastiaan:
deploy review

joren485:
@se-bastiaan, because this workflow triggers on `issue_comment` events, it cannot be tested in the PR itself, but only on `master`.

se-bastiaan:
ah yes

Issue #1013, opened 2020-03-18
ProgrammingError: more than one row returned by a subquery used as an expression


thaliawww-service:
In GitLab by _thaliatechnicie on Mar 1, 2020, 16:38

Sentry Issue: [CONCREXIT-27](https://sentry.io/organizations/thalia/issues/1540717591/?referrer=gitlab_integration)

```
ProgrammingError: more than one row returned by a subquery used as an expression

  File "manage.py", line 19, in <module>
    execute_from_command_line(sys.argv)
  File "django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "django/core/management/base.py", line 328, in run_from_argv
    self.execute(*args, **cmd_options)
  File "django/core/management/base.py", line 369, in execute
    output = self.handle(*args, **options)
  File "members/management/commands/dataminimisation.py", line 17, in handle
    processed = services.execute_data_minimisation(options["dry-run"])
  File "members/services.py", line 210, in execute_data_minimisation
    for member in members:
  File "django/db/models/query.py", line 276, in __iter__
    self._fetch_all()
  File "django/db/models/query.py", line 1261, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "django/db/models/query.py", line 57, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "django/db/models/sql/compiler.py", line 1144, in execute_sql
    cursor.execute(sql, params)
  File "django/db/backends/utils.py", line 68, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "django/db/backends/utils.py", line 77, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
  File "django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)

CardinalityViolation: more than one row returned by a subquery used as an expression

  File "django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
```

se-bastiaan:
[...]

Issue #1011, opened 2020-03-18
Add wiki page about Renovate


thaliawww-service:
In GitLab by @joren485 on Feb 28, 2020, 12:10

### One-sentence description
Add wiki page about Renovate.

### Why?
It is important to have documentation about the third-party applications we use.

se-bastiaan:
Done: https://github.com/thaliawww/concrexit/wiki/Renovate

Issue #961, opened 2020-03-18
Make mailinglist deletion safer/better informed


thaliawww-service:
In GitLab by @pingiun on Nov 20, 2019, 13:12

### Motivation
Currently deleting a mailinglist will disable the group on gsuite, but delete the list on the website. This means that the name can be added as an alias for another list on the website, but gsuite will not support his, so we have a syncing error. 

We could do an actual deletion on gsuite, and preserve archives another way. Or make sure the name cannot be used anymore after deletion, except when recreating the same list. If this method is chosen, the user should be informed on deletion.

Issue #943, opened 2020-03-18
Putting society chair 'role' in a mailing list


thaliawww-service:
In GitLab by touwerkerk on Oct 25, 2019, 15:14

### One-sentence description

I would like a way to be able to put "chair of society X" in a mailing list.

### Motivation

Societies will get emails with stuff like "Hey, can I join?" or "Hi, I have some questions", but there are a bunch of members that do not need (or want) to get these. Right now there is a mailing list with just the chair of the Adventurers' Guild in it. However when the chair changes, we will need to manually change the person in this mailing list. It would be better if 'Chair of Society' could be selected instead of 'Thomas van Ouwerkerk'

### Desired functionality

We would like a way for just the chair to be in a mailing list and when a new chair is appointed, this person is automatically the one that gets emails from that point on.

thaliawww-service:
In GitLab by @gijshendriksen on Oct 25, 2019, 15:18

Maybe we could generate a mailinglist with the chair of each member group automatically? For instance, if the member group has mailinglist `adventurers`, we could generate the mailinglist `adventurers-chair`. And then we'd do this for all member groups that have selected a chair?

Another option is to add a checkbox to enable generation of this mailinglist for each member group, so every society/committee can decide for themselves whether they want to use it?

thaliawww-service:
In GitLab by @se-bastiaan on Oct 25, 2019, 15:31

Putting a role inside a mailinglist would be rather complex. Especially since you also really want that mailinglist with all the members as well.

I do like Gijs' first option. The checkbox won't be really necessary imo since member groups could just choose to not use the list and just leave it be. We just iterate over the member groups, if it has a chair and an existing mailing list we also add an automatic mailinglist with the same name + `-chair` for said chair.

thaliawww-service:
In GitLab by @joren485 on Oct 30, 2019, 10:26

I am against this change. If someone is part of a group (whether that is a committee or society), mail about that group is always relevant to them. Members should be aware of what is going on in their group. 

This change will also make it very easy for members to miss important information about the group they are in, because someone only emailed the chair.

thaliawww-service:
In GitLab by @se-bastiaan on Oct 30, 2019, 10:31

[...]

PR #1044, opened 2020-03-18
Add payment batches functionality

### LINKED ISSUE #929: Thalia Pay: Export payment batches for ING
In GitLab by @se-bastiaan on Oct 3, 2019, 16:49

Subtask of #632, partially blocked by #930.

### One-sentence description

Export payment batches for ING for #632.

### Motivation

Because we want Thalia Pay (#632).

### Desired functionality

Export option for payment batches in the right format. These batches can be used to execute a direct debit batch.

### Suggested implementation

- Need to determine the format with the treasurer.
- We can do this by adding extra payment actions to the ~payments admin that only exports payments marked for direct-debit (#930).
- The selection for which time the payments are exported can already be selected using Django's built-in date filter.
- The mandate 'last used' field should be updated upon export of the payments.

### LINKED ISSUE #930: Thalia Pay: Export financial information that can be imported to Conscribo
In GitLab by @se-bastiaan on Oct 3, 2019, 16:51

### One-sentence description

Export financial information that can be imported to Conscribo

### Motivation

Because we want to introduce Thalia Pay (#632)

### Desired functionality

Have an option to export financial payment information that can be imported in Conscribo. We currently have an export option but the use format cannot be used with Conscribo. If we add the format Conscribo wants the work to process direct debits (which might be a lot) will be decreased a lot.

### Suggested implementation

- Need to determine the format with the treasurer.
- We can do this by adding extra payment actions to the ~payments.

lukovdm:
Closes #929 and closes #930 

### Summary
This PR adds batches to payments. These batches can be exported and processed.

### How to test
1. Go to the payments admin
2. Use the features, such as:
   - Creating a batch from selected payments
   - Processing a batch
   - Exporting a batch
   - Modifying a batch
   - Creating a batch from last months payments


lukovdm:
Some comments from @JobDoesburg 

- Maybe 'NEW BATCH FROM PREVIOUS MONTH OF THALIA PAY PAYMENTS' should be changed in 'NEW BATCH FROM UNBATCHED THALIA PAY PAYMENTS'. That leaves more freedom to how often you might want to create batches. It is also more clear that payments that are already in another batch (but are in the previous month) are not changed. Ultimately we could implement a cronjob that runs that command every month
- The Batch admin should display the topics of payments as well. And maybe also additional fields, such as creation date and notes
- As already discussed, an export per topic must be included as well, in addition to the export per member.
- The order of the columns of payments in the batch admin could be improved (for example: topic, paid by, amount, creation date, processing date, notes, remove)
- It should be more clear to the user when there are TPAY-typed payments that are not yet in a batch. Maybe display an info message in the batch admin overview or something like that.
- In the batch overview, 'Start date' and 'end date' of a batch should be renamed to 'first payment in batch' and 'last payment in batch'
- In the batch overview, maybe also list the number of payments and total amount.
- When a batch is processed, the 'remove payment from batch' column should be removed.
- There is a problem: it is possible to delete payments when they are in a processed batch. This just works and changes the batch even though it is processed. We should prevent deleting payments as soon as they are in a batch. This is already in #1013
- I think we shouldn't allow deleting processed batches.

#943 must also be realized before we can use this.
Naturally #1011 is also blocking.
The 'send email to notify all members' functionality on processing the batch is still missing.
I think we should also solve #961 before we should release this.

lukovdm:
> Some comments from @JobDoesburg
[...]
>     * When a batch is processed, the 'remove payment from batch' column should be removed.

This seems to be a lot of work, creating a custom template kind of work. I don't think it is worth it for now. The button doesn't work when the batch is processed.

lukovdm:
Fully preventing deletion of payments in processed batches seems quite impossible, because for example, doing `queryset.delete()` fully bypasses any overridden delete methods in `payments`. I have prevented deletion of objects in the admin, including the action. Should I look more into preventing deletion of these objects altogether?

JobDoesburg:
@lukovdm 
>This seems to be a lot of work, creating a custom template kind of work. I don't think it is worth it for now. The button doesn't work when the batch is processed.

In that case it is totally fine by me! If I remember correctly the button did work when I was testing it (but maybe you changed it now, or I don't remember it correctly

JobDoesburg:
> Fully preventing deletion of payments in processed batches seems quite impossible, because for example, doing `queryset.delete()` fully bypasses any overridden delete methods in `payments`. I have prevented deletion of objects in the admin, including the action. Should I look more into preventing deletion of these objects altogether?

If it is not possible from the admin then I think it is okay

website/payments/admin_views.py
```diff
@@ -42,3 +50,102 @@ def post(self, request, *args, **kwargs):
...
+
+        if "next" in request.POST:
+            return redirect(request.POST["next"])
+
+        return redirect("admin:payments_batch_change", kwargs["pk"])
+
+
+@method_decorator(staff_member_required, name="dispatch")
+@method_decorator(
+    permission_required("payments.process_batches"), name="dispatch",
```
JobDoesburg: Maybe a bit nitpicky, but since we're exporting IBAN's maybe we should require additional permissions as well

website/payments/locale/nl/LC_MESSAGES/django.po
```diff
@@ -272,6 +332,10 @@ msgstr "Verwerk (Thalia Pay)"
 msgid "Process (wire payment)"
 msgstr "Verwerk (overboeking)"
 
+#: templates/admin/payments/change_list.html
+msgid "New batch from previous month of Thalia Pay payments"
+msgstr "Nieuwe batch van TPAY betalingen van vorige maand"
```
JobDoesburg: Maybe we shouldn't call it TPAY in the UI

website/payments/locale/nl/LC_MESSAGES/django.po
```diff
@@ -43,6 +43,10 @@ msgstr ""
 msgid "Process selected payments (wire)"
 msgstr "Verwerk geselecteerde betalingen (overboeking)"
 
+#: admin.py
+msgid "Add selected TPAY payments to a new batch"
+msgstr "Voeg geselecteerde TPAY betalingen aan een nieuwe batch toe"
```
JobDoesburg: Maybe we shouldn't call it TPAY in the UI

JobDoesburg:
image

Some missing translations

JobDoesburg:
Maybe we should change the `__str__` method of payments to something that includes the type, payer and topic, instead of only the amount

JobDoesburg:
Would it be hard to create a new kind of permissions for only displaying batches? (But not editing or doing anything with them)? That would be nice for the audit committee...

JobDoesburg:
I just tested what @lukovdm said earlier in this thread:

We still have the undesirable flow:
1. add payment for an event registration
2. add payment to batch
3. process batch
4. change (remove) the payment for the event registration
5. the batch total amount stays, but it doesn't correspond anymore with the payments in the batch, because the payment is still deleted.

This is something we must fix...

Maybe it is too much overkill, but what about introducing something like 'finalized' status for events? So people cannot change the payments and attendance anymore after it is finalized.

lukovdm:
> Maybe we should change the `__str__` method of payments to something that includes the type, payer and topic, instead of only the amount

I think this is something for another issue.

lukovdm:
> Would it be hard to create a new kind of permissions for only displaying batches? (But not editing or doing anything with them)? That would be nice for the audit committee...

Django has builtin view permissions for every model.

JobDoesburg:
> > Would it be hard to create a new kind of permissions for only displaying batches? (But not editing or doing anything with them)? That would be nice for the audit committee...
> 
> 
> 
> Django has builtin view permissions for every model.

Yeah that’s what I thought, but I couldn’t find them when I tried it...

lukovdm:
> > > Would it be hard to create a new kind of permissions for only displaying batches? (But not editing or doing anything with them)? That would be nice for the audit committee...
> > 
> > 
> > Django has builtin view permissions for every model.
> 
> Yeah that’s what I thought, but I couldn’t find them when I tried it...

image

se-bastiaan:
> > Some comments from @JobDoesburg
> > [...]
> > * When a batch is processed, the 'remove payment from batch' column should be removed.
> 
> This seems to be a lot of work, creating a custom template kind of work. I don't think it is worth it for now. The button doesn't work when the batch is processed.

You could just hide the form element in `BatchPaymentInlineAdminForm`:
```
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        instance = kwargs.get("instance")
        if instance and instance.batch.processed:
            # Hide the remove batch field when the batch is processed
            self.fields["remove_batch"].widget = self.fields[
                "remove_batch"
            ].hidden_widget()
```

website/payments/admin_views.py
se-bastiaan: There will only be one person doing this: the treasurer. So adding another permission for this would be an immense overkill.

se-bastiaan:
> Maybe we should change the `__str__` method of payments to something that includes the type, payer and topic, instead of only the amount

This is completely out of scope, you could create a new issue and add it to the [Thalia Pay project](https://github.com/svthalia/concrexit/projects/3).

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
 
     process_wire_selected.short_description = _("Process selected payments (wire)")
 
+    def add_to_new_batch(self, request: HttpRequest, queryset: QuerySet) -> None:
+        """Add selected TPAY payments to a new batch"""
+        tpays = queryset.filter(type=Payment.TPAY)
+        if len(tpays) > 0:
+            batch = Batch()
+            batch.save()
+            tpays.update(batch=batch)
```
se-bastiaan: This whole part is doing business logic in admin.py, this should happen in services.py instead

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
...
+                    request,
+                    f"The last batch {batch} is already processed",
+                    messages.ERROR,
+                )
+        else:
+            self.message_user(
+                request,
+                f"No payments using Thalia Pay are selected, " f"no batch is created",
+                messages.ERROR,
+            )
```
se-bastiaan: This whole part is doing business logic in admin.py, this should happen in services.py instead.
Check the `process_card_selected` function for an example, the result of your function could be the message and the status.

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
...
+
+    add_to_new_batch.short_descriptionn = _("Add selected TPAY payments to a new batch")
+
+    def add_to_last_batch(self, request: HttpRequest, queryset: QuerySet) -> None:
+        """Add selected TPAY payments to the last batch"""
+        tpays = queryset.filter(type=Payment.TPAY)
+        if len(tpays) > 0:
+            batch = Batch.objects.last()
+            if not batch:
+                self.message_user(request, "No batches available.", messages.ERROR)
```
se-bastiaan: Not translated.

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
...
+            elif not batch.processed:
+                batch.save()
+                tpays.update(batch=batch)
+                self.message_user(
+                    request,
+                    f"Successfully added {len(tpays)} payments to {batch}",
```
se-bastiaan: Not translated.

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
...
+                    messages.SUCCESS,
+                )
+            else:
+                self.message_user(
+                    request,
+                    f"The last batch {batch} is already processed",
```
se-bastiaan: Not translated.

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
...
+                    messages.ERROR,
+                )
+        else:
+            self.message_user(
+                request,
+                f"No payments using Thalia Pay are selected, " f"no batch is created",
```
se-bastiaan: Not translated.

website/payments/admin.py
```diff
@@ -190,6 +245,53 @@ def process_wire_selected(self, request: HttpRequest, queryset: QuerySet) -> Non
 
     process_wire_selected.short_description = _("Process selected payments (wire)")
 
+    def add_to_new_batch(self, request: HttpRequest, queryset: QuerySet) -> None:
+        """Add selected TPAY payments to a new batch"""
+        tpays = queryset.filter(type=Payment.TPAY)
+        if len(tpays) > 0:
+            batch = Batch()
+            batch.save()
```
se-bastiaan:
```suggestion
            batch = Batch.objects.create()
```

website/payments/admin.py
```diff
@@ -285,6 +387,126 @@ def queryset(self, request, queryset) -> QuerySet:
...
+        ]
+        return custom_urls + urls
+
+    def save_formset(self, request, form, formset, change):
+        instances = formset.save(commit=False)
+
+        for instance in instances:
+            if instance.batch is None or (
+                instance.batch and not instance.batch.processed
+            ):
```
se-bastiaan:
```suggestion
            if instance.batch and not instance.batch.processed:
```

why would you set `instance.batch = None` if it is already `None`?

website/payments/admin.py
```diff
@@ -285,6 +387,126 @@ def queryset(self, request, queryset) -> QuerySet:
...
+        request: HttpRequest,
+        object_id: str = None,
+        form_url: str = "",
+        extra_context: dict = None,
+    ) -> HttpResponse:
+        """
+        Renders the change formview
+        Only allow when the batch has not been processed yet
+        """
+        extra_context = extra_context or {}
```
se-bastiaan:
```suggestion
        extra_context: dict = {},
    ) -> HttpResponse:
        """
        Renders the change formview
        Only allow when the batch has not been processed yet
        """
```
If the default value is a dictionary you don't need this check anymore I believe

website/payments/admin.py
```diff
@@ -285,6 +387,126 @@ def queryset(self, request, queryset) -> QuerySet:
...
+        obj = None
+        processed = False
+        if object_id is not None and request.user.has_perm("payments.process_batches"):
+            obj = Batch.objects.get(id=object_id)
+            processed = obj.processed
+
+        extra_context["batch"] = obj
+        extra_context["processed"] = processed
```
se-bastiaan:
```suggestion
        obj = None
        if object_id is not None and request.user.has_perm("payments.process_batches"):
            obj = Batch.objects.get(id=object_id)

        extra_context["batch"] = obj
```
You don't need to add this processed, you already have the batch and can access it in the template like `batch.processed`.

website/payments/admin_views.py
```diff
@@ -42,3 +50,102 @@ def post(self, request, *args, **kwargs):
...
+            messages.error(
+                request, _("{} already processed.").format(model_ngettext(batch, 1))
+            )
+        else:
+            batch.processed = True
+            payments = batch.payments_set.select_related("paid_by")
+            for payment in payments:
+                ba = payment.paid_by.bank_accounts.last()
+                ba.last_used = timezone.now()
+                ba.save()
```
se-bastiaan:
```suggestion
                bank_account = payment.paid_by.bank_accounts.last()
                bank_account.last_used = timezone.now()
                bank_account.save()
```
You have space. https://www.freshconsulting.com/development-principle-1-choose-appropriate-variable-names/

website/payments/admin_views.py
```diff
@@ -42,3 +50,102 @@ def post(self, request, *args, **kwargs):
...
+            _("Description"),
+            _("Mandate date"),
+        ]
+        writer.writerow([capfirst(x) for x in headers])
+
+        member_rows = batch.payments_set.values("paid_by").annotate(total=Sum("amount"))
+
+        for row in member_rows:
+            member: Member = Member.objects.get(id=row["paid_by"])
+            bankaccount: BankAccount = member.bank_accounts.last()
```
se-bastiaan:
```suggestion
            member = Member.objects.get(id=row["paid_by"])
            bankaccount = member.bank_accounts.last()
```
This typing isn't really useful I believe

website/payments/admin_views.py
```diff
@@ -42,3 +50,102 @@ def post(self, request, *args, **kwargs):
...
+    permission_required("payments.process_batches"), name="dispatch",
+)
+class BatchNewFilledAdminView(View):
+    """
+    View that adds a new batch filled with all payments that where not already in a batch.
+    """
+
+    def get(self, request, *args, **kwargs):
+        batch = Batch()
+        batch.save()
```
se-bastiaan:
```suggestion
        batch = Batch.objects.create()
```

Wait where have I seen this before? -> https://github.com/svthalia/concrexit/pull/1044/files#diff-6f253c2bb5cbce719d7bbcf5f9843763R248

If you make it one (more generic) helper in `services.py` you can just mock that here. Makes your tests easier.

website/payments/forms.py
```diff
@@ -41,3 +41,26 @@ class Meta:
...
+class BatchPaymentInlineAdminForm(forms.ModelForm):
+    """
+    Custom admin form for Payments model
+    for the Batch inlines to add remove
+    from batch option
+    """
+
+    remove_batch = forms.BooleanField(
+        required=False, label=_("Remove payment from batch")
+    )
```
se-bastiaan:
```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        instance = kwargs.get("instance")
        if instance and instance.batch.processed:
            # Hide the remove batch field when the batch is processed
            self.fields["remove_batch"].widget = self.fields[
                "remove_batch"
            ].hidden_widget()

    remove_batch = forms.BooleanField(
        required=False, label=_("Remove payment from batch")
    )
```

se-bastiaan: to solve the form element from showing in unprocessed batches

website/payments/locale/nl/LC_MESSAGES/django.po
```diff
@@ -159,6 +167,34 @@ msgstr "%s succesvol verwerkt."
...
+msgid "Successfully processed {}."
+msgstr "{} is succesvol verwerkt."
+
+#: admin_views.py
+msgid "Account holder name"
+msgstr "Rekeninghouder"
+
+#: admin_views.py
+msgid "Mandate id"
+msgstr "Mandatennummer"
```
se-bastiaan: Kenmerk machtiging

website/payments/admin_views.py
```diff
@@ -42,3 +50,102 @@ def post(self, request, *args, **kwargs):
...
+        response = HttpResponse(content_type="text/csv")
+        response["Content-Disposition"] = 'attachment;filename="batch.csv"'
+        writer = csv.writer(response)
+        headers = [
+            _("Account holder name"),
+            _("IBAN"),
+            _("Mandate id"),
+            _("Amount"),
+            _("Description"),
+            _("Mandate date"),
```
se-bastiaan:
```suggestion
            _("Account holder"),
            _("IBAN"),
            _("Mandate Reference"),
            _("Amount"),
            _("Description"),
            _("Mandate Date"),
```

website/payments/locale/nl/LC_MESSAGES/django.po
se-bastiaan: These terms are already part of the codebase, they're just in larger pieces of text. They have been taken from official SEPA forms and I think it would be good to stay consistent.

website/payments/models.py
```diff
@@ -92,6 +116,74 @@ def __str__(self):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    processed = models.BooleanField(verbose_name=_("processing status"), default=False,)
+
+    processing_date = models.DateTimeField(
+        verbose_name=_("processing date"), blank=True, null=True,
+    )
+
+    description = models.TextField(
+        verbose_name=_("description of batch"), default=_default_batch_description,
```
se-bastiaan:
```suggestion
        verbose_name=_("description"), default=_default_batch_description,
```
If this is a field called 'description' in a batch then it's pretty logical right? Just needs the translation.

website/payments/templates/admin/payments/change_form.html
```diff
@@ -11,6 +11,18 @@
     <script type="text/javascript" src="{% static 'admin/payments/js/payments.js' %}"></script>
 {% endblock %}
 
+{% block after_field_sets %}
+    {% if batch %}
+    <div class="submit-row payments-row">
+        {% if not processed %}
```
se-bastiaan:
```suggestion
        {% if not batch.processed %}
```
I mentioned this value in the view class that uses this template.

website/payments/models.py
```diff
@@ -60,22 +64,42 @@ class Payment(models.Model):
...
+            self.batch = self._batch
+
+        self._batch = self.batch
+
         super().save(force_insert, force_update, using, update_fields)
 
+    def clean(self):
+        if self.type != self.TPAY and self.batch is not None:
+            raise ValidationError(
+                {"batch": "Non Thalia Pay payments cannot be added to a batch."}
```
se-bastiaan: Not translated.

website/payments/models.py
```diff
@@ -92,6 +116,74 @@ def __str__(self):
...
+        return sum([payment.amount for payment in self.payments_set.all()])
+
+    total_amount.admin_order_field = "total amount"
+    total_amount.short_description = _("total amount")
+
+    def payments_count(self) -> Decimal:
+        return self.payments_set.all().count()
+
+    payments_count.admin_order_field = "payments count"
+    payments_count.short_description = _("payments count")
```
se-bastiaan: These functions could use @property

website/thaliawebsite/settings/settings.py
```diff
@@ -310,7 +310,7 @@
 )
 
 # Payment settings
-THALIA_PAY_ENABLED_PAYMENT_METHOD = False
+THALIA_PAY_ENABLED_PAYMENT_METHOD = True
```
se-bastiaan: Don't forget to change this to `False`. If you want to make sure this does not happen you should create a `localsettings.py` in `thaliawebsite/settings` with `THALIA_PAY_ENABLED_PAYMENT_METHOD = True`. That will override it locally and will not be committed since that file is a part of the gitignore.

website/payments/models.py
lukovdm: When they are a property, you can't give a short_description as far as I could find.

website/payments/admin.py
lukovdm: Should I then add them to services? They are quite short functions, so it seems weird.

website/payments/admin.py
lukovdm: Same here, it is not really possible to put the messages in services and the only common thing in these functions is `tpays.update(batch=batch)` as far as I can see.

website/payments/admin.py
lukovdm: Python does some weird stuff with default values: [https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

website/payments/admin_views.py
lukovdm: Fair enough, ignore my response on th other occurence.

lukovdm: I tried doing this, but I needed basically all inside information of the function and thus had to return all the pieces that were the function, and for such a short function it was overly complicated.

lukovdm:
If somebody could help with getting the preventing deletes working, it is the only thing missing. The current attempt sort of works in the admin (It gives an ugly error but no delete), but the tests still fail.

JobDoesburg:
I think overriding the `delete()` method and the `delete` action in the admin is sufficient. Or maybe a pre_delete or a post_delete signal (that restores the payment that was deleted) is an idea? What do you think?

website/payments/admin.py
se-bastiaan: Yes they are short, but it would make it much easier to test these parts of the code.

website/payments/admin.py
se-bastiaan: The function would be all the if and else statements with the messages as string and level outputs. The output can then be used here to show a message based on what happened

se-bastiaan:
I think using signals is harder to understand, if this works and you can test it I think it should be fine.

se-bastiaan (dismissed):

JobDoesburg:
image
Something is wrong here with the order of the columns (probably because the delete field is removed but the column for that is still there)

website/payments/templates/admin/payments/change_list.html
```diff
@@ -0,0 +1,11 @@
+{% extends "admin/change_list.html" %}
+{% load i18n admin_urls static compress %}
+
+{% block object-tools-items %}
+    {% if opts.model_name == "payment" or opts.model_name == "batch" %}
+    <li>
+        <a href="{% url 'admin:payments_batch_new_batch_filled' %}">{% trans "New batch from unbatched Thalia Pay payments" %}</a>
```
JobDoesburg, line 7: Maybe only display this button if there are unbatched Thalia Pay payments

se-bastiaan: I'm closing this because it is not really necessary and a lot of work

JobDoesburg:
deploy review

JobDoesburg:
remove review

JobDoesburg:
Too bad I couldn't test it properly because the `TPAY` setting wasn't turned on... I'll review it locally

JobDoesburg (approved):

PR #1054, opened 2020-03-25
Create payment for registrations and renewals

### LINKED ISSUE #1002: Use create_payment everywhere
In GitLab by @JobDoesburg on Feb 12, 2020, 20:37

### One-sentence description

Use create_payment everywhere where we use payments

### Why?

Prevent code duplication

### Current implementation

Shitty, chaotic, anarchy, tears

### Suggested implementation

Make models Payable (implement the interface / abstract class from #1010) and use `create_payment` as the only way to interact with / create payments.

JobDoesburg:
Closes #1002

### Summary

Registrations and renewals use the create_payment function now. And this PR removes all functionality of the payments app that marks an existing payment as paid.

### How to test

1. Create a registration as user
2. Confirm the mail
3. Accept the registration
4. Mark as paid
5. Check that the user is created with the correct membership

---

1. Create a renewal as user
3. Accept the renewal
4. Mark as paid
5. Check that the user is updated with the correct membership


JobDoesburg:
Todo:

[ ] Make the payment widget use payables and make it work in general
[ ] Remove payment.processed
[ ] Remove payment type None
[ ] Add migration for contribution field
[ ] Tests

website/registrations/models.py
```diff
@@ -140,9 +137,25 @@ class Meta:
         )
 
 
-class Registration(Entry):
+class Registration(Entry, Payable):
```
se-bastiaan: The actual Payable object here would be the `Entry` class. Then you can override any other fields like you are doing in `Renewal` already.

website/registrations/models.py
```diff
@@ -140,9 +137,25 @@ class Meta:
...
+    def payment_amount(self):
+        return self.contribution
+
+    @property
+    def payment_topic(self):
+        return f"Membership registration {self.membership_type} ({self.length})"
+
+    @property
+    def payment_notes(self):
+        return f"Created at {self.created_at}. Confirmed at {self.updated_at}."
```
se-bastiaan: The notes is the member-facing information about the payment. And it should be properly human readable. This can contain the same information as is in the topic, which is only used for the treasurer imo. That's how I implemented it for everything when adding the field. Like for events: https://github.com/svthalia/concrexit/blob/7255f42587745986bec46921a7122760879606ce/website/events/models.py#L650

website/registrations/models.py
JobDoesburg: Shouldn't Entry be an ABC?

se-bastiaan: No, both because it is a Django model and that is impossible. And it is already an abstract base model.

JobDoesburg:
Additional todo:

- [ ] Add possibility to sign a mandate when registering
- [ ] Merge the 'pay with Thalia Pay' code in here

website/registrations/migrations/0022_auto_20200325_2045.py
```diff
@@ -0,0 +1,28 @@
...
+        ('registrations', '0021_auto_20191106_2338'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='entry',
+            name='contribution',
+            field=models.FloatField(validators=[django.core.validators.MinValueValidator(7.5)], verbose_name='contribution'),
+        ),
+        migrations.RunPython(set_contribution_fields)
```
se-bastiaan, line 27: This should be split in a separate migration, running python with a database change has not worked in the past because it is no longer atomic.

JobDoesburg:
Nice work! I think it looks good, but I can't review this as I opened the PR. Anyway I think it would be good if a third person looks at it as well

website/events/admin.py
```diff
@@ -338,16 +340,25 @@ def has_delete_permission(self, request, registration=None):
             return False
         return super().has_delete_permission(request, registration)
 
+    def get_form(self, request, obj=None, **kwargs):
+        kwargs["formfield_callback"] = partial(
+            self.formfield_for_dbfield, request=request, obj=obj
+        )
+        return super().get_form(request, obj, **kwargs)
+
```
pingiun, line 348:
```suggestion
    def get_form(self, request, obj=None, **kwargs):
        return super().get_form(
            request, 
            obj, 
            formfield_callback=partial(self.formfield_for_dbfield, request=request, obj=obj), 
            **kwargs)

```

website/events/admin.py
```diff
@@ -338,16 +340,25 @@ def has_delete_permission(self, request, registration=None):
...
     def formfield_for_dbfield(self, db_field, request, **kwargs):
         """Customise the formfields of event and member"""
+        obj = kwargs.pop("obj", None)
```
pingiun, line 351:
```suggestion
    def formfield_for_dbfield(self, db_field, request, obj=None, **kwargs):
        """Customise the formfields of event and member"""
```

website/payments/admin_views.py
```diff
@@ -18,27 +20,41 @@
...
+        ):
+            raise SuspiciousOperation("Missing POST parameters")
+
+        if "next" in request.POST and not url_has_allowed_host_and_scheme(
+            request.POST.get("next"), allowed_hosts={request.get_host()}
+        ):
+            raise DisallowedRedirect
+
+        app_label = kwargs["app_label"]
+        model_name = kwargs["model_name"]
```
pingiun, line 39: Same thing as previous suggestions, why are these dict accesses instead of actual keyword arguments?

website/payments/widgets.py
```diff
@@ -11,10 +11,18 @@ class PaymentWidget(Widget):
     """
 
     template_name = "payments/widgets/payment.html"
+    obj = None
```
pingiun, line 14: This creates a class variable instead of an object variable, this line can be removed

website/registrations/admin.py
```diff
@@ -113,10 +115,19 @@ def reference_count(self, obj):
 
     reference_count.short_description = _("references")
 
+    def get_form(self, request, obj=None, **kwargs):
+        kwargs["formfield_callback"] = partial(
```
pingiun, line 119: Same thing as above, make this an actual keyword argument

website/registrations/admin.py
```diff
@@ -113,10 +115,19 @@ def reference_count(self, obj):
...
+            self.formfield_for_dbfield, request=request, obj=obj
+        )
+        return super().get_form(request, obj, **kwargs)
+
     def formfield_for_dbfield(self, db_field, request, **kwargs):
+        obj = kwargs.pop("obj", None)
```
pingiun, line 125: Same thing as above, make this an actual keyword argument

pingiun (changes requested):
I didn't get to check all the testing code but I care less about the code quality in there anyway. Tested by running a webserver locally

pingiun:
Translations can be removed as #1200 is merged

website/thaliawebsite/settings/settings.py
```diff
@@ -245,6 +245,9 @@
     "study": 30,
 }
 
+# Window during which a payment can be deleted again
+PAYMENT_CHANGE_WINDOW = os.environ.get("PAYMENTS_CHANGE_WINDOW", 10 * 60)
```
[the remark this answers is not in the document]
se-bastiaan:
```suggestion
PAYMENT_CHANGE_WINDOW = int(os.environ.get("PAYMENTS_CHANGE_WINDOW", 10 * 60))
```

pingiun (approved):

PR #1059, opened 2020-03-28
Refactor and fix photos app uploading code

### LINKED ISSUE #1017: Original images in photo albums are saved and downloaded
In GitLab by @gijshendriksen on Mar 11, 2020, 18:28

### One-sentence description

When downloading a photo album, both the original and downscaled images are downloaded, even though we're not supposed to save the original image.

1. Download a photo album
2. Notice the original images are saved and included in the album (`*.JPG`), alongside the downscaled images (`*.jpg`)

### Expected behaviour

1. Download a photo album
2. Only the downscaled images are saved and included

se-bastiaan:
Closes #1017

### Summary
I was going to work on #1017 and I found that I would rather refactor this piece of code first before making any changes. Moving logic and classes to better places within this codebase.

### How to test
1. Create a new album using a TAR file
2. Create a new album using a ZIP file
3. Create a single new photo
4. Any original photo will not survive the resizing process, test this by uploading a ZIP with PNG files because it is possible that your filesystem won't differentiate between .JPG and .jpg cases.


website/photos/services.py
```diff
@@ -1,8 +1,22 @@
...
 from PIL.JpegImagePlugin import JpegImageFile
-from PIL import ExifTags
+from PIL import ExifTags, Image, UnidentifiedImageError
+
+from photos.models import Photo
+
+logger = logging.getLogger(__name__)
 
 
 def photo_determine_rotation(pil_image):
```
lukovdm, line 22: When reading the exif docs, the image could also be flipped. Those would be the even cases, I don't know if we want to support this tho.

website/photos/services.py
```diff
@@ -70,3 +84,101 @@ def get_annotated_accessible_albums(request, albums):
...
+        messages.add_message(
+            request, messages.WARNING, _("Ignoring {}").format(photo_filename)
+        )
+        if photo_obj.file.path:
+            os.remove(photo_obj.file.path)
+        photo_obj.delete()
+
+
+def save_photo(photo_obj):
+    if photo_obj.original_file != photo_obj.file.path:
```
lukovdm, line 147: `original_file` is only defined on row 177, however, I can't find any reason for this check being there. This would stop the file from being called two times on the same photo_obj in the same request. However, that can't happen as it is only called in `extract_photo` or in `save_model`, both of which use a new Photo object every time they call `save_photo`. Thus, why is this check here?

website/photos/services.py
se-bastiaan: I think that this was an explicit choice, but since I'm not changing the functionality of the code here I do not think it is really relevant.

And when will we have flipped photos? Probably never.

website/photos/services.py
se-bastiaan: I think because it was here: https://github.com/svthalia/concrexit/pull/1059/files/c80d975b0d0208971969a054f5252ea2d850424f#diff-55499884164694c3509b150ab9e9b6a6L65

se-bastiaan: The check exists so that we only try to update the photos that are actually changed, you can save the photo without changing the file and this prevents it from saving. However, I do think that think check does not belong here since we are already saving the photo, and it is only relevant when updating a Photo object. So it should probably move to the save_model of the PhotoAdmin

se-bastiaan: Moved it to the right place.

website/photos/services.py
```diff
@@ -70,3 +84,99 @@ def get_annotated_accessible_albums(request, albums):
         )
 
     return albums
+
+
+def extract_archive(request, album, archive):
```
pingiun, line 89: Wouldn't it be easier to extract the file into a directory and then iterate through that directory? Seems like it may be less error prone. Also I would suggest using the `mimetypes` module to check if the file is a zip/tar file.

This method would also remove rechecking for zipfile/tarfile below

website/photos/services.py
```diff
@@ -70,3 +84,99 @@ def get_annotated_accessible_albums(request, albums):
...
+
+    if iszipfile:
+        with ZipFile(archive) as zip_file:
+            for photo in sorted(zip_file.infolist(), key=lambda x: x.filename):
+                extract_photo(request, zip_file, photo, album)
+    else:
+        # is_tarfile only supports filenames, so we cannot use that
+        try:
+            with tarfile.open(fileobj=archive) as tar_file:
+                for photo in sorted(tar_file.getmembers(), key=lambda x: x.name):
```
pingiun, line 101: I believe this is very inefficient for tar files because they don't have a directory like zip files do. In zip files you can read the file listing from the start of the file but for tar.{gz,bzip} you have to decompress and scan through the file.

se-bastiaan: Probably true, but based on the fact that this is only moving around code that already existed (https://github.com/svthalia/concrexit/pull/1059/files/ff339ac8d255ead53fef8dd406f29d58b5b8eccf#diff-a75c0fdbb5737477505a8678ac15706fL126) I think optimalisations to the code can be moved to a new issue

website/photos/services.py
se-bastiaan: Probably true, but same story as the other comment: just moving around. I really bumped into problems that the way this worked was all getting into quite large files, so that is why I refactored this into services.py before fixing the bug that I was actually working on.

I'd love to improve this code, but it's not getting any worse than what we had before. And since this has been open for 2 weeks I'd like to move those optimalisations to a new issue

PR #1049, opened 2020-03-19
Styled backend

JobDoesburg:
Closes #744 

### Summary
Added stylesheets for Django Admin to follow thaliawebsite styling

### How to test
Open the admin

PR #1062, opened 2020-03-29
Check for untranslated messages in CI

se-bastiaan:
Closes #1023 and closes #1033.

### Summary
This is the code that Jelle changed to check for untranslated strings in our messages. The MR was never merged in GitLab so I'm reopening here after fixing all the missing and fuzzy translations. Since I started touching single pages to do it correctly this closes some issues as well.

### How to test
Steps to test the changes you made:
1. Tests should pass.

PR #1065, opened 2020-03-30
Make large admin changelists better readable

JobDoesburg:
### Summary
Some often used admin change lists are a mess to navigate because of the numerous fields. By making some collapsable, they become easier to read.

### How to test
- Try to edit or create an event
- Try to edit a user

(the inline's are currently inconsistently uppercase, that will be fixed with #1049, I thought it was overkill to already do that in this PR)


se-bastiaan:
deploy review

website/events/admin.py
```diff
@@ -127,6 +106,41 @@ class EventAdmin(DoNextTranslatedModelAdmin):
     prepopulated_fields = {"map_location": ("location",)}
     filter_horizontal = ("documents",)
 
+    fieldsets = (
+        (_("General"), {"fields": ("title", "published", "organiser",)}),
```
se-bastiaan: I believe you're missing translations, but if you run those you will also have to make these changes probably: https://github.com/svthalia/concrexit/pull/1062/files#diff-830b743ee90a23a252b33004a915a561

se-bastiaan:
This is awesome, looked it up and did not know this was a documented feature. What do you mean by `inconsistently uppercase` exactly? The names of the fieldsets?

JobDoesburg:
> This is awesome, looked it up and did not know this was a documented feature. What do you mean by `inconsistently uppercase` exactly? The names of the fieldsets?

image

se-bastiaan:
Isn't that because you just have to uppercase the names of the fieldsets? https://github.com/svthalia/concrexit/pull/1065/files#diff-08e076f4891804b1623edce1dcacdcfeR169

JobDoesburg:
> Isn't that because you just have to uppercase the names of the fieldsets? https://github.com/svthalia/concrexit/pull/1065/files#diff-08e076f4891804b1623edce1dcacdcfeR169

Yeah either that, or make the inlines no-uppercase. Thing is that Django CSS states that titles of inlines should be uppercase (they have that defined explicitly) but for standard field sets that is not the case.

se-bastiaan:
Ah it's CSS, in that case it's definitely #1049 indeed. Otherwise it would become an upstream change imo, I don't care enough about this to give it any priority at all but if it can be fixed when overriding their styles it's fine by me.

JobDoesburg:
I can't resolve the conflicts it says...

JobDoesburg:
Never mind I can

I don't know enough about Git to say what's wise to do with regard to the order in which to merge things. I just cherry-picked that commit of yours on this branch

se-bastiaan:
The better solution would have been to rebase  on my branch, instead of cherry picking it. Now it cannot be merged even after my PR is merged because the history is incorrect.

PR #1077, opened 2020-04-09
grep the last commit hash

joren485:
### Summary
Currently, the review environment configuration takes the first commit hash of a pull request. However, the last commit hash should be used and when there are multiple commits, the first and the last are of course not the same.

This PR fixes that by taking the last commit.


joren485:
This cannot be tested properly (I tested it manually), because the config of `issue_comment` events only works when it is in the default branch.

PR #1093, opened 2020-05-03
Set SECURE_CONTENT_TYPE_NOSNIFF to False

joren485:
### Summary
In production we set the HTTP security headers (e.g. `X-Content-Type-Options`). Django could also set some of those headers. Since Django 3, setting `X-Content-Type-Options` has become [default in Django](https://docs.djangoproject.com/en/dev/ref/settings/#secure-content-type-nosniff).

This means that currently, both Nginx and Django are setting the `X-Content-Type-Options` header:
```
$ curl --silent -I https://thalia.nu/ | grep -i "x-content-type-options"
x-content-type-options: nosniff
x-content-type-options: nosniff
```

This PR disables setting the  `X-Content-Type-Options` header in Django.

### How to test
Look at the output of `curl --silent -I https://thalia.nu/ | grep -i "x-content-type-options"`.

### DIFF
--- a/website/thaliawebsite/settings/production.py
+++ b/website/thaliawebsite/settings/production.py
@@ -76,6 +76,8 @@ if os.environ.get("DJANGO_SSLONLY"):
     SESSION_COOKIE_SECURE = True
     CSRF_COOKIE_SECURE = True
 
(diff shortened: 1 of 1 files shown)


joren485:
deploy review

joren485:
remove review

Issue #972, opened 2020-03-18
Remove obsolete permissions


thaliawww-service:
In GitLab by @se-bastiaan on Dec 4, 2019, 21:29

### One-sentence description

Remove obsolete permissions that are left from removed things

### Why?

They're obsolete!

### Current implementation

We have the sentry backend, access to the board wiki and soon the nextcloud admin permissions.

### Suggested implementation

Should be removed.

PR #1109, opened 2020-05-13
Remove activemembers Nextcloud permissions and API

se-bastiaan:
Mentioning #972

### Summary

### How to test
You cannot test this, except for checking that the server still start.


website/members/migrations/0038_auto_20200513_2143.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 3.0.2 on 2020-05-13 19:43
```
pingiun, line 1: Where is this migration coming from?

se-bastiaan: https://github.com/svthalia/concrexit/pull/1109/files?file-filters%5B%5D=.mo&file-filters%5B%5D=.py#diff-612fa3582bd5e7670c6bbddc784428eaL97

se-bastiaan: removal of the permission from the member model

PR #1172, opened 2020-08-02
Remove dependency of pizzas app on Payment.NONE in API

se-bastiaan:
### Summary

To be able to completely remove Payment.NONE in the future the pizzas API should not rely on this constant any more. Furthermore,  the processed value is no longer used because the 

### How to test
1. Test that the pizzas orders in the backend still work
2. Test that the values in the user-facing pizza pages are still correct


se-bastiaan:
I took your original diff and specifically made some changes to make sure that would not happen :)

PR #1185, opened 2020-08-15
Change text of registrations payment email

se-bastiaan:
Closes #1184 

### Summary

Change registration payment email text to contain the wire transfer option.

### How to test
Steps to test the changes you made:
1. Register for Thalia on the review environment
2. Accept the registration you made
3. Check that the email contents are correct

PR #1186, opened 2020-08-15
Change text of registrations payment email

se-bastiaan:
See https://github.com/svthalia/concrexit/pull/1185


JobDoesburg:
Has the candiboard been involved in these texts? I don't know what they're planning on with corona next year at all regarding use of the board room and merc in general

pingiun:
Nick told me that he has discussed this with Kimberley

pingiun:
I got a single question in the mail about what should be entered in the "Name" field for a bank transfer. I think that can be added easily, but it is probably not a priority as I believe the name is not checked anywhere in the banking system

se-bastiaan:
This has already been deployed to production. Nick made the changes to the texts and I committed those.
If the 30th wants to make other changes we can always do that later.

PR #1194, opened 2020-08-26
Remove need of proxy server from review environment

joren485:
### Description
Currently, the review environment sets up EC2 instances in a private subnet. A proxy server redirects all HTTP(s) traffic to `x.public.review.technicie.nl` to the correct instance. Although, this works great, it does have a few points that can be improved:
- We need to maintain (and pay for) a proxy server running 24/7.
- The config of the proxy server needs to be automated using Ansible (harder than it sounds, because of some specific networking needs).
- DNS records are created, but never removed (if someone doesn't use the **remove review** command).
- Two days of uptime is too long in my opinion.

In this PR I address these issues with the following changes:
- [ ] Login to GitHub Packages to download the correct Docker image. This needs to happen regardless of any other changes, because of #1174.
- [ ] Setup EC2 instances in a public subnet (with a public ip)
- [ ] Use the public EC2 subdomain instead of setting up our own DNS records.
- [ ] Use self-signed certificates for HTTPs.
- [ ] Add the (amazing) [nginx-proxy](https://github.com/nginx-proxy/nginx-proxy) Docker image to handle HTTPs.
- [ ] Reduce the shutdown timer to 1 day.
- [ ] Remove the **remove review** command, because the server shuts everything down automatically.
- [ ] Setup all instance settings in `review-host-create.sh` (i.e. replace the EC2 launch template),  to make it clear with what settings the instance is started.

These changes might feel like removing some of the nicer features of the review environment (e.g. a `x.technicie.nl` domain and valid certificates). That is why I added the `request-for-comments` label. We can always decide to setup specific domains and/or add [certificates](https://github.com/nginx-proxy/docker-letsencrypt-nginx-proxy-companion).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,29 +1,31 @@
+---
+
 version: '3.5'
 
 services:
-    postgres:
-        image: postgres:11.1
-        volumes:
-            - /var/lib/postgresql/
-        environment: &postgresvars
-            POSTGRES_DB: thalia
-    web:
-        image: docker.pkg.github.com/svthalia/concrexit/local
-        build: .
-        command: runserver 0.0.0.0:8000
+    nginx-proxy:
+        image: jwilder/nginx-proxy
         ports:
-            - 8000:8000
-        depends_on:
-            - postgres
+            - 80:80
+            - 443:443
+        environment:
+            DHPARAM_GENERATION: "false"
+            DHPARAM_BITS: 2048
         volumes:
-            - ./website:/usr/src/app/website/
-            - concrexit:/concrexit/
-        environment: &webvars
-            <<: *postgresvars
-            DJANGO_DEBUG: 'True'
-            DJANGO_POSTGRES_HOST: postgres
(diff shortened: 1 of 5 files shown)


se-bastiaan:
- I noticed there is also a Docker that works with nginx-proxy when working on crash-and-compile. That would work for certificates maybe? I also noticed it [does not allow you to use the AWS domains](https://community.letsencrypt.org/t/policy-forbids-issuing-for-name-on-amazon-ec2-domain/12692) though. So that would need DNS again. And I don't think valid HTTPS is terribly important on review environments.
- I think the automatic shutdown can even be set to something like 6 hours.

se-bastiaan:
Will we still need this PR?

joren485:
I am still thinking about this and what a good solution to the underlying problems is. You can close this if you want (and I might reopen it later), but I am still working on this.

joren485:
See https://github.com/svthalia/serverconfig/issues/353#issuecomment-712761232

PR #1197, opened 2020-08-31
Fix the overlapping memberships

se-bastiaan:
### Summary
This PR fixes a problem with overlapping (membergroup) memberships.

### How to test
1. Go to a user
2. Add two memberships that have the end and start date on the same day
3. Save
4. Get no errors

### DIFF
--- a/website/utils/snippets.py
+++ b/website/utils/snippets.py
@@ -3,11 +3,16 @@ import datetime
 import hmac
 from _sha1 import sha1
 from base64 import urlsafe_b64decode, urlsafe_b64encode
+from collections import namedtuple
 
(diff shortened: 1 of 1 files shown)


website/utils/snippets.py
```diff
@@ -118,6 +123,45 @@ def overlaps(check, others, can_equal=False):
 
     The can_equal argument is used for boards, where the end date can't be the same
     as the start date.
+
+    >>> overlaps( \
```
se-bastiaan, line 127: These are [doctests](https://docs.python.org/3/library/doctest.html) :)

pingiun (approved):

PR #1202, opened 2020-09-02
Remove Dutch from settings.LANUAGES

### LINKED ISSUE #1200: Remove Dutch translation from the website
### Describe the change
The new policy of the board is that the website only has to support English, we should remove all the Dutch translations

To do:
- [x] Remove Dutch from settings.LANGUAGES (#1202)
- [x] Remove po and mo files (#1202)
- [ ] Remove mutilingualfield util functions and class

Remove multilingual field and the references to _en attributes from the following apps:
- [x] activemembers (#1216)
- [x] announcements (#1217)
- [x] documents (#1272, #1296)
- [x] education (#1273)
- [x] events
- [x] mailinglists
- [x] members
- [ ] merchandise?
- [ ] newsletters
- [x] partners
- [x] pizzas
- [ ] push notifications?
- [x] registrations
- [x] singlepages?
- [x] thabloid?

pingiun:
Previously I created #1200 to make Dutch fields optional. This PR goes a step further and removes all Dutch content from the website.

## How to test

Basically all pages should be checked for left over languages assumptions or Dutch content

### TODO
- [x] Fix tests referring to Dutch fields

Not all of this was manual work luckily, I used the following ugly oneliner:
```bash
find website -name '*.py' | rg --invert-match 'migrations' | xargs -n 1 -I{} bash -c 'rg --invert-match _nl {} > {}.new'
```


se-bastiaan:
Even though I understand that it may be easy to change the Dutch fields to be optional I do not see why we cannot just remove the complete Dutch functionality at once. I'm not really a fan of a half translated website just because we want to keep some Dutch explanations that are not required.

Technically we only need to keep the legal documents in Dutch but I'm not sure if that is connected to the translations in django.

se-bastiaan (changes requested):
Even though I understand that it may be easy to change the Dutch fields to be optional I do not see why we cannot just remove the complete Dutch functionality at once. I'm not really a fan of a half translated website just because we want to keep some Dutch explanations that are not required.

Technically we only need to keep the legal documents in Dutch but I'm not sure if that is connected to the translations in django.

.github/workflows/test.yaml
```diff
@@ -8,6 +8,8 @@ jobs:
     name: Codestyle and translation checks
     runs-on: ubuntu-18.04
     container: thalia/concrexit-dependencies
+    # Skip translations checks while there are no translations
+    if: false
```
se-bastiaan, line 12: If you disable this then you could also remove _only_ the translation checks

se-bastiaan:
deploy review

se-bastiaan:
deploy remove

se-bastiaan:
remove review

se-bastiaan:
deploy review

se-bastiaan:
remove review

se-bastiaan (changes requested):
It would be nice if the permissions of manage.py would stay on 0755. Otherwise `./manage.py` commands do not work anymore. Same for all the other files (100755 → 100644 in change list), unless this is required somehow.

website/documents/forms.py
```diff
@@ -21,7 +21,6 @@ class DocumentFileInput(widgets.ClearableFileInput):
     def get_context(self, name, value, attrs):
         context = super().get_context(name, value, attrs)
         if hasattr(value, "url"):
-            doc = models.Document.objects.get(Q(file_en=value) | Q(file_nl=value))
```
se-bastiaan, line 24: I am not 100% this is something we should do. The Dutch documents should actually stay available to members right?

website/documents/models.py
```diff
@@ -80,13 +80,10 @@ def save(self, *args, **kwargs):
         self.category = "annual"
         if self.subcategory == "report":
             self.name_en = "Annual report %d" % self.year
-            self.name_nl = "Jaarverslag %d" % self.year
```
se-bastiaan, line 83: Are these files in a archive somewhere?

website/documents/forms.py
se-bastiaan: And are Dutch board documents and/or gm documents in an archive somewhere?

website/education/migrations/0013_auto_20200907_1835.py
```diff
@@ -0,0 +1,31 @@
...
+            name='name_nl',
+        ),
+        migrations.RemoveField(
+            model_name='course',
+            name='name_nl',
+        ),
+        migrations.AlterField(
+            model_name='exam',
+            name='language',
+            field=models.CharField(choices=[('en', 'English')], max_length=2, null=True),
```
se-bastiaan, line 24: The actual language field of the documents is something we may want to keep. The documents in these models are now not suddenly translated

website/mailinglists/services.py
```diff
@@ -163,7 +163,7 @@ def get_automatic_lists():
                 f"to send newsletters in {language[1]}",
                 "addresses": _get_members_email_addresses(
                     Member.current_members.all().filter(
-                        profile__receive_newsletter=True, profile__language=language[0]
```
se-bastiaan, line 166: This line is technically required to actually consider this functionality for each language. So either:
1. This becomes a separate issue (my preference)
2. Is also changed

website/members/models/profile.py
```diff
@@ -209,13 +209,9 @@ class Profile(models.Model):
...
 
-    language = models.CharField(
-        verbose_name=_("Preferred language"),
-        help_text=_("Preferred language for e.g. newsletters"),
-        max_length=5,
-        choices=settings.LANGUAGES,
-        default="nl",
-    )
+    @property
+    def language(self):
```
se-bastiaan, line 213:
```suggestion
    def language(self):
        """
            @todo: Remove usage of this property
        """
```

website/photos/api/viewsets.py
```diff
@@ -12,7 +12,6 @@ class AlbumsViewSet(ModelViewSet):
     permission_classes = (permissions.IsAuthenticated,)
     queryset = Album.objects.all()
     filter_backends = (filters.SearchFilter,)
-    search_fields = ("title_en", "title_nl", "date", "slug")
```
se-bastiaan, line 15: I don't think this should be deleted completely

website/pizzas/tests.py
```diff
@@ -71,7 +65,7 @@ def setUpTestData(cls):
 
     def test_title(self):
         """Check the title attribute"""
-        for lang in ["en", "nl"]:
+        for lang in ["en"]:
```
se-bastiaan, line 68: This does not need to be a for-loop anymore

website/registrations/models.py
```diff
@@ -170,12 +170,9 @@ class Registration(Entry):
...
     birthday = models.DateField(verbose_name=_("birthday"), blank=False,)
 
-    language = models.CharField(
-        verbose_name=_("language"),
-        max_length=5,
-        choices=settings.LANGUAGES,
-        default="nl",
-    )
+    @property
+    def language(self):
```
se-bastiaan, line 174:
```suggestion
    def language(self):
        """
            @todo: Remove usage of this property
        """
```

website/thaliawebsite/templates/base.html
```diff
@@ -157,7 +148,7 @@
                 aria-label="Toggle navigation"><i class="fa fa-bars"></i>
         </button>
         <div
-            class="collapse navbar-collapse justify-content-end order-lg-0 order-2"
+            class="collapse navbar-collapse order-lg-0 order-2"
```
se-bastiaan, line 151:
```suggestion
            class="collapse navbar-collapse justify-content-end pr-lg-4 order-lg-0 order-2"
```

pingiun:
> It would be nice if the permissions of manage.py would stay on 0755. Otherwise `./manage.py` commands do not work anymore. Same for all the other files (100755 → 100644 in change list), unless this is required somehow.

I think this was an accident with the grep command

joren485:
deploy review

joren485:
@se-bastiaan The review environment is failing, because the `manage.py` permission bits change from `755` to `644` (i.e. removing the execute bit).

pingiun:
Review environment should work again

website/documents/forms.py
pingiun: > Not about this line but the file. I am not 100% this is something we should do. The Dutch documents should actually stay available to members right? Minutes maybe not, but the articles of association and the privacy policy?

I think for the general docs we can make a migration that converts these into "english" documents, with a suffix in the name. The privacy policy and minutes and stuff I'm not sure about yet.

> And are Dutch board documents and/or gm documents in an archive somewhere?

Yes the board has all of them in the Drive

website/mailinglists/services.py
pingiun: I don't really get what you're saying. What is required?

website/documents/forms.py
se-bastiaan: I'm not sure what you mean by converting them, and how we are going to show them. The files themselves will never be deleted.

website/mailinglists/services.py
se-bastiaan: We just need 1 mailinglist without the language in it. Thus the for-loop should also be removed.

se-bastiaan:
deploy review

se-bastiaan (approved):

PR #1207, opened 2020-09-06
Introduce a centralised payments page 

se-bastiaan:
### Summary

This PR introduces a centralised solution to be able to create payments when paying using Thalia Pay.
Since events already had such a page in place I have replaced this existing functionality so that it can be used as an example.

It basically works as follows:
A form submission is submitted to `/user/finance/payments/process/` with a body that contains
```
app_label=event
model_name=eventregistration
payable=<pk of event registration>
next=<redirection result>
```

This uses the [ContentType](https://docs.djangoproject.com/en/3.1/ref/contrib/contenttypes/#module-django.contrib.contenttypes) to dynamically obtain the model which is helped by the Payable interface we implemented before so that all these models can get a payment using this view.

### How to test
1. Create an event with registrations
2. Make sure you can pay using Thalia Pay
   - Enable the option in settings
   - Add a bank account with mandate to your user
3. Register for the event
4. Pay using Thalia Pay for your registration


website/events/templates/events/event.html
```diff
@@ -132,15 +132,19 @@ <h1 class="text-center section-title">{{ event.title }}</h1>
...
-                                                    <form action="{% url 'events:pay' event.id %}" method="post">
+                                                    <form action="{% url 'payments:payment-process' %}" method="post">
                                                         {% csrf_token %}
+                                                        <input type="hidden" name="model_name" value="eventregistration" />
+                                                        <input type="hidden" name="app_label" value="events" />
+                                                        <input type="hidden" name="payable" value="{{ registration.pk }}" />
+                                                        <input type="hidden" name="next" value="{{ request.path }}"/>
                                                         <input type="submit" class="btn btn-primary"
                                                                value="{% trans "Pay with Thalia Pay" %}"/>
                                                     </form>
```
JobDoesburg, line 150: Maybe, if you're doing this, make this a template tag in the payments app too

JobDoesburg (changes requested):

website/events/templates/events/event.html
se-bastiaan: that is something I considered indeed, but I was lazy

website/events/templates/events/event.html
```diff
@@ -132,18 +132,14 @@ <h1 class="text-center section-title">{{ event.title }}</h1>
...
                                             {% if registration.is_paid %}
                                                 {% if registration.payment.type == payment_method_tpay %}
-                                                    {% trans "You pay with Thalia Pay" %}
+                                                    {% trans "You are paying with Thalia Pay" %}
                                                 {% else %}
                                                     {% trans "You have paid" %}
                                                 {% endif %}
                                             {% else %}
                                                 {% trans "You have not paid yet" %}
                                                 {% if registration.member.tpay_enabled %}
```
JobDoesburg, line 141: Maybe the `tpay_enabled` check should be in the template tag

JobDoesburg (dismissed):
Looks nice! I have one small suggestion but maybe you had a good reason why you didn't do it that way

website/events/templates/events/event.html
se-bastiaan: I just didn't think about it :p

JobDoesburg (approved):

PR #1210, opened 2020-09-07
Fix used Docker image during test builds in Github Actions

se-bastiaan:
### Summary
Fix used Docker image during test builds in Github Actions
### How to test
Pls just merge.


pingiun (approved):
Lol

PR #1212, opened 2020-09-08
Remove download button for unauthorized users and make thabloids private

### LINKED ISSUE #965: Make Thabloid .pdf only available for authenticated users
In GitLab by @JobDoesburg on Dec 3, 2019, 16:44

### One-sentence description

Thabloids are crawable by Google, and we discussed that we rather not have that.

Google crawls our Thabloids. You can search for text fragments

### Expected behaviour

Not have them appear in Google. 
We do not necessarily have them only available on login, but add them to robots.txt to disallow search machines to display them would be nice

JobDoesburg:
Closes #965

### Summary
Thabloids are not accessible anymore if you are not logged in (so Google can't crawl them)

### How to test
1. Upload a thabloid
2. Try to download the thabloid as logged in user, this should work
3. Try to download the thabloid as AnonymousUser, this should not work

Note the migration for the media url!

### DIFF
--- a/website/thabloid/models.py
+++ b/website/thabloid/models.py
@@ -15,7 +15,7 @@ from utils.threading import PopenAndCall
 
 def thabloid_filename(instance, filename):
     ext = os.path.splitext(filename)[1]
-    return os.path.join("public/thabloids/", slugify(instance) + ext)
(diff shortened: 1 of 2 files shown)


se-bastiaan:
Closing this in favour of a new one that completely changes the workings to make it private.

PR #1217, opened 2020-09-09
Remove MultilingualField in announcements

pingiun:
This depends on #1202
## Summary

Remove MultilingualField in announcements
## How to test

Everything should still work as before


website/photos/api/viewsets.py
```diff
@@ -12,7 +12,6 @@ class AlbumsViewSet(ModelViewSet):
     permission_classes = (permissions.IsAuthenticated,)
     queryset = Album.objects.all()
     filter_backends = (filters.SearchFilter,)
-    search_fields = ("title_en", "title_nl", "date", "slug")
```
se-bastiaan, line 15: Why is this removed

se-bastiaan (changes requested):

se-bastiaan (approved):

PR #1225, opened 2020-09-18
Fix renewal form

JobDoesburg:
The renewal form had a required `contribution` field after #1054, which made it impossible to create regular membership renewals

### DIFF
--- a/website/registrations/migrations/0024_auto_20200325_2045.py
+++ b/website/registrations/migrations/0024_auto_20200325_2045.py
@@ -14,7 +14,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='entry',
             name='contribution',
-            field=models.FloatField(validators=[django.core.validators.MinValueValidator(7.5)], verbose_name='contribution'),
+            field=models.FloatField(validators=[django.core.validators.MinValueValidator(7.5)], default=7.5, verbose_name='contribution'),
         ),
         migrations.AlterField(
             model_name='entry',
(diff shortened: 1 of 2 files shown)


website/registrations/forms.py
```diff
@@ -99,6 +99,8 @@ class RenewalForm(forms.ModelForm):
         required=False, label=_("I am an employee of iCIS")
     )
 
+    contribution = forms.IntegerField(required=False,)
```
se-bastiaan, line 102: Now the field is never required, but it is for benefactors.

JobDoesburg: I tested this, and apparently the `clean()` will still require this

se-bastiaan:
deploy review

se-bastiaan (changes requested):
I have found two things that somehow are not perfectly working:
1. The default value of 7.5 is no longer entered in the field
2. My manually entered value of 7.5 is not accepted by the form, it only accepts integers and not floats
3. Optional: (When I enter something in the contribution field that is below 7,5 the whole form is no longer working)

se-bastiaan:
remove review

JobDoesburg:
I feel very stupid but I can't get it to work...

se-bastiaan:
So why I did not work:
1. You used an IntegerField instead of a DecimalField inside the form, this determines what is accepted by the browser HTML as well
2. The default value that was in the model before was no longer there, this value is what made it work ultimately

PR #1227, opened 2020-09-18
Thalia Pay for registrations

### LINKED ISSUE #935: Thalia Pay: Add an option to membership registrations to pay with Thalia Pay
In GitLab by @JobDoesburg on Oct 16, 2019, 20:24

### One-sentence description

Add an option to membership renewals and registrations to pay with Thalia Pay 

### Motivation

Everyone wants to do this via Thalia Pay remotely!

### Desired functionality

During registration (or after registration is approved, which is easier to build but less user friendly) people should have the possibility to immediately add a bank account and select payment via Thalia Pay.

### Suggested implementation

JobDoesburg:
Closes #935

### Summary
New members can now indicate that they want to pay their membership fees with Thalia Pay. When a member is created from a registration, not only a `Member` but now also a `BankAccount` object will be created. Moreover, when accepting entries, those registrations will immediately go from `accepted` to `completed` and automatically a payment will be created.

### How to test
1. Go, as an anonymous user, to the member/benefactor registration page
2. Register (and choose to use a bank account)
3. Go through the review proces
4. Check that all messages that are sent are correct
5. Check that a BankAccount is created, a payment is added and the registration is completed automatically after accepting
6. Make sure that for non-thalia-pay registrations everything also keeps working


JobDoesburg:
I would like to get your opinion on how I should design it.

- during registration, people can also already create a BankAccount object (but it will not have an owner yet as there is no user). After confirmation by the board, the bank account's owner field is updated.

- the registration itself will contain a copy of all fields a bank account requires, and after confirmation, these fields are copied to a new BankAccount object.

se-bastiaan:
I think it is best to include the required information in the registration itself instead of creating real model instances of the bank accounts. That will make sure that registrations won't directly depend on payments anymore than required

website/registrations/models.py
```diff
@@ -275,6 +277,38 @@ def language(self):
...
+    # ---- Bank account -----
+
+    direct_debit = models.BooleanField(
+        null=False,
+        blank=False,
+        default=False,
+        help_text="If selected, when the registration is accepted, immediately a "
+        "Thalia Pay payment will be created for this user and the registration "
+        "will be completed directly. This can only be selected if a bank "
+        "account is added with direct debit authorization.",
```
se-bastiaan, line 289:
```suggestion
        help_text=_("When the registration is accepted and this checkbox is enabled, a "
        "Thalia Pay payment will be created for this user and the registration "
        "will be completed immediately. This can only be selected if a bank "
        "account is added with direct debit authorisation during registration."),
```

website/registrations/models.py
```diff
@@ -338,6 +372,35 @@ def clean(self):
         if self.programme is None and self.membership_type != Membership.BENEFACTOR:
             errors.update({"programme": _("This field is required.")})
 
+        if self.direct_debit:
+
```
se-bastiaan, line 376:
```suggestion
```

website/registrations/templates/registrations/register_benefactor.html
```diff
@@ -90,6 +92,66 @@ <h1 class="text-center section-title">{% trans "registration" %} {% trans "Benef
...
+                                   name="signature"
+                                   class="form-control"
+                                   placeholder="Signature"
+                                   required=""
+                                   id="id_signature"
+                            >
+
+                            {% include 'payments/includes/signature_canvas.html' %}
+                        </div>
+                    </fieldset>
```
se-bastiaan, line 152: Maybe this part should be an imported template? Since it is completely copied in the member registration form.

se-bastiaan (changes requested):

website/registrations/templates/registrations/register_benefactor.html
JobDoesburg: I tried but the styling is subtly different

se-bastiaan (approved):

PR #1230, opened 2020-09-19
Add OAuth2 authentication & activemembers in the API

se-bastiaan:
#### OAuth

This PR proposes the addition of OAuth2 authentication to our setup and allows us to connect applications based on scopes (even though those are not yet implemented) and different tokens per application. 

Use cases are for example:
  - A website showing face recognition information for the lustrum
  - A discord bot that uses API information to hand out roles

In the second case you do not necessarily want to authenticate as a user, you just want to use client credentials for server side authentication to the API. That is something that is not possible with the way the current authentication works.

Moreover, if a member wants to implement something using our API this can be done by giving them their own client id.
This change does not mean that everything that uses the API (aka the app) suddenly has to switch, but we could decide to do so in the future. Especially since our current token system is quite rudimentary.

I have already added one scope so that a client that uses client_credentials can use the API to request user information without user credentials.

_Note_: This PR will not work in the review environment since the dependencies are not installed.

#### Active members

Since this OAuth is meant for the Discord bot I have also included changes to add a path for an activemembers API. This API will decrease the number of calls necessary to sync users.

### How to test
1. Download Postman
2. Set the authorisation in Postman to 'OAuth 2' and start requesting a new token
3. With the settings from Postman create a new application in the admin: http://localhost:8000/admin/oauth2_provider/application/
3. Complete the authentication in Postman
4. Use the obtained access token (Authorization header should be `Bearer <access-token>` to access `/api/v1/members/me`
5. Check the user menu for the authorised applications management pages
6. With your new token call `/api/v1/activemembers/groups/`


se-bastiaan:
deploy review

se-bastiaan:
remove review

se-bastiaan:
deploy review

se-bastiaan:
remove review

se-bastiaan:
deploy review

se-bastiaan:
remove review

se-bastiaan:
@mark-boute @JobDoesburg you are also allowed to review this ;)

JobDoesburg:
> @mark-boute @JobDoesburg you are also allowed to review this ;)

I will do it on Wednesday, first want to read a bit about how these things really work :)

pingiun (dismissed):

pingiun (approved):

PR #1243, opened 2020-09-23
Replace pizzas payment button for Thalia Pay with templatetag

### LINKED ISSUE #1240: Use payment_button for pizza payments with Thalia Pay
### Describe the change
Use payment_button for pizza payments with Thalia Pay

### Motivation
Uniformity

### Current implementation
We are using special payment views for pizza payments.

### Suggested implementation
Make pizza orders Payable and use the payment_button and PaymentProcessView from payments

se-bastiaan:
Closes #1240

### Summary

Replace pizzas payment button for Thalia Pay with templatetag. Among others this PR also adds:
- Checks when saving a payment so that an admin cannot create Thalia Pay payments for users who don't have it enabled
- Blocks deletion of payments in a processed batch

### How to test
1. Place a pizza order
2. Make sure your can pay with Thalia Pay

### DIFF
--- a/website/pizzas/models.py
+++ b/website/pizzas/models.py
@@ -10,6 +10,7 @@ from events.models import Event
 import members
 from members.models import Member
 from payments.models import Payment, Payable
+from payments.services import delete_payment
 from pushnotifications.models import ScheduledMessage, Category
 from utils.translation import ModelTranslateMeta, MultilingualField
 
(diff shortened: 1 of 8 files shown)


JobDoesburg:
TPAY payments should be removed if the order is cancelled

website/pizzas/models.py
```diff
@@ -126,6 +127,8 @@ def save(self, *args, **kwargs):
     def delete(self, using=None, keep_parents=False):
         if self.end_reminder is not None and not self.end_reminder.sent:
             self.end_reminder.delete()
+        if self.payment is not None and not self.payment.batch:
```
JobDoesburg, line 130:
```suggestion
        if self.payment is not None and not (self.payment.batch and self.payment.batch.processed):
```

JobDoesburg (changes requested):

website/pizzas/models.py
se-bastiaan: Nice catch, I should not have committed this code

JobDoesburg (approved):

PR #1248, opened 2020-09-23
Block Thalia Pay payment save when user does not have it enabled

se-bastiaan:
### Summary

This PR blocks saving a payment with the type Thalia Pay if the selected user does not have it enabled.

### How to test
1. Create a payment for a user that has Thalia Pay -> works
2. Create a payment for a user that does not have Thalia Pay -> blocked


lukovdm (approved):

website/payments/models.py
```diff
@@ -88,9 +88,13 @@ def save(self, **kwargs):
         super().save(**kwargs)
 
     def clean(self):
+        if self.type == Payment.TPAY and not self.paid_by.tpay_enabled:
+            raise ValidationError(
+                {"paid_by": _("This user does not have Thalia Pay enabled")}
```
JobDoesburg, line 93: This will crash if we try to change something about a payment after a user has revoked its mandate. So we should only do this check when we are creating a payment (or changing the payment type). So I think it should be in a pre_save signal. 
Or we should do a check whether the user **had** Thalia Pay enabled at the moment the payment was created.

JobDoesburg (changes requested):

website/payments/models.py
se-bastiaan: I have changed the validation to be only for new payments or payments that previously had a different type.

I do not entirely understand why this should be in a pre_save signal though, model validation is meant to be in `clean` and is executed before saving. A signal would not allow you to throw a ValidationError. Signals are more for decoupling of events that happen, this validation is something that is a crucial part of the model.

JobDoesburg (approved):
Nice, I did not know of the existence of `self._state.adding`

PR #1267, opened 2020-09-29
Add withdrawal date to batches

### LINKED ISSUE #1254: Add debit execution date to payment batch and email
We should notify users, when they receive a withdrawal notice email, when the amount will be withdrawn from their account.

Add a 'withdrawal date' to a payment batch and include that date in the emails. The default value should be current date + 14 days.

JobDoesburg:
Closes #1254 

### Summary
A withdrawal date is added to batches

### How to test
1. Create a batch, see the 'withdrawal date' being suggested 14 days from now
2. Process the batch, the email should contain the withdrawal date
3. Check out the payment list view, it should contain a settlement date


website/payments/models.py
```diff
@@ -146,6 +146,13 @@ class Batch(models.Model):
         verbose_name=_("description"), default=_default_batch_description,
     )
 
+    withdrawal_date = models.DateField(
+        verbose_name=_("withdrawal date"),
+        blank=False,
+        null=False,
+        default=timezone.now().date() + timezone.timedelta(days=14),
```
se-bastiaan, line 153: Could this also be a setting?

JobDoesburg: Yes

se-bastiaan (approved):

PR #1269, opened 2020-09-29
Add tpay_balance property to users

### LINKED ISSUE #1260: Display current amount of Thalia Pay debts on PaymentProcessView page
It should be clear to people, every time they pay with Thalia Pay, what their debts are. Also it is nice for the board to have this information at hand. 

- A user property that calculates the sum of all unprocessed Thalia Pay payments for that user
- Display that amount on the page where users confirm a Thalia Pay payment: "After this payment, your Thalia Pay balance will be -€ xx.xx"

JobDoesburg:
Closes #1260 

### Summary
The current Thalia Pay balance for a user is added

### How to test
1. Have a user with Thalia Pay payments
2. If the balance is not 0, the balance is visible on the Payment List page
3. When processing the balance, the new balance amount is shown
4. The balance is visible in the user admin

### DIFF
--- a/website/payments/admin.py
+++ b/website/payments/admin.py
@@ -575,24 +575,23 @@ class PaymentInline(admin.TabularInline):
 
 
 class ThaliaPayEnabledFilter(admin.SimpleListFilter):
-    title = "Thalia Pay enabled"
+    title = _("Thalia Pay enabled")
     parameter_name = "tpay_enabled"
 
     def lookups(self, request, model_admin):
-        return ("Yes", "Yes"), ("No", "No")
+        return ("1", _("Yes")), ("0", _("No"))
 
     def queryset(self, request, queryset):
-        value = self.value()
-        tpay_enabled = [x.id for x in PaymentUser.objects.all() if x.tpay_enabled]
-        if value == "Yes":
-            return PaymentUser.objects.filter(id__in=tpay_enabled)
-        if value == "No":
-            return PaymentUser.objects.exclude(id__in=tpay_enabled)
-        return None
+        tpay_enabled = [x.id for x in queryset.all() if x.tpay_enabled]
+        if self.value() == "1":
+            return queryset.filter(id__in=tpay_enabled)
+        if self.value() == "0":
+            return queryset.exclude(id__in=tpay_enabled)
+        return queryset
 
 
 class ThaliaPayBalanceFilter(admin.SimpleListFilter):
-    title = "Thalia Pay balance"
+    title = _("Thalia Pay balance")
     parameter_name = "tpay_balance"
 
     def lookups(self, request, model_admin):
@@ -602,13 +601,12 @@ class ThaliaPayBalanceFilter(admin.SimpleListFilter):
         )
 
     def queryset(self, request, queryset):
-        value = self.value()
-        tpay_balance = [x.id for x in PaymentUser.objects.all() if x.tpay_balance != 0]
-        if value == "0":
-            return PaymentUser.objects.exclude(id__in=tpay_balance)
-        if value == "1":
-            return PaymentUser.objects.filter(id__in=tpay_balance)
-        return None
+        tpay_balance = [x.id for x in queryset.all() if x.tpay_balance != 0]
+        if self.value() == "0":
+            return queryset.exclude(id__in=tpay_balance)
+        if self.value() == "1":
+            return queryset.filter(id__in=tpay_balance)
+        return queryset
 
 
 @admin.register(PaymentUser)
@@ -645,12 +643,12 @@ class PaymentUserAdmin(admin.ModelAdmin):
     def get_tpay_balance(self, obj):
         return f"€ {obj.tpay_balance:.2f}"
 
-    get_tpay_balance.short_description = "Balance"
+    get_tpay_balance.short_description = _("balance")
 
     def get_tpay_enabled(self, obj):
         return obj.tpay_enabled
 
-    get_tpay_enabled.short_description = "Thalia Pay enabled"
+    get_tpay_enabled.short_description = _("Thalia Pay enabled")
     get_tpay_enabled.boolean = True
 
     def user_link(self, obj):
--- a/website/payments/admin.py
+++ b/website/payments/admin.py
@@ -531,12 +531,147 @@ class BankAccountAdmin(admin.ModelAdmin):
     export_csv.short_description = _("Export")
 
 
+class BankAccountInline(admin.TabularInline):
+    model = BankAccount
+    fields = (
+        "iban",
+        "bic",
+        "mandate_no",
+        "valid_from",
+        "valid_until",
+        "last_used",
+    )
+    show_change_link = True
+
+    can_delete = False
+
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+
+class PaymentInline(admin.TabularInline):
+    model = Payment
+    fields = (
+        "created_at",
+        "type",
+        "amount",
+        "topic",
+        "notes",
+        "batch",
+    )
+
+    show_change_link = True
+
+    can_delete = False
+
+    def has_add_permission(self, request, obj=None):
+        return False
+
+    def has_change_permission(self, request, obj=None):
+        return False
+
+
+class ThaliaPayEnabledFilter(admin.SimpleListFilter):
+    title = "Thalia Pay enabled"
+    parameter_name = "tpay_enabled"
+
+    def lookups(self, request, model_admin):
+        return ("Yes", "Yes"), ("No", "No")
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        tpay_enabled = [x.id for x in PaymentUser.objects.all() if x.tpay_enabled]
+        if value == "Yes":
+            return PaymentUser.objects.filter(id__in=tpay_enabled)
+        if value == "No":
+            return PaymentUser.objects.exclude(id__in=tpay_enabled)
+        return None
+
+
+class ThaliaPayBalanceFilter(admin.SimpleListFilter):
+    title = "Thalia Pay balance"
+    parameter_name = "tpay_balance"
+
+    def lookups(self, request, model_admin):
+        return (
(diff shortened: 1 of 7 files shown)


website/members/models/member.py
```diff
@@ -5,13 +5,13 @@
...
 from django.contrib.auth.models import User, UserManager
-from django.db.models import Q
+from django.db.models import Q, Sum
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from activemembers.models import MemberGroup, MemberGroupMembership
-from payments.models import BankAccount
+from payments.models import BankAccount, Payment
```
se-bastiaan, line 14: I would rather keep the `Payment` model out of the members application as much as possible.

website/members/models/member.py
```diff
@@ -196,3 +196,13 @@ def tpay_enabled(self):
...
+
+    @property
+    def tpay_balance(self):
+        """The total amount of open, unprocessed Thalia Pay payments"""
+        payments = Payment.objects.filter(
+            Q(paid_by=self, type=Payment.TPAY)
+            & (Q(batch__isnull=True) | Q(batch__processed=False))
+        )
+        total = payments.aggregate(Sum("amount"))["amount__sum"]
+        return -1 * total if total else 0
```
se-bastiaan, line 208: When would total be None or False? Isn't a sum of nothing 0?

se-bastiaan (changes requested):

website/members/models/member.py
JobDoesburg: Nope apparently the sum of nothing is `None`

website/members/admin.py
```diff
@@ -189,6 +201,14 @@ class UserAdmin(BaseUserAdmin):
         ),
     )
 
+    readonly_fields = ("tpay_balance",)
+
+    def tpay_balance(self, obj):
```
se-bastiaan, line 206: If you register a separate admin in the payments package with a per user overview of the payments you can also add custom filters for users that have a balance that is 0. 

This way you would have two different admins for users and the balance and this admin is complete decoupled from payments. For 99% of the time this data is not really necessary

website/payments/templates/payments/payment_list.html
```diff
@@ -17,6 +17,12 @@ <h1 class="text-center section-title">{% trans "payments"|capfirst %}</h1>
                 {% endfor %}
             {% endif %}
 
+            {% if request.member.tpay_balance != 0 %}
+                <h5 class="text-center mb-4">
+                   Your current Thalia Pay balance is: € {{ request.member.tpay_balance|floatformat:2 }}
```
se-bastiaan, line 22: I believe this will not work

website/payments/models.py
```diff
@@ -88,6 +90,8 @@ def save(self, **kwargs):
         super().save(**kwargs)
 
     def clean(self):
+        from payments.services import tpay_enabled
```
se-bastiaan, line 93: Imports should be global

website/payments/models.py
```diff
@@ -339,3 +343,9 @@ def payment_payer(self):
 
     def save(self):
         raise NotImplementedError
+
+
+class TPayUser(Member):
```
se-bastiaan, line 348:
```suggestion
class FinanceMember(Member):
```

website/payments/services.py
```diff
@@ -124,3 +124,23 @@ def send_tpay_batch_processing_emails(batch):
                 },
             )
     return len(member_payments)
+
+
+def tpay_enabled(member):
```
se-bastiaan, line 129: This is not used everywhere

website/payments/models.py
se-bastiaan:
```suggestion
class PaymentMember(Member):
```

website/payments/admin.py
```diff
@@ -525,3 +525,107 @@ def export_csv(self, request: HttpRequest, queryset: QuerySet) -> HttpResponse:
...
+            x.id for x in TPayUser.objects.all() if services.tpay_enabled(x)
+        ]
+        if value == "Yes":
+            return TPayUser.objects.filter(id__in=tpay_enabled)
+        elif value == "No":
+            return TPayUser.objects.exclude(id__in=tpay_enabled)
+        return TPayUser.objects
+
+
+class TPayBalanceFilter(admin.SimpleListFilter):
```
se-bastiaan, line 552:
```suggestion
class ThaliaPayBalanceFilter(admin.SimpleListFilter):
```

website/payments/admin.py
```diff
@@ -525,3 +525,107 @@ def export_csv(self, request: HttpRequest, queryset: QuerySet) -> HttpResponse:
         return response
 
     export_csv.short_description = _("Export")
+
+
+class TPayEnabledFilter(admin.SimpleListFilter):
```
se-bastiaan, line 530:
```suggestion
class ThaliaPayEnabledFilter(admin.SimpleListFilter):
```

se-bastiaan (changes requested):

website/payments/models.py
JobDoesburg: Yeah so the problem is that this results in a circular import. And now I also don't really know what is the best to fix this 🙁 I do like the fact that Payments depends on Members and Members does not depend on Payments at all, though

JobDoesburg: Okay I have been thinking about this and I think, codequalitywise, we should do this:
- refactor `BankAccount`s and `Payment`s to use `PaymentUser`s
- refactor the `registrations`, `events` and `pizzas` app to also user `PaymentUser`s (and for `events` perhaps even create a new `EventUser`)
- #1277 can then be added to `PaymentUser` (and perhaps the `can_access_event` should be moved to `EventUser` too)

this way, we have better uncluttered code where it is clear how all apps depend on each other:

- `members` depends on Django `auth`
- `payments` (and almost all other apps) depends on `members`
- `registrations` and `events` depend on `payments`
- `pizzas` depends on `events`

and this way we will never have circular imports ever again. 

It is quite a change but I think it is one for the better. What do you think?

JobDoesburg: @se-bastiaan what is your opinion?

lukovdm: I just discussed this with Job, I think it is a good idea.

pingiun:
Please rebase this to include changes from #1292

JobDoesburg:
I did the proposed refactor in #1320, if accepted I will rebase this PR (and after that I can work on #1277)

website/payments/admin.py
```diff
@@ -531,12 +531,147 @@ def export_csv(self, request: HttpRequest, queryset: QuerySet) -> HttpResponse:
...
+    def has_change_permission(self, request, obj=None):
+        return False
+
+
+class ThaliaPayEnabledFilter(admin.SimpleListFilter):
+    title = "Thalia Pay enabled"
+    parameter_name = "tpay_enabled"
+
+    def lookups(self, request, model_admin):
+        return ("Yes", "Yes"), ("No", "No")
```
se-bastiaan, line 582: Why use these capitalised words as filter values? Wouldn't it make more sense to use integers or plain booleans? Something that is at least lowercased like for the other filter.

website/payments/admin.py
```diff
@@ -531,12 +531,147 @@ def export_csv(self, request: HttpRequest, queryset: QuerySet) -> HttpResponse:
...
+
+    def lookups(self, request, model_admin):
+        return (
+            ("0", "€0,00"),
+            ("1", ">€0.00"),
+        )
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        tpay_balance = [x.id for x in PaymentUser.objects.all() if x.tpay_balance != 0]
```
se-bastiaan, line 606: You should be able to create 1 query here instead of 3:
1. You get the user
2. You calculate the balance
3. You get the users based on the balance

Or at least reduce it to 2.

website/payments/admin.py
```diff
@@ -531,12 +531,147 @@ def export_csv(self, request: HttpRequest, queryset: QuerySet) -> HttpResponse:
...
         "last_name",
-        "email",
         "username",
+        "email",
     )
+
+    def get_tpay_balance(self, obj):
+        return f"€ {obj.tpay_balance:.2f}"
+
+    get_tpay_balance.short_description = "Balance"
```
se-bastiaan, line 648:
```suggestion
    get_tpay_balance.short_description = _("balance")
```

Even though we do not have translations, just like for the emails, we can still keep adding the translation helpers.

se-bastiaan: You even did it on line 668.

website/payments/templates/payments/payment_form.html
```diff
@@ -34,6 +34,10 @@ <h1 class="text-center section-title">{% trans "confirm payment"|capfirst %}</h1
                 {% endblocktrans %}
             </p>
 
+            <p class="text-center">
+                 After this payment, your Thalia Pay balance will be € {{ new_debt|floatformat:2 }}.
```
se-bastiaan, line 38: Translation block

website/payments/templates/payments/payment_list.html
```diff
@@ -17,6 +17,12 @@ <h1 class="text-center section-title">{% trans "payments"|capfirst %}</h1>
                 {% endfor %}
             {% endif %}
 
+            {% if tpay_balance and tpay_balance != 0 %}
+                 <h5 class="text-center mb-4">
+                    Your current Thalia Pay balance is: € {{ tpay_balance|floatformat:2 }}
```
se-bastiaan, line 22: Translation block

website/payments/views.py
```diff
@@ -184,6 +189,14 @@ def dispatch(self, request, *args, **kwargs):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context.update({"payable": self.payable})
+        context.update(
+            {
+                "new_debt": PaymentUser.objects.get(
```
se-bastiaan, line 194: if you use 'balance' everywhere this should be 'new_balance'

se-bastiaan (changes requested):

se-bastiaan (approved):

PR #1276, opened 2020-09-30
Prevent event registration cancelling after payment

### LINKED ISSUE #1256: Prevent event registrations cancelling when registration is paid (or where you were present)
Currently it is technically possible to pay for an event and afterwards cancel your registration. That should not be possible (at least not by users themselves). This never was a real problem in practice but now with Thalia Pay the problem appears.

When a payment exists for an event registration, don't allow cancelling. Also when paying with Thalia Pay we might want to make explicit that after paying you can't cancel anymore (without maybe contacting the board). 

- Only really creating the payment at the date of the event → too complex (code)
- As proposed here, but do allow cancellation when the payment is a TPay payment that is not processed yet → too complex (for users)
- Keeping things as it is (allow cancelling and accept people may pay without joining - it could be considered a feature) → felt undesirable, also without TPay

9Y0:
Closes #1256 

### Summary
An user can no longer cancel it's registration after having payed

### How to test
1. Register for an event with a payment
2. Pay for the event
3. Try to cancel the registration

### DIFF
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -150,7 +150,7 @@ def cancel_registration(member, event):
         registration.date_cancelled = timezone.now()
         registration.save()
     else:
-        if registration.payment is not None:
+        if registration is not None and registration.payment is not None:
             raise RegistrationError(
                 _("You can't cancel a registration after having payed.")
             )
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -62,6 +62,7 @@ def event_permissions(member, event, name=None):
         registration is not None
         and registration.date_cancelled is None
         and (event.cancellation_allowed or name)
+        and registration.payment is None
     )
     perms["update_registration"] = (
         registration is not None
@@ -136,10 +137,6 @@ def cancel_registration(member, event):
         pass
 
     if event_permissions(member, event)["cancel_registration"] and registration:
-        if registration.payment is not None:
-            raise RegistrationError(
-                _("You can't cancel a registration after having payed.")
-            )
         if registration.queue_position == 0:
             emails.notify_first_waiting(event)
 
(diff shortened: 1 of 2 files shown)


website/events/services.py
```diff
@@ -137,7 +137,9 @@ def cancel_registration(member, event):
 
     if event_permissions(member, event)["cancel_registration"] and registration:
```
se-bastiaan, line 138: The events package has the `event_permissions` function in this same file that determines if you should be allowed to do an action like cancelling the registration. So instead of blocking this here you may want to change that permissions function instead.

website/events/services.py
```diff
@@ -151,7 +150,12 @@ def cancel_registration(member, event):
         registration.date_cancelled = timezone.now()
         registration.save()
     else:
-        raise RegistrationError(_("You are not registered for this event."))
+        if registration is not None and registration.payment is not None:
+            raise RegistrationError(
+                _("You can't cancel a registration after having payed.")
+            )
+        else:
+            raise RegistrationError(_("You are not registered for this event."))
```
9Y0, line 158: I personally don't really like this approach since I'm basically checking for the same condition again, but I'm not sure either if returning a general error message like "You can't deregister for this event" is ideal.

website/events/services.py
```diff
@@ -151,7 +150,12 @@ def cancel_registration(member, event):
         registration.date_cancelled = timezone.now()
         registration.save()
     else:
-        raise RegistrationError(_("You are not registered for this event."))
+        if registration is not None and registration.payment is not None:
+            raise RegistrationError(
+                _("You can't cancel a registration after having payed.")
```
se-bastiaan, line 155:
```suggestion
                _("You can't cancel a registration after having paid.")
```

website/events/services.py
se-bastiaan: Technically you are creating an if/else inside an else. So it could already be be

```
elif registration and registration.payment:
    ... paid
else:
   ... not registered
```

se-bastiaan: I'm not sure if the error is even used though, so maybe a general one doesn't matter?

se-bastiaan (approved):

PR #1277, opened 2020-09-30
Create personal Thalia Pay permissions

### LINKED ISSUE #947: Thalia Pay: Add user-instructions for activating in bankaccounts
In GitLab by @JobDoesburg on Oct 30, 2019, 21:09

### One-sentence description

Add information about Thalia Pay to bankaccounts

### Motivation

People should know what Thalia Pay is and how to activate it.

### Desired functionality

Something about that if you sign a direct debit, it will be used for Thalia Pay etc.

### Suggested implementation

Maybe ask the board? Or just come up with something ourselves. It should not be too extensive in my opinion, just something short.

### LINKED ISSUE #1257: Add a per user 'is allowed to use Thalia Pay' property
We decided we want to roll out Thalia Pay as a closed beta. Also, we want to blacklist people from using Thalia Pay, for example if they have too large debts or for any other reason. 

Every user gets a boolean property 'is allowed to use Thalia Pay'. Only if that is set to True, **and** he/she has a valid bank account, the `tpay_enabled` property (that exists already) will be True too. 

### Motivation
- This way we can roll out Thalia Pay as a closed beta first
- It is a easy kill switch for the board

### Additional context
In my opinion we should **also** keep the global Thalia Pay setting as an easy global kill switch

### LINKED ISSUE #1258: Make TPay payment button visible to everyone who is allowed to use Thalia Pay
We want to roll out Thalia Pay to as much people as possible. For that, people must know about the existence of Thalia Pay

The `Pay with Thalia Pay` button now is only displayed to people that have a valid bank account. We should change this to everyone who is allowed to use Thalia Pay (#1257), but make the button disabled for people without a bank account, and show them how they can add a bank account to use it

JobDoesburg:
Closes #1257 and closes #1258 and closes #947

### Summary
Every user profile now also has a boolean (default True) whether that member can use Thalia Pay or not. Only members that have that set to True can actually use and see Thalia Pay. People that are allowed to use Thalia Pay, but do not have a valid bank account, will see disabled Thalia Pay buttons with a popover instructing them how to start using it.

If users are disallowed to use Thalia Pay, all buttons will disappear. And still, when the global setting `THALIA_PAY_ENABLED` is disabled, paying will never be possible.

Also added two actions for the user admin

### How to test
1. Check the boolean field in user profiles
2. Check the 2 actions in the user admin
3. Check out the Thalia Pay button everywhere


JobDoesburg:
For some weird reason I can't get the popover to work yet. I think it is specific to my machine as I get some weird errors when loading stylesheets.

se-bastiaan (commented):
Isn't this exactly the opposite of my suggestion in the issue?

JobDoesburg:
> Isn't this exactly the opposite of my suggestion in the issue?

The default is ‘True’, and this felt more intuitive for me actually

se-bastiaan:
So yes. But isn't that exactly what it will always be? In my opinion you're creating a boolean that is supposed to be never false. So turning it around and make it a blacklist would make more sense.

In any way, if this is a default user permission it could as easily be a permission on the Payment model that is set to every user object in a post creation save. That's what permissions are for.

website/payments/signals.py
```diff
@@ -0,0 +1,23 @@
...
+
+@receiver(post_save, sender=Member)
+def give_new_users_tpay_permissions(instance, created, **kwargs):
+    """If a new user is created, give this user Thalia Pay permissions."""
+    if created and settings.THALIA_PAY_FOR_NEW_MEMBERS:
+        allow_tpay = PaymentUser.allow_tpay
+        allow_tpay(instance)
+
+        # Note that because Django does not support accessing the instance
+        # in the PaymentUser proxy model in this signal, we rely on the fact
```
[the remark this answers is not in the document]
JobDoesburg: I tried connecting the signal to the ProxyModel but that doesn’t work. It’s still an ongoing django discussion how and whether proxy models signals should be triggered (if you ask me, they should but I see the underlying problem they have). 

I think explaining the code doesn’t hurt and it might prevent future concrexiteers a lot of searching why things behave the way they do

website/thaliawebsite/settings/settings.py
```diff
@@ -329,6 +329,9 @@
 THALIA_PAY_ENABLED_PAYMENT_METHOD = (
     os.environ.get("THALIA_PAY_ENABLED", "False") == "True"
 )
+THALIA_PAY_FOR_NEW_MEMBERS = (
+    os.environ.get("THALIA_PAY_FOR_NEW_MEMBERS", "True") == "True"
```
se-bastiaan, line 333:
```suggestion
    os.environ.get("THALIA_PAY_FOR_NEW_MEMBERS", "False") == "True"
```

se-bastiaan: Potato potato :)

Let's not forget this in the release

se-bastiaan (approved):

website/thaliawebsite/settings/settings.py
JobDoesburg: > Let's not forget this in the release

Even if we forget it, but we don't change `THALIA_PAY_ENABLED` its not a problem

PR #1279, opened 2020-09-30
Add Makefile

pingiun:
Adds a Makefile, which makes it easier to run the right commands for the tests and such. The makefile includes a help command, which looks like this:

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,133 @@
+.POSIX:
+.SUFFIXES:
+
+VIRTUALENV := $(shell poetry env info | grep Path: | sed 's/Path:[[:space:]]*//')
+PYTHONFILES := $(shell find website -name '*.py')
+MIGRATIONS := $(shell find website -name '*.py' | grep migrations)
+DOCSFILES := $(shell find docs -name '*.rst')
+TEMPLATEFILES := $(shell find website -name '*.html')
+
+PORT ?= 8000
+
+ifdef CI
+	POETRY_FLAGS := $(POETRY_FLAGS) --no-interaction
+	BLACK_FLAGS := $(BLACK_FLAGS) --quiet
+	DOCKER_FLAGS := $(DOCKER_FLAGS) --quiet
+endif
+
+help:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,132 @@
...
+
+.make/check: .make $(VIRTUALENV) $(PYTHONFILES)
+	poetry run python website/manage.py check
+	@touch .make/check
+
+check: .make/check ## Run internal Django tests
+
+.make/templatecheck: .make $(VIRTUALENV)
+	poetry run python website/manage.py templatecheck --project-only
+	@touch .make/templatecheck
```
se-bastiaan, line 75: What does this touch  do?

Makefile
```diff
@@ -0,0 +1,132 @@
...
+.make/migrationcheck: .make $(VIRTUALENV) $(PYTHONFILES)
+	poetry run python website/manage.py makemigrations --no-input --check --dry-run
+	@touch .make/migrationcheck
+
+migrationcheck: .make/migrationcheck ## Check if migrations are created
+
+.coverage: $(VIRTUALENV) $(PYTHONFILES)
+	poetry run coverage run website/manage.py test website/
+
+tests: .coverage ## Run tests with coverage
```
se-bastiaan, line 88: Wouldn't it make more sense to call this command 'coverage' and have it run the tests _and_ generate the coverage reports?

Makefile
pingiun: This should depend on the template files, which I forgot. But then it would only do the templatecheck if the template files are newer than the touched file in the .make directory

Makefile
pingiun: I think you missed the `coverage` task which does exactly this. The tests command can be used if you don't care about the report

Makefile
pingiun: Added the dependency now

Makefile
se-bastiaan: Oh I missed that it also runs the tests themselves

se-bastiaan (approved):

Issue #1290, opened 2020-10-03
Move TODOs to issues


joren485:
### Describe the change
Move all `TODO`s in the code to issues.

### Motivation
`TODO`s in the code means that there is some issue to solve. These issues are not easily visible by developers. It would be better to have them as issues on GitHub (along with all other issues).

### Additional context
```
$ grep -ir TODO --exclude="*.pyc" --exclude="*fullcalendar*" --exclude-dir="docs" --exclude-dir="*font*"
website/registrations/models.py:            @todo: Remove usage of this property
website/utils/translation.py:            # TODO consider implementing this when there is a need for it.
website/members/models/profile.py:            @todo: Remove usage of this property
website/payments/templates/payments/bankaccount_list.html:{#            TODO: Add a mention to payments functionality when we build it  #}
website/thabloid/models.py:        # TODO reconsider if this resolution / quality is sufficient
website/thaliawebsite/menus.py:                # TODO: Remove submenu when the new template is implemented
```

PR #1292, opened 2020-10-03
Introduce PyLint to the codebase

se-bastiaan:
### Summary

This PR reintroduces a linter into our codebase, which should improve anything that is not actual codestyle.

We remove our PEP checks from our codebase when we introduced black, but now we have unused imports (#1289) and random todos that we never see again (#1290) plus static methods that are not actually marked static (#1288).

PyLint will take care of these mistakes! No more common mistakes and better code! Together with Black this is perfect for our codebase.

### How to test
1. Run `make pylint`


website/announcements/models.py
```diff
@@ -42,7 +42,7 @@ class Meta:
         ordering = ("-since",)
 
     def __str__(self):
-        return self.content
+        return str(self.content)
```
pingiun, line 45: Should these things not always be strings anyway?

website/events/forms.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, *args, **kwargs):
             self.fields[key].initial = field["value"]
 
     def field_values(self):
-        for key, field in self.information_fields.items():
+        for key, __ in self.information_fields.items():
```
pingiun, line 82: Looping through keys is what happens when you do `for key in dict`, so the .items() should be removed here

website/photos/services.py
```diff
@@ -130,7 +130,7 @@ def extract_photo(request, archive_file, photo, album):
 
     # Generate unique filename
     num = album.photo_set.count()
-    filename, extension = os.path.splitext(photo_filename)
+    __, extension = os.path.splitext(photo_filename)
```
pingiun, line 133: Why are two underscores used instead of one?

Makefile
```diff
@@ -65,6 +65,12 @@ fmt: .make/fmt ## Format python code with black
 blackcheck: $(VIRTUALENV) $(PYTHONFILES) ## Check if everything is formatted correctly
 	poetry run black $(BLACK_FLAGS) --check website
 
+.make/pylint: .make $(VIRTUALENV) $(PYTHONFILES)
+	cd website && DJANGO_SETTINGS_MODULE=thaliawebsite.settings poetry run pylint --load-plugins pylint_django ./**/*.py
```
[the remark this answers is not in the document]
se-bastiaan: Could you try to get that working maybe? I had significant problems when trying. Especially since the website folder is not a module, it is only the root folder. So when you run Python from the root directory all the imports are no longer valid.

website/.pylintrc
```diff
@@ -0,0 +1,549 @@
...
+
+# Number of spaces of indent required inside a hanging  or continued line.
+indent-after-paren=4
+
+# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
+# tab).
+indent-string='    '
+
+# Maximum number of characters on a single line.
+max-line-length=79
```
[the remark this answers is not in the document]
se-bastiaan:
```suggestion
```

website/photos/services.py
se-bastiaan: `_` shadows the import of `lazy_gettext`

Makefile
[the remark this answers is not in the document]
se-bastiaan: Okay somehow it works like this, and it is now finding more somehow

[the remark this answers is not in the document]
se-bastiaan: Changed

website/events/forms.py
```diff
@@ -79,8 +79,8 @@ def __init__(self, *args, **kwargs):
             self.fields[key].initial = field["value"]
 
     def field_values(self):
-        for key, field in self.information_fields.items():
-            yield (key, self.cleaned_data[key])
+        for (key,) in self.information_fields:
```
pingiun, line 82: I think this should not be a singleton tuple, just `for key in ...`

website/members/models/profile.py
```diff
@@ -259,36 +259,36 @@ def language(self):
...
             return "{} '{}' {}".format(
                 self.user.first_name, self.nickname, self.user.last_name
             )
-        elif pref == "nicklast" and self.nickname is not None:
+        if pref == "nicklast" and self.nickname is not None:
             return "'{}' {}".format(self.nickname, self.user.last_name)
-        else:
-            return self.user.get_full_name() or self.user.username
+        return self.user.get_full_name() or self.user.username
+        # pylint: disable=too-many-return-statements
```
pingiun, line 279: It seems this line can be removed? Doesnt the line at the top disable it for just this function?

pingiun (dismissed):
Seems to be all good, please rebase and resolve conflicts so this can be merged

pingiun (approved):

PR #1293, opened 2020-10-03
Save newsletter to disk when sending

### LINKED ISSUE #1256: Prevent event registrations cancelling when registration is paid (or where you were present)

9Y0:
Closes #1256 

### Summary
The newsletter was saved by the email sending code which for some reason doesn't actually work. The newsletter is now saved by the newsletter services itself, which does work.

### How to test
1. Create a newsletter
2. Send it to the members
3. Notice that the newsletter got saved to disk

### DIFF
--- a/website/newsletters/emails.py
+++ b/website/newsletters/emails.py
@@ -62,8 +62,6 @@ def send_newsletter(newsletter):
         html_message = html_template.render(context)
         text_message = text_template.render(context)
 
-        services.write_to_file(newsletter.pk, language[0], html_message)
-
         msg = EmailMultiAlternatives(
             subject=subject,
             body=text_message,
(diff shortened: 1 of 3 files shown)


se-bastiaan (changes requested):
The `add_argument` calls in the html creation command could also be removed

se-bastiaan (approved):

PR #1295, opened 2020-10-04
Remove NL preview button from newsletters admin

se-bastiaan:
### Summary

Remove NL preview button from newsletters admin. Because clicking them results in an error.

### How to test
1. Create a newsletter
2. Check that there is only the default preview button


9Y0 (commented):
The `preview-localised` url could be removed since it's no longer used

se-bastiaan:
Technically it could be removed, but old (Dutch) newsletters can still be accessed by serving their HTML. So I figured that we should keep them for the time being.

9Y0 (approved):

PR #1296, opened 2020-10-04
Fix creating Document fixtures

pingiun:
### Summary
createfixtures was broken because of the l12n changes, this fixes it by renaming the attribute names

### How to test
Before createfixtures would not work and crash, after this change it should not crash anymore


se-bastiaan (approved):

Legosteen11 (approved):
Fixed the issue for me :+1:

PR #1299, opened 2020-10-05
Use DecimalField for contribution

### LINKED ISSUE #1229: Change registration contribution from FloatField to DecimalField
### Describe the change
We currently user FloatFields for contribution amounts. We should use DecimalField's.

### Motivation
Events also use DecimalFields for the price of an event. Also, consider [https://docs.python.org/3/library/decimal.html#module-decimal](https://docs.python.org/3/library/decimal.html#module-decimal)

### Current implementation
FloatFields

### Suggested implementation
Just change it

9Y0:
Closes #1229 

### Summary
Changed the FloatField to a DecimalField

### How to test
1. Add a renewal
2. The contribution field is now a DecimalField


se-bastiaan (approved):

pingiun:
Why a decimal and not use cents?

PR #1320, opened 2020-10-12
Refactor payments to use PaymentUser proxy model

JobDoesburg:
Check #1269 for more information. This PR is the required (partial) refactor for that PR.

### Summary
Payments are refactored to use a proxy model. This way it is easy to add some methods to these users.

### How to test
Make sure this doesn't break things.


lukovdm (approved):
Nice quick work

website/payments/services.py
```diff
@@ -8,22 +8,21 @@
...
 from registrations.emails import _send_email
 from .exceptions import PaymentError
-from .models import Payment, BankAccount, Payable
+from .models import Payment, BankAccount, Payable, PaymentUser
 
 
 def create_payment(
     payable: Payable,
-    processed_by: Member,
+    processed_by: PaymentUser,
```
se-bastiaan, line 18: Why does this need to be a PaymentUser exactly? You're proxying the model and the value here is saved immediately.

website/events/models/event_registration.py
```diff
@@ -151,7 +151,7 @@ def payment_notes(self):
 
     @property
     def payment_payer(self):
-        return self.member
+        return PaymentUser.objects.get(pk=self.member.pk) if self.member else None
```
se-bastiaan, line 154: Why does this need to be a PaymentUser exactly? The place where this property is used could convert it, then you don't need to import the model here in events.

website/pizzas/models.py
```diff
@@ -218,7 +218,7 @@ def payment_notes(self):
 
     @property
     def payment_payer(self):
-        return self.member
+        return PaymentUser.objects.get(pk=self.member.pk)
```
se-bastiaan, line 221: Why does this need to be a PaymentUser exactly? The place where this property is used could convert it, then you don't need to import the model here in pizzas

website/registrations/models.py
```diff
@@ -108,7 +108,7 @@ def payment_amount(self):
     @property
     def payment_payer(self):
         if self.membership:
-            return self.membership.user
+            return PaymentUser.objects.get(pk=self.membership.user.pk)
```
se-bastiaan, line 111: Why does this need to be a PaymentUser exactly? The place where this property is used could convert it, then you don't need to import the model here

website/registrations/models.py
```diff
@@ -364,7 +364,7 @@ class Renewal(Entry):
 
     @property
     def payment_payer(self):
-        return self.member
+        return PaymentUser.objects.get(pk=self.member.pk)
```
se-bastiaan, line 367: Why does this need to be a PaymentUser exactly? The place where this property is used could convert it, then you don't need to import the model here

website/registrations/services.py
```diff
@@ -367,7 +368,7 @@ def process_entry_save(entry: Entry) -> None:
         # gets the welcome email
         emails.send_renewal_complete_message(entry.renewal)
 
-    entry.payment.paid_by = member
+    entry.payment.paid_by = PaymentUser.objects.get(pk=member.pk)
```
se-bastiaan, line 371: Why does this need to be a PaymentUser exactly? The place where this property is used could convert it, then you don't need to import the model here

se-bastiaan (changes requested):

website/registrations/services.py
JobDoesburg: So actually here we are doing a bit of a trick because we are 'fixing' the payment after the user has been created. So we are not using the native `create_payment` flow here.

se-bastiaan: So yes, I did mean create payment. But that does not mean it is no longer true. The PaymentUser model is a proxy model. So it should not require you to get it before it saves it directly again right? The foreignkey only saves the pk and the proxy model doesn't change the database fields.

JobDoesburg: Okay maybe it works, but it also feels weird to rely on the fact that its a proxy model. Better idea: use `create_payment` again. If a payment is already there, `create_payment` will update the payment so that would work.

JobDoesburg: Nvm that fucks things up. I'll just go with Member and I'll add a comment explaining it

se-bastiaan (approved):

PR #1324, opened 2020-10-13
Make thabloids private

### LINKED ISSUE #965: Make Thabloid .pdf only available for authenticated users

JobDoesburg:
Closes #965 

### Summary
Thabloids are only accessible by logged-in users now

### How to test
1. check all thabloid urls
2. note that also the media urls are now private

After this PR, old thabloid files must be moved on production (or we can write a migration for that if we really want that)

### DIFF
--- a/website/thabloid/migrations/0006_make_thabloids_private.py
+++ b/website/thabloid/migrations/0006_make_thabloids_private.py
@@ -5,28 +5,30 @@ from django.db import migrations
 
 
 def make_public_thabloids_private(apps, schema_editor):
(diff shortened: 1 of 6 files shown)


se-bastiaan:
How do you want to transfer the existing files in the directories? You could write a migration to do this.

se-bastiaan (approved):

PR #1333, opened 2020-10-18
Add presence information to the events api

pingiun:
### Summary
Add presence information to the events api

This change supports a feature in [Thalia Voting](https://github.com/pingiun/helios-server)

### How to test
1. Register for an event
2. Change presence to present
3. View events list api
4. Check present = true


9Y0:
I'm not entirely sure if this information should be available in the public api.

se-bastiaan (approved):

pingiun:
> I'm not entirely sure if this information should be available in the public api.

You can only view your own presence via this API

PR #1335, opened 2020-10-19
Add install_dev_requirements ARG to Dockerfile

joren485:
### Summary
Currently the Dockerfile does not have `install_dev_requirements` (https://github.com/svthalia/concrexit/pull/1190) set, because before https://github.com/svthalia/concrexit/pull/1230 it was not used.

https://github.com/svthalia/concrexit/pull/1230 introduced a `RUN` statement that uses `install_dev_requirements`. Currently that statement generates a error (because `install_dev_requirements` is not set) and runs `poetry install --no-interaction --no-dev` removing the development packages from the image. 

If there are dependency changes, they also need to be installed in the Docker image.

### DIFF
--- a/.github/workflows/build-production-docker.yaml
+++ b/.github/workflows/build-production-docker.yaml
@@ -27,7 +27,7 @@ jobs:
             - name: Build Docker images
               run: |
                   docker build --file Dockerfile.dependencies --quiet --build-arg "install_dev_requirements=0" --tag "${DOCKER_DEPENDENCIES_LATEST}" .
-                  docker build --quiet --build-arg "source_commit=${{ github.sha }}" --tag "${DOCKER_LATEST}" .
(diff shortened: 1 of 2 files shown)


joren485:
deploy review

pingiun:
Shouldn't the normal docker image always poetry install, so package upgrades after the dependencies docker was generated are stsill picked up?

se-bastiaan (changes requested):
This will always use the dependencies installed by the scheduled dependencies image, so we cannot trust the dependency update test builds since the build is always done based on the dependency version in master.

joren485:
Ah yes that is a good point @se-bastiaan @pingiun. Looking at the [MR in Gitlab](https://gitlab.science.ru.nl/thalia/concrexit/-/merge_requests/1488) you are right. 

> `poetry install` is still being run on every job, to make sure dependencies specific to that branch are being installed.

Which is true, except for the Docker images themselves.

I will rewrite this PR.

se-bastiaan (approved):

PR #1341, opened 2020-10-20
Set better defaults for database variables

pingiun:
## Summary

Set better defaults for database variables. Now when the environment variables are not specified, the database will try local authentication (this means that the postgres socket on the localhost is used, and authentication is based on the username)

### DIFF
--- a/website/thaliawebsite/settings/production.py
+++ b/website/thaliawebsite/settings/production.py
@@ -35,9 +35,9 @@ DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
         "USER": os.environ.get("POSTGRES_USER", "postgres"),
-        "PASSWORD": os.environ.get("POSTGRES_PASSWORD", ""),
+        "PASSWORD": os.environ.get("POSTGRES_PASSWORD", None),
         "NAME": os.environ.get("POSTGRES_DB"),
-        "HOST": os.environ.get("DJANGO_POSTGRES_HOST"),
+        "HOST": os.environ.get("DJANGO_POSTGRES_HOST", ""),
         "PORT": 5432,
     }
 }


website/thaliawebsite/settings/production.py
```diff
@@ -35,9 +35,9 @@
     "default": {
         "ENGINE": "django.db.backends.postgresql",
         "USER": os.environ.get("POSTGRES_USER", "postgres"),
-        "PASSWORD": os.environ.get("POSTGRES_PASSWORD", ""),
+        "PASSWORD": os.environ.get("POSTGRES_PASSWORD", None),
         "NAME": os.environ.get("POSTGRES_DB"),
-        "HOST": os.environ.get("DJANGO_POSTGRES_HOST"),
+        "HOST": os.environ.get("DJANGO_POSTGRES_HOST", ""),
```
joren485, line 40: This could be set to `postgres` which is a reasonable assumption for the name of a Postgres Docker host. However an empty string is fine too, given that that is the [default](https://docs.djangoproject.com/en/dev/ref/settings/#std:setting-HOST) anyway.

website/thaliawebsite/settings/production.py
```diff
@@ -35,9 +35,9 @@
     "default": {
         "ENGINE": "django.db.backends.postgresql",
         "USER": os.environ.get("POSTGRES_USER", "postgres"),
-        "PASSWORD": os.environ.get("POSTGRES_PASSWORD", ""),
+        "PASSWORD": os.environ.get("POSTGRES_PASSWORD", None),
```
joren485, line 38: What is the difference between `""` and `None`? Does `""` mean an empty password and `None` mean no password is send at all?

website/thaliawebsite/settings/production.py
pingiun: I specifically choose the empty host because of this part
>If you’re using PostgreSQL, by default (empty HOST), the connection to the database is done through UNIX domain sockets (‘local’ lines in pg_hba.conf).

This means that local authorization is easier when not running on docker. I think this is a good default, as the name of the postgres host in docker systems can differ.

website/thaliawebsite/settings/production.py
pingiun: This is indeed the case

joren485 (approved):

Legosteen11 (approved):

PR #1347, opened 2020-10-21
TESTING: Warn reviewer about migrations

pingiun:
I'm testing a way to check for migrations in the changes


pingiun:
image
This is what the comment looked like previously when I had a migration

se-bastiaan:

PR #1353, opened 2020-10-25
Add OAUTH_SCHEME setting

se-bastiaan:
### Summary

Add OAUTH_SCHEME setting that sets a scheme for OAuth that the app can use.

### How to test
1. Check that you can also add a redirect url with this scheme

### DIFF
--- a/website/thaliawebsite/settings/settings.py
+++ b/website/thaliawebsite/settings/settings.py
@@ -156,11 +156,11 @@ CORS_ORIGIN_ALLOW_ALL = True
 CORS_URLS_REGEX = r"^/(?:api|user/oauth)/.*"
 
 # OAuth configuration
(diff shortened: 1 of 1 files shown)


se-bastiaan:
Probably don't need this after all.

se-bastiaan:
Jk, we do.

website/thaliawebsite/settings/settings.py
```diff
@@ -156,8 +156,11 @@
 CORS_URLS_REGEX = r"^/(?:api|user/oauth)/.*"
 
 # OAuth configuration
+OAUTH_SCHEME = os.getenv("OAUTH_SCHEME", "nu.thalia")
```
pingiun, line 159: Maybe rename this to APP_OAUTH_SCHEME to signify it's intended usage

se-bastiaan: done

pingiun (approved):

PR #1359, opened 2020-10-26
Fix payment button

JobDoesburg:
### Summary
Fix payment button to actually work (remove unnecessary script, fix the javascript)

### How to test
Try to pay for an event


se-bastiaan (changes requested):
Screenshot 2020-10-26 at 23 15 11

nope

se-bastiaan (approved):

PR #1377, opened 2020-11-07
Set template settings in one single place

se-bastiaan:
### Summary
Set template settings in one single place

### How to test
1. Cannot really test this except for checking that it works without debug mode

### DIFF
--- a/website/thaliawebsite/settings/settings.py
+++ b/website/thaliawebsite/settings/settings.py
@@ -33,10 +33,7 @@ SECRET_KEY = "#o-0d1q5&^&06tn@8pr1f(n3$crafd++^%sacao7hj*ea@c)^t"
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 INTERNAL_IPS = ["127.0.0.1"]
-ALLOWED_HOSTS = ["*"]
-
-if not DEBUG:  # Django 1.10.3 security release changed behaviour
-    ALLOWED_HOSTS = ["*"]
+ALLOWED_HOSTS = ["*"] if DEBUG else []
 
 SITE_DOMAIN = os.environ.get("SITE_DOMAIN", "thalia.localhost")
 BASE_URL = f"https://{SITE_DOMAIN}"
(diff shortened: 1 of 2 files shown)


website/thaliawebsite/settings/settings.py
```diff
@@ -247,6 +256,7 @@
 
 # Compressor settings
 COMPRESS_ENABLED = True
+COMPRESS_OFFLINE = not DEBUG
```
pingiun, line 259: This line can be removed as it's the default value

website/thaliawebsite/settings/settings.py
```diff
@@ -130,11 +127,23 @@
...
                 "thaliawebsite.context_processors.thumbnail_sizes",
                 "thaliawebsite.context_processors.lustrum_styling",
-            ],
+            ]
         },
     },
 ]
 
+if not DEBUG:
+    del TEMPLATES[0]["APP_DIRS"]
```
pingiun, line 136: I think it's better to do something like:

```python
templates_base = {dict with base options}
templates_debug = {dict with debug options}
templates_production = {dict with production options}

if debug:
     TEMPLATES = [ { **templates_base, **templates_debug } ]
else:
    TEMPLATES = [ { **templates_base, **templates_production } ]
```

website/thaliawebsite/settings/settings.py
```diff
@@ -33,10 +33,7 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
```
pingiun, line 34: Isn't `DEBUG` true in this whole file because it's defined here?

pingiun (changes requested):

website/thaliawebsite/settings/settings.py
se-bastiaan: That is not true, the default value is `False`: https://django-compressor.readthedocs.io/en/stable/settings/#django.conf.settings.COMPRESS_OFFLINE

website/thaliawebsite/settings/settings.py
se-bastiaan: yes that could be what happens

se-bastiaan:
Think we should create an issue to do this better

PR #1379, opened 2020-11-08
Introduce dependabot for Github Actions

pingiun:
### Summary
Introduce dependabot for Github Actions

This will create pull requests every monday that look like this: https://github.com/pingiun/concrexit-nix/pull/2


se-bastiaan:
I believe that renovate can also do this: https://github.com/renovatebot/renovate/pull/7434

PR #1380, opened 2020-11-08
Refactor batch processing

### LINKED ISSUE #1364: BankAccount last used should be updated to withdrawal date not processing date.
### Describe the bug
We keep the last date a mandate was used so mandates that have not been used for more than <s>a year</s> three years can be revoked (because that are the sepa rules). However the ‘last used’ date is actually the indicated withdrawal date and not the processing date of the batch

### Expected behaviour
Update with the withdrawal date

JobDoesburg:
Closes #1364 

### Summary
If batch is processed, the withdrawal date is used to update the bankaccount 'last used' instead of the current date. Also contains a little refactor to turn batch processing into a service

### How to test
1. Process a payment batch
2. Check that emails are still being sent, the batch becomes 'processed' and the bank account 'last_used' date is the withdrawal date of the batch

### DIFF
--- a/website/payments/admin_views.py
+++ b/website/payments/admin_views.py
@@ -85,17 +85,7 @@ class BatchProcessAdminView(View):
                 request, _("{} already processed.").format(model_ngettext(batch, 1))
             )
         else:
-            batch.processed = True
-            payments = batch.payments_set.select_related("paid_by")
-            for payment in payments:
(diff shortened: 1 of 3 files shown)


se-bastiaan (changes requested):
I am pretty sure this should also change some tests

se-bastiaan (approved):

lukovdm (approved):

PR #1382, opened 2020-11-08
Add OpenAPI scheme generation and Swagger UI

se-bastiaan:
### Summary

Add OpenAPI scheme generation and Swagger UI.

This uses Django Rest Framework's [built-in capability](https://www.django-rest-framework.org/api-guide/schemas/) to generate a [OpenAPI 3.0](https://swagger.io/specification/) specification of our website's API.

It also includes the latest version of Swagger UI, which has the ability to read the generated schema and use it directly from the browser. The PR also extends the basic built-in functionality to add oauth2 support to the docs, which makes it easy to test and try the API with OAuth scopes.

Most API endpoints are available, the event registration is missing an option for the information fields however. I'm afraid we can't do much about that currently.

### How to test
1. Go to `/api/docs`
2. Check that most endpoints are available
3. Try a couple of endpoints


9Y0:
`AssertionError: Using OpenAPIRenderer, but 'pyyaml' is not installed.` is what I get when I run `make run` on your branch

9Y0:
The styling looks a bit weird for some buttons (e.g. `Download` button in image) and the `favicon.ico` is also missing. Not sure if you actually care about it but I wanted to make sure that you know about it.

Weird styling

se-bastiaan:
I did not change the CSS of Swagger except for the icon and topbar. So that is supposed to look like this, it is a direct copy of the files from their latest release: https://github.com/swagger-api/swagger-ui/releases/tag/v3.36.2

Purposefully not included the favicon because it is not really necessary

9Y0 (approved):

PR #1394, opened 2020-11-09
Added API endpoints for slides

### LINKED ISSUE #1318: API endpoints for slide models
We want to build the website sliders in the app. For that we need to be able to get the website sliders from the API.

There should be an API feature such that the app is able to get the current website sliders.

9Y0:
Closes #1318 

### Summary
The API now has support for slide, both event slides and global slides

### How to test
1. Add a slide (don't fill in the dates such that it isn't a global slide)
2. Add the slide to an event
3. Go to `/api/v1/events/{id}` and see that the slide is returned
4. Go to `/api/v1/announcements/slides`
5. See that the slide isn't there
6. Add a new slide and now fill in the `Display since`
7. Go to `/api/v1/announcements/slides` and see that the slide is there

### DIFF
--- a/website/announcements/api/serializers.py
+++ b/website/announcements/api/serializers.py
@@ -1,0 +1,20 @@
+"""DRF serializers defined by the announcements package."""
+from rest_framework import serializers
+
+from announcements.models import Slide
+
+
+class SlideSerializer(serializers.ModelSerializer):
+    """MemberGroup serializer."""
+
+    class Meta:
+        """Meta class for the serializer."""
+
+        model = Slide
+        fields = (
+            "pk",
+            "title",
+            "content",
+            "order",
+            "url",
+        )
--- a/website/announcements/api/serializers.py
+++ b/website/announcements/api/serializers.py
@@ -2,6 +2,8 @@
 from rest_framework import serializers
 
 from announcements.models import Slide
+from thaliawebsite.settings import settings
+from utils.media.services import get_thumbnail_url
 
 
 class SlideSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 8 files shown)


website/announcements/api/urls.py
```diff
@@ -0,0 +1,8 @@
+"""DRF routes defined by the announcements package."""
+from rest_framework import routers
+
+from announcements.api import viewsets
+
+router = routers.SimpleRouter()
+router.register(r"announcements/slide", viewsets.SlideViewset)
```
se-bastiaan, line 7:
```suggestion
router.register(r"announcements/slides", viewsets.SlideViewset)
```

website/events/api/serializers/events/retrieve.py
```diff
@@ -51,6 +53,7 @@ class Meta:
     is_pizza_event = serializers.SerializerMethodField("_is_pizza_event")
     google_maps_url = serializers.SerializerMethodField("_google_maps_url")
     is_admin = serializers.SerializerMethodField("_is_admin")
+    slide = SlideSerializer()
```
se-bastiaan, line 56: Nice :D

website/announcements/api/serializers.py
```diff
@@ -0,0 +1,20 @@
...
+    """Slide serializer."""
+
+    class Meta:
+        """Meta class for the serializer."""
+
+        model = Slide
+        fields = (
+            "pk",
+            "title",
+            "content",
```
se-bastiaan, line 17: This currently return the full size image, which is probably not what we want because it could be _too_ large.

The slides on the website are resized according to the SLIDE image size setting, the members serializer should have a way to get the right url.

se-bastiaan (approved):

PR #1403, opened 2020-11-14
Concrexit API v2

se-bastiaan:
### Summary

Not sure if this should be a refactor or a new feature. But I found that the following is the case:
1. We use Django Rest Framework for our API and every implementation we have uses [viewsets](https://www.django-rest-framework.org/api-guide/viewsets/). This is a way to tightly couple the different views and can work wonderful for the [easy cases](https://github.com/svthalia/concrexit/blob/master/website/activemembers/api/viewsets.py). But [in](https://github.com/svthalia/concrexit/blob/master/website/pizzas/api/viewsets.py#L51) [most](https://github.com/svthalia/concrexit/blob/master/website/members/api/viewsets.py#L41) [situations](https://github.com/svthalia/concrexit/blob/master/website/events/api/viewsets/events.py#L49) it only brings extra checks in our code to accommodate different situations. Especially in cases where there are permissions involved.
2. We used [Django token authentication](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) but have now also authentication using OAuth 2. This last implementation should be a first-class citizen in our API but most of the APIs do not have scoping available. Moreover, they are not readable because the API code _expects_ a user.

To be able to solve this I'd like to introduce the Concrexit API v2. This PR is here for comments, what you think about this and maybe to make some design choices.

Thus:
1. The implementation of the API views should happen using [class-based views](https://www.django-rest-framework.org/tutorial/3-class-based-views/), just like most views in the (newer) apps. This is much more similar to the normal views and is thus easier to grasp if you have to write a view for both users (HTML) and applications (API). No more trying to understand viewsets to try to get them to do what you want. This PR contains an example of such APIs
2. Implement scopes and Django permissions for all views. No longer allow TokenAuthentication for the API. This is going to require a custom implementation that checks both the user permissions and the client scopes. A client logging in with client_credentials will get access as read-only anonymous.
3. Throttle requests based on the amount made per day because the API is basically public due to the awesome Swagger documentation (this also applies to API v1)
4. Pagination should always be enabled by default because some views are really slow now (i.e. https://thalia.nu/api/v1/photos/albums/)
5. It is probably smart to split the regular user APIs from the admin APIs. So for organisers that are changing the payments for events or pizzas for example, this will increase duplicated code but reduce API view and serializer complexity. The same endpoints are currently used for both normal users and for admin changes.

Please let me now what you think.

### How to test
1. Go to `/api/docs` and select the v2 documentation
2. Yay


Legosteen11:
deploy review

Legosteen11:
Good idea Sebastiaan, I think having correct scopes and rate limiting on all our endpoints is important. This upgrade would make the Concrexit API more accessible for developers (of which we have a lot at Thalia of course :smile: ). Separating user and admin endpoints also makes it clear that some endpoints are not for 3rd party applications, while still allowing us to benefit from better documentation (e.g. Swagger, which I find really awesome), easier API implementation and scopes.

Maybe it would be a good idea to document what APIs are in use tomorrow evening? Then we can decide which ones we want to add to v2 and/or make admin-only?

se-bastiaan:
We need to implement all the APIs that we currently have, we only have APIs currently for things that are used in the app ;)

se-bastiaan:
deploy remove

se-bastiaan:
remove review

PR #1410, opened 2020-11-18
Improvements to payments in the events admin

### LINKED ISSUE #1406: Setting payment to 'none' in events admin does not work
### Describe the bug
Setting payment to 'none' in events admin does not work

### How to reproduce
0. Make sure your user is superuser
1. Create an event with registrations
2. Register a payment for an event registration
3. Try to remove the payment again
4. You are not authorized!

### Expected behaviour
It should show an more informative error message

JobDoesburg:
Closes #1406 and some minor other tweaks

### Summary
Payments are now hidden when no payment is required, and the payment amount is explicitly shown when paying. Also error messages are improved

### How to test
1. Check an events admin for free events
2. Check an events admin for paid events

### DIFF
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -195,6 +195,10 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
     def registration_required(self):
         return bool(self.registration_start) or bool(self.registration_end)
 
+    @property
+    def payment_required(self):
+        return not (self.price == 0)
+
     def has_fields(self):
         return self.registrationinformationfield_set.count() > 0
 
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -197,7 +197,7 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
 
     @property
     def payment_required(self):
-        return not (self.price == 0)
+        return self.price != 0
 
     def has_fields(self):
         return self.registrationinformationfield_set.count() > 0
--- a/website/events/models/event_registration.py
+++ b/website/events/models/event_registration.py
@@ -141,11 +141,11 @@ class EventRegistration(models.Model, Payable):
 
     @property
     def payment_topic(self):
-        return f'{self.event.title} [{date(self.event.start, "Y-m-d")}]'
+        return f'{self.event.title_en} [{date(self.event.start, "Y-m-d")}]'
 
     @property
     def payment_notes(self):
-        notes = f"Event registration {self.event.title}. "
+        notes = f"Event registration {self.event.title_en}. "
         notes += f"{date(self.event.start)}. Registration date: {date(self.date)}."
         return notes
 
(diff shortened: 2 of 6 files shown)


website/events/models/event.py
```diff
@@ -195,6 +195,10 @@ def registration_started(self):
     def registration_required(self):
         return bool(self.registration_start) or bool(self.registration_end)
 
+    @property
+    def payment_required(self):
+        return not (self.price == 0)
```
se-bastiaan, line 200: ????

```suggestion
        return self.price != 0

File: website/events/models/event_registration.py [hunk 2]
```diff
@@ -141,11 +141,11 @@ def payment_amount(self):
 
     @property
     def payment_topic(self):
-        return f'{self.event.title_en} [{date(self.event.start, "Y-m-d")}]'
```
se-bastiaan, line 144: I do not believe the fields of the events have been renamed yet.

website/events/templates/events/admin/registrations_table.html
```diff
@@ -39,23 +41,24 @@
...
+                {% if event.payment_required %}
                 <td>
                     <select name="payment"
                             data-none="no_payment"
                             {% if registration.payment %}class="paid"{% endif %}>
                         <option value="no_payment"
                                 {% if not registration.payment %}selected{% endif %}>{% trans 'None' %}</option>
                         <option value="{{ payment.CARD }}"
-                                {% if registration.payment and registration.payment.type == payment.CARD %}selected{% endif %}>{% trans 'Card' %}</option>
+                                {% if registration.payment and registration.payment.type == payment.CARD %}selected{% endif %}>{% trans 'Card' %} (€ {% if registration.payment %}{{ registration.payment.amount }}{% else %}{{ event.price }}{% endif %})</option>
```
se-bastiaan, line 52: Why?

se-bastiaan (changes requested):

website/events/models/event.py
se-bastiaan: Screenshot 2020-11-24 at 20 45 34

website/events/templates/events/admin/registrations_table.html
JobDoesburg: I always felt like it is weird that you don't see the amount when selecting that something is paid. I think it is a nice confirmation.

website/events/models/event.py
JobDoesburg: cannot reproduce this.....

website/events/templates/events/admin/registrations_table.html
se-bastiaan: But the price is the same for everyone? The event organiser knows exactly what the price should be.

Moreover, this adds more text to the event admin page that is not really necessary. I haven't checked mobile but it may be forcing other things out of view.

JobDoesburg: Hmm you're right about the problem on mobile

se-bastiaan (approved):

Issue #992, opened 2020-03-18
Internet required to run tests


thaliawww-service:
In GitLab by @joren485 on Jan 26, 2020, 20:07

### Description
Internet required to run tests, because [`activemembers/signals.py` always initializes the G Suite API](https://gitlab.science.ru.nl/thalia/concrexit/blob/master/website/activemembers/signals.py#L15). This is not (and should not be) always necessary (e.g. when running tests).

thaliawww-service:
In GitLab by @se-bastiaan on Jan 26, 2020, 21:31

mmm it's better to initialise it globally because of the caching, but we should be able to disable it in test environments

thaliawww-service:
In GitLab by @joren485 on Jan 26, 2020, 22:07

This problem does not only occur with the tests, but also when running a development environment (i.e. `python website/manage.py runserver` does not work without internet).

se-bastiaan:
We can borrow code to do this from https://github.com/GipHouse/website as soon as @lukovdm is done there :p

PR #1411, opened 2020-11-19
Deploy to staging via Nix

pingiun:
### Summary
Deploy to staging via Nix. I think this PR depends on a better fix to #992 because Nix doesn't allow internet access when building. It seems that networkless building still fails.

### How to test
If you are a trusted user on the staging machine, you can run
```bash
scripts/send-conf-and-switch.sh
```
to apply the latest configuration to the staging NixOS server


nix/configuration.nix
```diff
@@ -0,0 +1,307 @@
...
+      };
+    };
+
+    # These configuration options are defined and implemented by the timed-commands.nix file
+    concrexit-timers = {
+      send_scheduled_messages = {
+        every = 60 * 5;
+        description = "Send scheduled push notifications";
+      };
+      sendplannednewsletters.every = 60 * 60;
```
Legosteen11, line 211: I think we should add some more descriptions

nix/configuration.nix
```diff
@@ -0,0 +1,307 @@
...
+    environment.systemPackages = [ pkgs.concrexit-manage ];
+    networking.hostName = "staging";
+
+    security.acme.email = "www@thalia.nu";
+    security.acme.acceptTerms = true;
+
+    nix = {
+      gc.automatic = true;
+      # This is required to be able to use nix-copy-closure
+      trustedUsers = [ "root" "deploy" "jelle" ];
```
Legosteen11, line 121: I will add myself as well :wink:

nix/configuration.nix
```diff
@@ -0,0 +1,307 @@
...
+    services = {
+      nginx = {
+        enable = true;
+
+        recommendedGzipSettings = true;
+        recommendedOptimisation = true;
+        recommendedTlsSettings = true;
+
+        # Reloads NGINX instead of restarting it whenever possible
+        enableReload = true;
```
Legosteen11, line 232: Let's just always restart nginx

nix/configuration.nix
```diff
@@ -0,0 +1,307 @@
...
+              extraConfig = securityHeaders;
+            };
+          in
+          {
+            "${cfg.domain}" = {
+              # Request a certificate from letsencrypt
+              enableACME = cfg.ssl;
+              # Enable redirects to https
+              forceSSL = cfg.ssl;
+              default = !cfg.ssl;
```
Legosteen11, line 250: ???

nix/configuration.nix
```diff
@@ -0,0 +1,307 @@
...
+              '';
+            };
+            "www.${cfg.domain}" = {
+              enableACME = cfg.ssl;
+              locations."/".return = "301 https://${cfg.domain}$request_uri";
+              extraConfig = securityHeaders;
+            };
+            "pizza.${cfg.domain}" = pizzaConfig;
+            "pasta.${cfg.domain}" = pizzaConfig;
+            "xn--vi8h.${cfg.domain}" = pizzaConfig;
```
Legosteen11, line 278: Add pasta emoji (xn--3i8h, 🍝)

resources/maintenance.html
```diff
@@ -0,0 +1,38 @@
...
+	.st0{fill:#E62272;}
+</style>
+<meta name="viewport" content="width=device-width, initial-scale=1">
+</head>
+<body>
+<article>
+<svg version="1.1" id="image" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="-233 408.5 374 375" style="enable-background:new -233 408.5 374 375;" xml:space="preserve">
+<g>
+	<path id="hoekje_onder_3_" class="st0" d="M25.3,712.7c-4.4,0,89.8,10.9,91.2,10.9l-51.6,40.1L25.3,712.7L25.3,712.7z"/>
```
Legosteen11, line 22: Panic!!! We have introduced more Dutch language on the concrexit site 😱😱😱🚨🚨🚨

Legosteen11 (changes requested):
Looks good, I will improve it a bit

resources/maintenance.html
pingiun: Lol

nix/configuration.nix
Legosteen11: I should add an issue to ask someone to explain what these management commands do

nix/configuration.nix
```diff
@@ -76,6 +76,11 @@ in
       default = true;
     };
 
+    concrexit.rejectUnknownHost = mkOption {
+      type = types.bool;
+      default = cfg.ssl;
```
pingiun, line 81: If this is the default, the option should be named `allowUnknownHost` or the logic should be reversed. Just run through it and you'll see what I mean

nix/configuration.nix
pingiun: Yes good idea

nix/configuration.nix
pingiun: Also I recommend to add a description to this option

Legosteen11 (approved):

PR #1425, opened 2020-11-23
Event guest registration

### LINKED ISSUE #1345: Can't set registration information fields from admin (for non-member registrations)
### Describe the bug
It is not possible to manually set the event registration information fields from the admin. This is quite problematic when adding non-member's registrations

### How to reproduce
1. add a manual event registration
2. try to edit the information fields (this works)
3. you cannot save it

### Expected behaviour
It should save

mtsatsev:
Closes #1345

### Summary
Organizers of events should be able to edit the members' fields at their events even for non-members. Now they can.

### How to test
1. Add a manual event registration.
2. Edit the information fields of the participants.
3. It works.

### DIFF
--- a/website/events/migrations/0001_initial.py
+++ b/website/events/migrations/0001_initial.py
@@ -7,7 +7,7 @@ from django.db import migrations, models
 import django.db.models.deletion
 
 
-class Migration(migrations.Migration):
+class Migration(migrations.Migration): 
 
     initial = True
 
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -1,4 +1,4 @@
-    from collections import OrderedDict
+from collections import OrderedDict
 
 from django.utils import timezone
 from django.utils.datetime_safe import date
@@ -192,7 +192,8 @@ def update_registration(
 
     if (
         (not event_permissions(member, event, name)["update_registration"]
-        or not field_values) and is_organiser(member, event)
+        or not field_values)
+        and is_organiser(member, event)
     ):
         return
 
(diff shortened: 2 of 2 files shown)


website/events/migrations/0001_initial.py
```diff
@@ -7,7 +7,7 @@
 import django.db.models.deletion
 
 
-class Migration(migrations.Migration):
+class Migration(migrations.Migration):
```
9Y0, line 10: I'm not sure why you edited this but you should definitely not commit it

website/events/services.py
```diff
@@ -191,8 +191,9 @@ def update_registration(
         name = registration.name
 
     if (
-        not event_permissions(member, event, name)["update_registration"]
-        or not field_values
+        (not event_permissions(member, event, name)["update_registration"]
+        or not field_values)
+        and is_organiser(member, event)
```
9Y0, line 196: This if statement will continue if `field_values` is `None` and `member` isn't an organizer causing some tests to fail

9Y0 (changes requested):
In general, try to run the tests (`make tests`) and format your code (`make fmt`) before you push it, it'll save you a lot of time.

website/events/migrations/0001_initial.py
mtsatsev: I am not sure why that was changed tbh.

se-bastiaan:
Closing in favour of https://github.com/svthalia/concrexit/pull/1512

PR #1426, opened 2020-11-23
Disallow zero euro payments

### LINKED ISSUE #1365: Event registrations and €0 payments
### Describe the change
Currently, free events still display a payment box for each registrations (and possibly, it is even allowed to create €0 payments, I haven't looked at the code yet). This should be prevented. A payment of €0 should not be able to exist and the admin should not display the possibility to create such a payment.

### Motivation
Undesired behaviour

### Current implementation
Not sure, havent looked at the code yet.

### Suggested implementation
The things below (some might already be implemented though)

- Payment amount should have a database constraint of > 0
- `create_payment` should not accept payables with an amount of 0
- The admin widget should not appear if the payment is €0

JobDoesburg:
Closes #1365 

### Summary
Zero euro payments cannot exist anymore

### How to test
1. Check out all the ways of accessing/creating payments from the changed files

### DIFF
--- a/website/payments/models.py
+++ b/website/payments/models.py
@@ -136,6 +136,8 @@ class Payment(models.Model):
         super().save(**kwargs)
 
     def clean(self):
+        if self.amount == 0:
+            raise ValidationError({"amount": _(f"Payments cannot be €{self.amount}")})
         if self.type != Payment.TPAY and self.batch is not None:
             raise ValidationError(
                 {"batch": _("Non Thalia Pay payments cannot be added to a batch")}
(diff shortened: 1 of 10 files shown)


website/payments/models.py
```diff
@@ -136,6 +136,8 @@ def save(self, **kwargs):
         super().save(**kwargs)
 
     def clean(self):
+        if self.amount == Decimal(0):
```
se-bastiaan, line 139: `self.amount == 0` does not work?

JobDoesburg: Hmm yeah I don't know why I did that

PR #1427, opened 2020-11-26
Better names for Make commands

se-bastiaan:
### Summary
Better names for Make commands that are not 'make create' or 'make make' which makes (pun intended) no sense.

### How to test
1. Yay


pingiun (changes requested):
You should reflect these changes in the README

pingiun:
Why was this closed?

se-bastiaan:
Bit of cleanup in the overview, still need to fix this.

PR #1429, opened 2020-11-27
Move API v1 to subpackages

se-bastiaan:
### Summary
This PR moves all existing code that we have for API endpoints to v1 packages inside the Django apps.
It also introduces basic functionality for API v2 that we do not want introduced into v1 and vice versa. Examples are the pagination and authentication tokens.

### How to test
1. Check that all APIs are still available in the docs `/api/docs`
2. Test some of the APIs
3. Profit


9Y0 (changes requested):
Going to `/api/docs` results in an error: `Reverse for 'schema-v2' not found. 'schema-v2' is not a valid view function or pattern name.`

pingiun:
deploy review

pingiun (approved):

9Y0:
Why exactly is this one merged? `/api/docs` still crashes, also on the review environment

PR #1432, opened 2020-11-30
Prevent registrations table from not showing true checkbox value

### LINKED ISSUE #1400: Event checkboxes show no value if False
### Describe the bug
In the backend of the site, in the list of registered members, if there is a checkbox and members do not check them, they will show as empty in the back-end. This is very confusing, because when members click the after-registration page away without filling in any information, it will show as empty as well!

### How to reproduce
For example, for the [Digital Beer Tasting](https://thalia.nu/admin/events/event/679/details/), as you can see, some ```Delivery``` fields are empty. If you export this data to a ```.csv```, you can see that some of them are actually ```FALSE```/```ONWAAR```, and some are just empty.

### Expected behaviour
It is clear whether a checkbox is ```FALSE``` or ```null```

PeaceDucko:
Closes #1400 

### Summary
Prevents the _Registrations_ tab in admin for a specific event from hiding the true value of a checkbox.

**Note:** even though it is checked if a field of the type _checkbox_ is being shown in the _registrations_table.html_ file, the actual type of the checkbox is actually a _boolean_. Check if this should actually be the case or whether it has to be refactored.

### How to test

1. Register yourself (or another member) for an event that includes an optional checkbox.
2. Register directly (without checking the checkbox) **or** click out of the tab entirely without completing the registration.
3. Head over to the admin panel and click on 'Events' -> the event you registered for.
4. Check if the value of the checkbox you filled in is _None_ (when you closed the tab) or _False_ (if you completed your registration but did not check the checkbox).

### DIFF
--- a/website/events/templates/events/admin/registrations_table.html
+++ b/website/events/templates/events/admin/registrations_table.html
@@ -62,7 +62,7 @@
                 <td data-sortval="{{ registration.date_cancelled|date:'c' }}">{{ registration.date_cancelled }}</td>
                 {% endif %}
                 {% for field in registration.information_fields %}
-                    {% if not field.value %}
+                    {% if not field.value and not field.field.type == 'boolean' %}
                     <td></td>
                     {% elif field.field.type == 'checkbox' %}
                     <td>{{ field.value|yesno }}</td>
--- a/website/events/templates/events/admin/registrations_table.html
+++ b/website/events/templates/events/admin/registrations_table.html
@@ -64,7 +64,7 @@
                 {% for field in registration.information_fields %}
                     {% if not field.value and not field.field.type == 'boolean' %}
                     <td></td>
-                    {% elif field.field.type == 'checkbox' %}
+                    {% elif field.field.type == 'boolean' %}
                     <td>{{ field.value|yesno }}</td>
                     {% else %}
                     <td>{{ field.value }}</td>


website/events/templates/events/admin/registrations_table.html
```diff
@@ -62,7 +62,7 @@
                 <td data-sortval="{{ registration.date_cancelled|date:'c' }}">{{ registration.date_cancelled }}</td>
                 {% endif %}
                 {% for field in registration.information_fields %}
-                    {% if not field.value %}
+                    {% if not field.value and not field.field.type == 'boolean' %}
```
se-bastiaan, line 65: I don't believe this is the change Thijs asked for. The thing is that the checkbox is also false when no value was entered. But we don't have a 'boolean' field type.

se-bastiaan (changes requested):
.

website/events/templates/events/admin/registrations_table.html
PeaceDucko: > But we don't have a 'boolean' field type.
@se-bastiaan I first thought this was the case but the type of, what we assume is, 'checkbox' is actually of the type 'boolean' (try printing field.field.type). 

As for the change itself. I checked the root of this issue with a couple of other Technicie members from which I concluded that the way checkboxes are displayed in the CSV (e.g. False or null) should be mirrored in the admin panel. The problem (as described by Thijs) in my opinion asks for clarity in the admin panel whether a checkbox is False or null in stead of just showing an empty field.

If there is still any confusion feel free to PM me.

se-bastiaan: I assumed based on [the line below](https://github.com/svthalia/concrexit/pull/1432/files#diff-0ffb59d377093550828358c44ac02d869263ae11f7c93e2b8d819c1baae6892bR67) that is was supposed to be checkbox :p 

Now that I have checked an actual event the value is indeed always true or false.

Maybe include that line in the changes?

PeaceDucko: Yeah it's confusing, trust me I've bumped my head over that a couple of times :)

I'll update the changes!

PeaceDucko: Yeah trust me, I've bumped my head over that issue a couple of times :)

I'll update the changes to the line to mentioned, changing the value from 'checkbox' to 'boolean' (if that's what you mean).

se-bastiaan: That's exactly what I meant :)

se-bastiaan (approved):

PR #1444, opened 2020-12-07
Show more events statistics in the events admin

JobDoesburg:
### Summary
With this new features, it becomes easy for the board to see whether events have still unpaid registrations

### How to test
Check out the events admin. It should now contain numbers for the amount of registrations that are (un)paid


website/events/admin.py
```diff
@@ -143,6 +145,29 @@ class EventAdmin(DoNextTranslatedModelAdmin):
...
+
+        count_present_registrations = (
+            obj.present_registrations.count() if obj.present_registrations else 0
+        )
+        count_unpaid_present_registrations = (
+            obj.unpaid_present_registrations.count()
+            if obj.unpaid_present_registrations
+            else 0
+        )
+        count_participants = obj.participants.count() if obj.participants else 0
```
se-bastiaan, line 165: These are highly non-optimised queries my friend.

website/events/templates/events/admin/registrations_table.html
```diff
@@ -49,13 +52,13 @@
                                 {% if registration.payment and registration.payment.type == payment.CARD %}selected{% endif %}>{% trans 'Card' %}</option>
                         <option value="{{ payment.CASH }}"
                                 {% if registration.payment and registration.payment.type == payment.CASH %}selected{% endif %}>{% trans 'Cash' %}</option>
-                        <option value="{{ payment.TPAY }}"
-                                {% if not registration.member.tpay_enabled %} disabled {% endif %}
```
se-bastiaan, line 53: This should be disabled right?

se-bastiaan:
From your description I am not sure what to test exactly.

JobDoesburg:
Hmm I see something weird has happened with git (because I added a new commit to a branch that has been squashed and merged). I will fix it next week...

website/events/templates/events/admin/registrations_table.html
JobDoesburg: My git has been fixed again :)

website/events/admin.py
JobDoesburg: Is it? It is a count on a (filtered) set of registrations...

JobDoesburg:
Don't mind my git history 🙂

website/events/admin.py
lukovdm: you can look at the bug/speeduppaymentusers branch for inspiration

website/events/models/event.py
```diff
@@ -247,6 +247,41 @@ def cancellation_allowed(self):
             and self.registration_start <= now < self.start
         )
 
+    @property
+    def present_registrations(self):
+        return self.participants.filter(present=True)
+
+    @property
+    def unpresent_registrations(self):
```
se-bastiaan, line 255: This is never used?

website/events/models/event.py
```diff
@@ -247,6 +247,41 @@ def cancellation_allowed(self):
...
+        return self.participants.filter(present=False)
+
+    @property
+    def unpaid_registrations(self):
```
se-bastiaan, line 259: This is only used in `everything_paid` I believe, so it could be inside that function instead making the code more concise.

website/events/models/event.py
```diff
@@ -247,6 +247,41 @@ def cancellation_allowed(self):
...
+    @property
+    def unpaid_present_registrations(self):
+        return (
+            self.present_registrations.filter(payment__isnull=True)
+            if self.payment_required
+            else None
+        )
+
+    @property
+    def everything_paid(self):
```
se-bastiaan, line 275:
```suggestion
    def everyone_paid(self):
```

It's about everyone who is participating right?

se-bastiaan (changes requested):
I think this still requires changes, so just putting out this status here.

website/events/models/event.py
JobDoesburg: I think this is better as it separates the different logic steps better

website/events/admin.py
JobDoesburg: Yeah this is nasty

PR #1445, opened 2020-12-07
Create sales app backend

### LINKED ISSUE #1386: Thalia Pay at borrels
With Thalia Pay being almost released, the next thing would be to also use it for buying beers and such at borrels.

### Possible solutions
I think the best way to do this is with some form of QR code. Members should scan the QR code with their favorite QR-scanner app and this will forward them to a site where they need to confirm (after logging in) their payment. A little brainstorm:

- Dynamic QR codes: every transaction requires a new QR code. This would result in a system where borrelcie members need to generate the QR code on their own device (not necessarily from concrexit, btw!), which will display the QR code on their screen and a member will then scan that code. The borrelcie member will see a confirmation on their own device after the payment has been accepted by the member. Possibly, we can also implement the 18+ check here!
    - pro: secure, nice to model
    - con: requires a device for borrelcie members, might be a bit of work in practice

- Static QR codes: every event / kind of payment will get a single QR code that users scan and confirm. The member then shows the payment-confirmation on their own device to the borrelcie.
    - pro: easy and simple
    - con: ugly, a little less secure (we could get payments of people not buying anything, which is not really bad but still)

There are different choices to make about for example, when to set the amount of beers someone orders. 

### What's next?
I think it would be nice to have a meeting with the borrelcie so they can think about their own preferred workflow, so we can consider that in further decisions. 

I could think of us building a separate system like borrel.thalia.nu that only functions as a borrel console, which uses concrexit for payments, but might be a bit overkill though as well.

JobDoesburg:
Closes #1386 and #1616

### Summary
A new app arrived to concrexit! This app provides a backend for Point of Sale orders. It features products and product lists that are used during Shifts. During a shift, orders can be placed. It should be all quite self-explanatory.

This system does **not** provide any real front end (except for maybe a rather clumsy Django admin), but instead, it exposes a rather specified API that will be the main way of use. The front end will be built in a different repository to be more versatile (both in programming language and authentication methods, depending on the specific use case).

Note that only shift admins will need to use this. Members paying for objects will use the OrderPaymentView to confirm their payments. The idea is that the url to this OrderPaymentView relies on the (unpredictable) UUID of an order (for security), and is shared with the user via a QR code. 

### How to test
Try to create all kind of orders, update them in all kinds of ways, pay them, et cetera. Just try out all the different things that we could want to do with the orders, both via the Django admin as the API!


website/sales/models/order.py
```diff
@@ -0,0 +1,146 @@
...
+    @property
+    def age_restricted(self):
+        return any(
+            self.orderitem_set.values_list(
+                "product__product__age_restricted", flat=True
+            )
+        )
+
+    @property
+    def total_amount(self):
```
lukovdm, line 74: This is a bit unclearly named

JobDoesburg:
This is just an initial version. Still todo:

- [ ] Resolve the in-code TODO's
- [ ] Prevent changing orders for closed shifts
- [ ] Implement API scoping and Django permissions / Shift permissions
- [ ] Add test coverage 100%


Also, the changes to the payments app that I made now are a bit messy. They were required because I wanted to keep using the existing code in the payments app for all payments. However, I think there are better solutions possible:

- we could add a ForeignKeyField with a Member to the order model. On visiting the OrderPaymentView, we could update this field to the accepting user and then we could keep using the existing payments codebase without any changes. The benefit is that this makes the whole app more versatile. The drawback is that it is again a new field.
- we could implement our own payment flow and bypass the payment_button from the payments app and possibly also the create_payment service. The benefit is that we can tweak the payment flow, the drawback is that it splits up the payment code

website/sales/api/v2/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Partners app API v2 urls."""
+
+urlpatterns = []
```
se-bastiaan, line 3: I believe this file can be empty

website/sales/migrations/0002_order_payer.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('members', '0039_remove_profile_language'),
+        ('sales', '0001_initial'),
+    ]
+
+    operations = [
```
se-bastiaan, line 14: Since this is a new app these migrations can be merged into 1.

website/sales/api/v2/urls.py
```diff
@@ -0,0 +1,15 @@
...
+
+from sales.api.v2.views import (
+    ShiftListView,
+    ShiftDetailView,
+    OrderListView,
+    OrderDetailView,
+)
+
+urlpatterns = [
+    path("sales/", ShiftListView.as_view(), name="shift-list"),
```
se-bastiaan, line 11: Something with your url definitions makes it impossible for the docs schema to be generated.

website/sales/api/v2/views.py
```diff
@@ -0,0 +1,63 @@
...
+    """Returns an overview of all sales shifts."""
+
+    serializer_class = ShiftSerializer
+    queryset = Shift.objects.all()
+    filter_backends = (
+        filters.OrderingFilter,
+        filters.SearchFilter,
+    )
+    ordering_fields = ("start_date", "end_date")
+    permission_classes = [IsAuthenticated]
```
se-bastiaan, line 29: This needs scope permissions and probably even the DjangoModelPermissions authenticator

website/sales/api/v2/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class ShiftDetailView(RetrieveAPIView):
+    serializer_class = ShiftSerializer
+    queryset = Shift.objects.all()
+    permission_classes = [IsAuthenticated]
```
se-bastiaan, line 35: This needs scope permissions and probably even the DjangoModelPermissions authenticator

website/sales/api/v2/views.py
```diff
@@ -0,0 +1,63 @@
...
+    serializer_class = OrderListSerializer
+    permission_classes = [
+        IsAuthenticatedOrTokenHasScope,
+        DjangoModelPermissions,
+    ]
+    required_scopes = ["sales:write"]
+
+    def get_queryset(self):
+        shift = get_object_or_404(Shift, pk=self.kwargs.get("pk"))
+        return Order.objects.filter(shift=shift.pk)
```
se-bastiaan, line 47: This does not seem to be the right spot to check if the shift exists to me. You are getting it for the context too and only using it's pk.

Maybe you should get it here, set it to`self.shift` and reuse it in get_serializer_context

website/sales/api/v2/urls.py
```diff
@@ -0,0 +1,15 @@
...
+    path("sales/<int:pk>/", ShiftDetailView.as_view(), name="shift-detail",),
+    path("sales/<int:pk>/orders/", OrderListView.as_view(), name="shift-orders",),
+    path("sales/order/<uuid:pk>/", OrderDetailView.as_view(), name="order-detail",),
```
se-bastiaan, line 14:
```suggestion
    path("sales/orders/<uuid:pk>/", OrderDetailView.as_view(), name="order-detail",),
```

All urls are supposed to be plural

website/sales/api/v2/urls.py
```diff
@@ -0,0 +1,15 @@
...
+    ShiftListView,
+    ShiftDetailView,
+    OrderListView,
+    OrderDetailView,
+)
+
+urlpatterns = [
+    path("sales/", ShiftListView.as_view(), name="shift-list"),
+    path("sales/<int:pk>/", ShiftDetailView.as_view(), name="shift-detail",),
+    path("sales/<int:pk>/orders/", OrderListView.as_view(), name="shift-orders",),
```
se-bastiaan, line 13:
```suggestion
    path("sales/shifts/<int:pk>/", ShiftDetailView.as_view(), name="shift-detail",),
    path("sales/shifts/<int:pk>/orders/", OrderListView.as_view(), name="shift-orders",),
```

The object needs to be in the URL. Otherwise you are not getting anything.

website/sales/models/product.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    product_list = models.ForeignKey(
+        ProductList,
+        related_name="product_items",
+        null=False,
+        blank=False,
+        on_delete=models.CASCADE,
+    )
+    price = models.DecimalField(
+        max_digits=6, decimal_places=2, validators=[MinValueValidator(Decimal("0.00"))]
```
se-bastiaan, line 54: Why is `Decimal` needed?

website/sales/api/v2/serializers/shift.py
```diff
@@ -0,0 +1,53 @@
...
+    total_revenue = serializers.DecimalField(
+        max_digits=6, decimal_places=2, read_only=True
+    )
+
+    num_orders = serializers.SerializerMethodField("_num_orders")
+
+    def _num_orders(self, instance):
+        return instance.orders.count()
+
+    orders = serializers.HyperlinkedIdentityField(view_name="shift-orders")
```
se-bastiaan, line 53: I am not at all planning to add hyperlinks in any v2 API. Because:
1. Everything is identified by a primary key
2. The documentation is available to everyone
Therefore I am not sure I would like to see this in this part of the API.

website/sales/api/v2/serializers/order.py
```diff
@@ -0,0 +1,177 @@
...
+        read_only_fields = (
+            "pk",
+            "created_at",
+            "payment_amount",
+            "payment",
+            "payer",
+            "details",
+        )
+
+    payer = serializers.HyperlinkedRelatedField(
```
se-bastiaan, line 171: This could just be a `MemberSerializer` to immediately include the information. You basically already know you are going to need it.

se-bastiaan: Maybe in a smaller form but that would be pretty easy to do

website/sales/admin.py
```diff
@@ -0,0 +1,210 @@
...
+
+@register(Order)
+class OrderAdmin(admin.ModelAdmin):
+    inlines = [
+        OrderItemInline,
+    ]
+
+    list_display = (
+        "id",
+        "shift",
```
se-bastiaan, line 81: This prints a really long value

se-bastiaan (changes requested):
Currently, it seems that the initial migration does not work sadly enough.

website/sales/models/order.py
```diff
@@ -0,0 +1,271 @@
...
+        if errors:
+            raise ValidationError(errors)
+
+    def get_age_restricted(self):
+        return any(
+            self.order_items.values_list("product__product__age_restricted", flat=True)
+        )
+
+    def get_subtotal(self):
+        qs = self.order_items.aggregate(subtotal=Coalesce(Sum("total"), Value(0.00)))
```
se-bastiaan, line 164: I was wondering, since you replace the default queryset manager by the one with the annotations do you not already have this information for every time you get the models? Since it is essentially already part of the queryset.

JobDoesburg: Yes and no. When you materialize an Order from a QuerySet, the shitty thing is that the annotations get lost. It is really ugly. For example, in the API I need to turn all those annotations into serializermethodfields, because when querying an order the annotations are present, but on updating them, the order is materialized to a python object and suddenly the annotated fields don't exist anymore. That's exactly what django-queryable-properties tries to fix (and does fix quite decently actually): it allows you to define python properties in terms of Django ORM calculus and that works fine (but the problem with that package is that you can't use the annotations without explicitly using `.select_properties(()` on the QuerySet).

The reason why I still can use the `order_items.aggregate` here is that `order_items` will return a new manager whatsoever.

website/payments/templates/payments/widgets/payment.html
```diff
@@ -5,7 +5,7 @@
     {% elif not payment and obj.payment_amount != 0 %}
         {% trans "Unpaid" %}
         -
-        € {{ obj.payment_amount }}
+        € {{ obj.payment_amount | floatformat:2}}
```
se-bastiaan, line 8:
```suggestion
        € {{ obj.payment_amount|floatformat:2}}
```

Not really necessary but I believe we format it like this everywhere.

website/sales/api/v2/serializers/order.py
```diff
@@ -0,0 +1,168 @@
...
+            "discount",  # TODO make this writable if user has custom_prices permission
+            "payment",
+            "total_amount",
+            "num_items",
+            "age_restricted",
+            "order_description",
+            "payment_url",
+        )
+
+    shift = serializers.PrimaryKeyRelatedField(read_only=True)
```
se-bastiaan, line 94: Isn't is possible to make this the ShiftSerializer directly?

website/sales/api/v2/serializers/shift.py
```diff
@@ -0,0 +1,56 @@
...
+    products = ProductListItemSerializer(
+        source="product_list.product_items", many=True, read_only=True
+    )
+
+    title = serializers.SerializerMethodField("_get_title")
+
+    def _get_title(self, instance):
+        return instance.title
+
+    product_sales = serializers.JSONField("product_sales")
```
se-bastiaan, line 56:
```suggestion
    product_sales = serializers.JSONField()
```
should this not just work?

website/sales/models/order.py
```diff
@@ -0,0 +1,213 @@
...
+        verbose_name_plural = "orders"
+        permissions = [
+            ("custom_prices", "Can use custom prices and discounts in orders"),
+        ]
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+
+    created_at = models.DateTimeField(default=timezone.now)
+
+    # created_by = models.ForeignKey(Member, on_delete=SET_NULL, related_name='orders_created',) # TODO add this field
```
se-bastiaan, line 47: Should this still be added?

website/sales/models/shift.py
```diff
@@ -0,0 +1,148 @@
...
+                AND ("payment_id" IS NOT NULL OR ("payment_id" IS NULL AND "total__"=0))
+                )""",
+            [],
+        )
+
+    @queryable_property(annotation_based=True)
+    @classmethod
+    def num_orders(cls):
+        return RawSQL(
+            """(SELECT COUNT(*) FROM "sales_order"
```
se-bastiaan, line 99: This really looks like something you should not need raw SQL for.

website/sales/views.py
```diff
@@ -0,0 +1,41 @@
...
+from django.shortcuts import get_object_or_404, redirect, render
+from django.utils.decorators import method_decorator
+from django.utils.translation import gettext_lazy as _
+from django.views import View
+
+from sales.models.order import Order
+
+
+@method_decorator(login_required, name="dispatch")
+class OrderPaymentView(View):
```
se-bastiaan, line 12: Looks like you really need a error/success page for the redirect. Loads of TODOs here :)

se-bastiaan (changes requested):

website/sales/models/order.py
JobDoesburg: Yes I think this should be desirable.

website/sales/models/shift.py
JobDoesburg: Agreed

website/sales/views.py
JobDoesburg: Yes. Also see #1616

website/sales/api/v2/serializers/order.py
JobDoesburg: I figured that would just be too much data to send that's not really required to know

website/sales/models/shift.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+    locked = models.BooleanField(
+        blank=False,
+        null=False,
+        default=False,
+        help_text=_(
+            "Prevent orders being changed or created for this shift. This will also clean up all unpaid orders in this shift."
+        ),
+    )
```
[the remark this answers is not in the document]
JobDoesburg: Right now, no not from the admin. If we would ever want to do that, it should only be possible with special permissions (only the board) and even then I think we should log it specifically (because of audit stuff). 

So: no

website/sales/models/shift.py
```diff
@@ -0,0 +1,171 @@
...
+    def save(
+        self, force_insert=False, force_update=False, using=None, update_fields=None
+    ):
+        if self.locked:
+            self.orders.filter(
+                (Q(payment__isnull=True) & Q(total_amount__gt=0))
+                | Q(order_items__isnull=True)
+            ).delete()
+
+        return super(Shift, self).save(force_insert, force_update, using, update_fields)
```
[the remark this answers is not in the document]
JobDoesburg: paid*

Orders that are not paid will be deleted on locking the shift (see help text of locked).

website/sales/models/shift.py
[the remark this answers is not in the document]
se-bastiaan: For future reference: logging is possible to write to the django object history.

website/sales/views.py
se-bastiaan: I saw the issue already, but should we have that block this PR?

JobDoesburg: No not necessarily. 

I'm also wondering what you guys think about the exact payment flow. Currently there is a 2-step flow because I'm using the payment templatetag (see the comments in the code)

website/sales/models/shift.py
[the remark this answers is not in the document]
JobDoesburg: @se-bastiaan what do you mean exactly? I don't see anything specific in the payments app with special object history

JobDoesburg:
Todo on this pr:

- [ ] Test coverage
- [ ] Fix the order payment view / flow
- [ ] Shift permissions
- [ ] Order created_by
- [ ] Make orderitem total and discount writable for members with special price permissions
- [ ] PaymentSerializer API v2

website/sales/models/shift.py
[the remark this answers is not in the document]
se-bastiaan: Oh sorry it is registrations: https://github.com/svthalia/concrexit/blob/master/website/registrations/services.py#L176

website/sales/models/order.py
JobDoesburg: I decided to implement this using LogEntry objects

website/sales/views.py
se-bastiaan: I tried this and I would say that it is somewhat cumbersome, seems to me that you could just redirect to the actual payment page instead.

JobDoesburg: The sad thing is that the normal payment confirmation page is a form that requires a `POST`. But a QR code can't do that. So I will need to create a special payment confirmation page for this :(

JobDoesburg:
The coverage right now is about 95% for models and API. I have checked the missing coverage lines and they are not that interesting to test (and actually quite hard to create actual test for).

Yes the tests are quite ugly, not optimally structured and contain a lot of duplicate code. But it are tests so I'm inclined to just not care.

I will write some tests for the admin still

website/sales/views.py
se-bastiaan: You could add some Javascript to the rendered page that automatically submits the form if JS is enabled to skip the step? Maybe that is an option.

I was also thinking about adding this page to thadmin, but that would make it harder because then you would need to login to Thadmin also instead of just the website. Regular users should probably not be bothered with another website.

JobDoesburg: I also thought about the javascript option yeah, I think I will be doing that.

I think `thadmin` shouldn't be for regular users. Thadmin is just a nice frontend to optimize the UI/UX for shift managers. The whole system should operate without `thadmin` as well (which it does right now, you can just use the Django admin for everything and it will work fine)

se-bastiaan: I think we have reached a good consensus 👍

JobDoesburg:
We're currently at about 95% coverage and all actually interesting critical stuff is tested (like everything with specific permissions). I think it's fine to keep it like this.

I could still add some documentation and fixtures if we really want it. But we can also just merge it :)

JobDoesburg:
Can anyone help me figure out why pylint is failing on code that I didn't touch?

website/thaliawebsite/settings.py
```diff
@@ -571,7 +575,7 @@ def from_env(
         "thaliawebsite.api.throttling.AnonRateThrottle",
         "thaliawebsite.api.throttling.UserRateThrottle",
     ],
-    "DEFAULT_THROTTLE_RATES": {"anon": "100/day", "user": "1000/day"},
+    "DEFAULT_THROTTLE_RATES": {"anon": "100/day", "user": "10000/day"},
```
se-bastiaan, line 578: Maybe it would be better to make it like 20/min? This means that it is basically unlimited if you want to do 10000 calls in a minute.

```suggestion
    "DEFAULT_THROTTLE_RATES": {"anon": "100/day", "user": "20/min"},
```

website/thaliawebsite/templates/index.html
```diff
@@ -14,6 +14,11 @@
 {% block body %}
     <section id="thaliawebsite-frontpage" class="page-section">
         <div class="container">
+            {% if messages %}
```
se-bastiaan, line 17: This should be in a separate PR if we want to do it properly since it closes #1616

website/sales/views.py
```diff
@@ -0,0 +1,38 @@
...
+    def get(self, request, *args, **kwargs):
+        order = get_object_or_404(Order, pk=kwargs["pk"])
+        if order.payment:
+            messages.warning(request, _("This order was already paid for."))
+            return redirect("index")
```
se-bastiaan, line 18: I think it would be neat to render a different page instead of redirecting to the index. That way the error will be the most important thing on the page.

website/sales/templates/sales/order_payment.html
```diff
@@ -0,0 +1,22 @@
...
+    <section class="page-section">
+        <div class="container">
+            <h1 class="section-title text-center">{% trans "Pay for order" %}</h1>
+            <div class="text-center">
+                {{ order.order_description }}
+            </div>
+            <div class="text-center mt-4">
+                {% url 'index' as next %}
+                {% payment_button order next %}
+                <script type="text/javascript">
```
se-bastiaan, line 14: this script should be in a `{% block js_body %}` block (don't forget the super)

se-bastiaan: maybe it would also be smart to give the payments button an id or name so that it is possible to select it by name instead of by index, which may be affected if we somehow add a form to the base template.

website/events/models/event.py
```diff
@@ -181,6 +181,8 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
 
     tpay_allowed = models.BooleanField(_("Allow Thalia Pay"), default=True)
 
+    shift = models.OneToOneField("sales.Shift", models.SET_NULL, null=True)
```
se-bastiaan, line 184: I do not completely understand why the shift is part of the event and not the other way around, I don't see the shift used anywhere in the events package. Which means that this is a dependency from events on sales that does not seem necessary to me. Wouldn't it then be better to make the field a part of the shift instead?

website/payments/templates/payments/widgets/payment.html
```diff
@@ -5,7 +5,7 @@
     {% elif not payment and obj.payment_amount != 0 %}
         {% trans "Unpaid" %}
         -
-        € {{ obj.payment_amount }}
+        € {{ obj.payment_amount|floatformat:2}}
```
se-bastiaan, line 8:
```suggestion
        € {{ obj.payment_amount|floatformat:2 }}
```

website/sales/admin.py
```diff
@@ -0,0 +1,560 @@
...
+    inlines = [
+        OrderInline,
+    ]
+    search_fields = (
+        "id",
+        "custom_title",
+        "start",
+        "end",
+    )
+    date_hierarchy = "start"
```
se-bastiaan, line 451:
```suggestion
    filter_horizontal = ("managers",)
    date_hierarchy = "start"
```

website/sales/models/shift.py
```diff
@@ -0,0 +1,189 @@
...
+)
+
+from activemembers.models import MemberGroup
+from sales.models.product import ProductList
+
+
+class Shift(models.Model):
+    class Meta:
+        permissions = [
+            ("bypass_manager", _("Can bypass shift manager restrictions")),
```
se-bastiaan, line 24:
```suggestion
            ("override_manager", _("Can access all shifts as manager")),
```
I'd like to see this called an override instead of a bypass, since that is the name the permission also has in the events app.

website/sales/models/shift.py
```diff
@@ -0,0 +1,189 @@
...
+        return ""
+
+    def __str__(self):
+        if self.title != "":
+            return f"Shift {self.pk} - {self.title}"
+        return f"Shift {self.pk}"
+
+    @staticmethod
+    def currently_active():
+        return Shift.objects.filter(active=True)
```
se-bastiaan, line 177: I think a better way to do this would be creating a custom manager and making it `Shift.active_objects`

website/sales/models/shift.py
```diff
@@ -0,0 +1,189 @@
...
+    def is_manager(self, member):
+        if member and member.is_authenticated:
+            if member.is_superuser or member.has_perm("sales.bypass_manager"):
+                return True
+            return (
+                member.get_member_groups()
+                .filter(pk__in=self.managers.values_list("pk"))
+                .count()
+                != 0
+            )
```
se-bastiaan, line 188:
```suggestion
            return member.is_superuser or member.has_perm("sales.bypass_manager") or (
                member.get_member_groups()
                .filter(pk__in=self.managers.values_list("pk"))
                .count()
                != 0
            )
```

website/sales/models/order.py
```diff
@@ -0,0 +1,260 @@
...
+    class Meta:
+        verbose_name = _("order")
+        verbose_name_plural = _("orders")
+        permissions = [
+            ("custom_prices", _("Can use custom prices and discounts in orders")),
+        ]
+        ordering = ["created_at"]
+        indexes = [
+            models.Index(fields=["shift"]),
+            models.Index(fields=["id"]),
```
se-bastiaan, line 47: I believe both of these indices are obsolete.

website/sales/api/v2/views.py
```diff
@@ -0,0 +1,245 @@
...
+        "PUT": ["sales:write"],
+        "DELETE": ["sales:write"],
+    }
+
+    def retrieve(self, request, *args, **kwargs):
+        if not Order.objects.get(pk=kwargs["pk"]).shift.is_manager(
+            Member.objects.get(pk=self.request.user.pk)
+        ):
+            response = {"message": "You are not allowed to view this order."}
+            return Response(response, status=status.HTTP_403_FORBIDDEN)
```
se-bastiaan, line 182:
```suggestion
            raise PermissionDenied
```

website/sales/api/v2/views.py
```diff
@@ -0,0 +1,245 @@
...
+    method_serializer_classes = {
+        ("GET",): OrderListSerializer,
+        ("POST",): OrderSerializer,
+    }
+
+    def get_serializer_class(self):
+        for methods, serializer_cls in self.method_serializer_classes.items():
+            if self.request.method in methods:
+                return serializer_cls
+        raise exceptions.MethodNotAllowed(self.request.method)
```
se-bastiaan, line 91: This piece of code could be positioned a bit better

website/sales/api/v2/views.py
```diff
@@ -0,0 +1,245 @@
...
+    queryset = Shift.objects.all()
+    permission_classes = [
+        IsAuthenticatedOrTokenHasScope,
+        DjangoModelPermissions,
+    ]
+    required_scopes = ["sales:read"]
+
+    def retrieve(self, request, *args, **kwargs):
+        if not Shift.objects.get(pk=kwargs["pk"]).is_manager(
+            Member.objects.get(pk=self.request.user.pk)
```
se-bastiaan, line 74:
```suggestion
            self.request.member
```

website/sales/api/v2/serializers/order.py
```diff
@@ -0,0 +1,192 @@
...
+        shift = self.context["shift"]
+        order = Order.objects.create(shift=shift)
+        if "order_items" in validated_data:
+            items_data = validated_data.pop("order_items")
+            for item_data in items_data:
+                OrderItem.objects.create(order=order, **item_data)
+        return order
+
+    def update(self, instance, validated_data):
+
```
se-bastiaan, line 147:
```suggestion
```

website/sales/api/v2/serializers/order.py
```diff
@@ -0,0 +1,192 @@
...
+        max_digits=6, decimal_places=2, min_value=0, read_only=True
+    )
+
+    discount = serializers.DecimalField(
+        max_digits=6, decimal_places=2, min_value=0, read_only=True
+    )
+
+    total_amount = serializers.DecimalField(
+        max_digits=6, decimal_places=2, min_value=0, read_only=True
+    )
```
se-bastiaan, line 119: I do not entirely understand why these fields are not just part of the `read_only_fields` list. Please enlighten me :)

se-bastiaan: I also see that some still have the read only attribute and are in that list as well?

website/sales/admin.py
```diff
@@ -0,0 +1,560 @@
+from functools import partial
```
se-bastiaan, line 1: Maybe it would be nice to split these admin configs into multiple files for readability? :p

se-bastiaan (changes requested):
I have posted a few comments :p

poetry.lock
```diff
@@ -894,30 +970,33 @@ snowballstemmer = "*"
...
+version = "2.8.1"
 description = "Pygments is a syntax highlighting package written in Python."
 category = "main"
 optional = true
 python-versions = ">=3.5"
 
 [[package]]
 name = "pylint"
-version = "2.6.0"
+version = "2.7.4"
```
se-bastiaan, line 981: You updated pylint in the lock file, so this is why you get new errors :p

website/events/models/event.py
JobDoesburg: For me it feels fundamentally weird to have a dependency from sales on events... Shifts should be standalone things if you ask me. It's only that certain events will have a shift.

website/sales/api/v2/serializers/order.py
JobDoesburg: I see, I'm not consistent in this. I do need to specify it are decimal fields explicitly, because Django is not able to correctly derive the field type for these queryable properties on all database systems. The read only thing can be done more consistently. Will fix

website/events/models/event.py
se-bastiaan: Yes that is also true, but you create the optional dependency on events already by checking for the event attribute. So now you have a dependency in two places I think?

I'm not sure, maybe it would just be better to not set the title using the event? That is the only thing it does right?
I am also afraid that people will not set the shift on the event just because they are lazy, since shifts will probably only be created once they will need one.

website/sales/models/shift.py
JobDoesburg: I rather don't because then I would need to override the QueryablePropertiesManager for that. But this code isn't used anymore anyway, so I'll just remove it

website/thaliawebsite/templates/index.html
JobDoesburg: Do we really want to do it properly, or shall I just close 1616 with this 😅

website/sales/views.py
JobDoesburg: I think it isn't necessary to do this, as it is a hypothetical situation that should never happen in practice if people don't try to mess with the system

website/sales/models/order.py
JobDoesburg: I will just remove them, I trust the documentation that they are not required and consider the speedup I thought I got is due to something different

website/thaliawebsite/templates/index.html
se-bastiaan: Well that is what I was asking you.

website/sales/views.py
se-bastiaan: That is a good point.

website/events/models/event.py
JobDoesburg: It's not a _real_ dependency as I use the `has_attr` thingy. But you are right, maybe I should just delete it. Or at least, delete the title property. I think we could keep the field in the events. In the end, I think it will become something similar to pizza events.

JobDoesburg:
Right now the autocompletefilter that I'm using in the admin is broken because of Django 3.2, but a fix is on its way I think: https://github.com/farhan0581/django-admin-autocomplete-filter/issues/53

website/events/models/event.py
se-bastiaan: If you're not using it then it should probably not exist.

To be fair though, pizzas and events are pretty tightly coupled (which is bad). They also have a circular dependency. But at least the key to the Event is in the PizzaEvent :p

website/sales/admin.py
JobDoesburg: Wow sick, didn't know this existed. I thought I had to define a whole admin form with widgets for this

website/events/models/event.py
JobDoesburg: Maybe in (far) the future, `sales` could replace `pizzas` 🤔...

I will remove the title thingy from the sales model, but I do want to keep the the OneToOneField actually so we can just start keeping the data about Shifts related to Events from the beginning, just like we're doing with Photo albums as of now.

JobDoesburg:
> Code Climate has analyzed commit [194089c](https://github.com/svthalia/concrexit/commit/194089c24f4bcff88cf1582e49af8661e7bb7831) and detected **59 issues** on this pull request.
> 
> Here's the issue category breakdown:
> 
> Category	Count
> Complexity	18
> Duplication	41
> View more on [Code Climate](https://codeclimate.com/github/svthalia/concrexit/pull/1445).

This is largely because of the test data I set up in the tests. This could actually be moved to a fixtures file if we really want that (but it will still be sóms duplication

website/events/models/event.py
se-bastiaan: You make the assumption that people are going to fill in a field that is not required. :)

I am not sure if you were already in the Technicie when we introduced categories for events, but the default value was always set to 'Other'. And thus people did not change the value of the field to the correct one. I'm fine with keeping the field but I expect the same to happen here.

JobDoesburg: We can see, a man can dream. I do have faith, as the 'category' field doesn't enable a real feature or something, while this will (I could even think about writing the code in such a way that committee members can only create shifts that are linked to an event - but not in this PR).

se-bastiaan: But you should not want shifts that are only connected to events. Maybe you want shifts without events too.

And for that the field has to be in the shift since it has to exist before you can connect it to and event.

JobDoesburg: I don't mean implementing it at the model level. I mean disabling the add permission on the ShiftAdmin for members that don't have the shift manager override permission, and only enabling the add permission via the EventAdmin for them.

se-bastiaan:
@pingiun build in Nix fails here

se-bastiaan (approved):

PR #1451, opened 2020-12-09
Fix bug when deleting users crashed

### LINKED ISSUE #1328: Deleting a user results in a full crash
### Describe the bug
Have a user with profile and payments and such. If we try to delete the user, Django will crash completely:

```
Fatal Python error: Cannot recover from stack overflow.
Python runtime state: initialized
[...]
Process finished with exit code 250
```

### How to reproduce
1. have a user with a profile and payments
2. try to delete the user (via the auth user admin)

### Expected behaviour
Not crash (but disallow deleting? as we don't want it, we can only minimize their data) 

### Additional context
I think it is related to the payment, because in the stack trace there are references to the Payment __init__ code. But I can't see the problem directly

pingiun:
Closes #1328

### Summary
When looking up foreign keys in the __init__ method things go wrong with
some kind of infinite recursion. This removes the database lookup within
__init__ and moves it to the clean() method where it's actually used.

### How to test
1. Go to user which has done Payments
2. Delete user
3. Website doesn't crash

### DIFF
--- a/website/payments/models.py
+++ b/website/payments/models.py
@@ -8,7 +8,7 @@ from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Q, Sum
+from django.db.models import DEFERRED, Q, Sum
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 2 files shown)


pingiun:
Tbf I don't know if this is a good solution, but it does fix the problem and still passes the tests. In the future we may want to move checks across model changes to a different place, or use signals, or save the data in some other way

JobDoesburg:
I'm also unsure whether this is the best way to fix these kind of things

JobDoesburg:
We could maybe also use [django-simple-history](https://django-simple-history.readthedocs.io/en/latest/) to keep old versions of instances

se-bastiaan:
But we don't want to keep old versions, we just want to compare the existing version with the new value. If it works it is fine by me but I really don't understand this code tbh.

pingiun:
> [...] but I really don't understand this code tbh.

I hoped the comment explaining would be enough. Basically what this change does is delay the database lookup to the `clean` method, so the `__init__` method does not touch the database. 

I basically looked at the base `__init__` code to see how you can get the required things from the args and kwargs.

Does this make it clearer and should I add some more to the comment?

se-bastiaan (approved):

PR #1456, opened 2020-12-10
Fix sorting by payments in events admin

### LINKED ISSUE #1419: Sorting by payment in events admin is broken
### Describe the bug
Sorting by payment in events admin is broken

### How to reproduce
1. Go to an event with payments
2. Try to sort on payments
3. It is not completely correct

### Expected behaviour
It should sort by paid/not paid and payment type and when clicking again it should revert the order.

pingiun:
Closes #1419

### Summary
Rewrote the payment sorting parser and changed the sort behaviour when
changing payments. Now the table will not reorder when changing a
payment, as this happens after a network call and can cause an
unexpected moving of the edited row. Now instead the table will not be
resorted, and instead is reverted to an unordered state.

### How to test
1. Go to event with payments
2. Sort based on payments
3. Check if sorting changes the order

### DIFF
--- a/website/events/static/events/js/admin.js
+++ b/website/events/static/events/js/admin.js
@@ -24,6 +24,7 @@ django.jQuery(function () {
         if (payment_previous === $(this).val()) {
             return;
         }
+        $('table').trigger('update', false).trigger('sortReset');
 
         if ($(this).val() === none) {
             $(this).removeClass('paid');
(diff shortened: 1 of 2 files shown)


se-bastiaan:
deploy review

se-bastiaan:
remove review

se-bastiaan (changes requested):
When you change the value of the payment the sorting won't work properly until the page is refreshed.

pingiun:
This makes sense, I guess I'll have to fix the payment parser then

se-bastiaan:
Well it is weird because `$('table').trigger('update');` should take care of the sorter update.

se-bastiaan:
deploy review

pingiun:
This seems to work great on the reviewhost

se-bastiaan (approved):

se-bastiaan:
remove review

PR #1459, opened 2020-12-13
Refactor settings

pingiun:
### Summary
Put all the settings in one file for clarity.

This should be a better version of  #1377

I wanted to make it clearer how settings are loaded, and before you had
to check in multiple files, depending on some vague environment
variables.

Now every setting is set in settings.py, and you can more easily check
which setting applies based on the DJANGO_ENV you have set. I copied
this concept from nodejs. Two helper functions are added which are
useful when you have different values or different defaults in these
environments.

### How to test
Nothing should be changed, so if the tests and running the website work it should be fine


pingiun:
I have no clue why building the docs goes wrong, it totally works locally

se-bastiaan:
It seems to be expecting a host for postges when building the docs.

Anyways I find this really hard to correctly review but the idea seems good to me.

website/thaliawebsite/settings.py
```diff
@@ -0,0 +1,604 @@
...
+
+# Many of the settings are dependent on the environment we're running in.
+# The default environment is local, because the environment is set to production in wsgi.py
+DJANGO_ENV = os.environ.get("DJANGO_ENV", "local")
+_environments = ["production", "staging", "testing", "local"]
+if DJANGO_ENV not in _environments:
+    raise Misconfiguration(f"Set DJANGO_ENV to one of: {', '.join(_environments)}")
+
+
+def setting(*, local, production, staging=_NOT_SET, testing=_NOT_SET):
```
lukovdm, line 57: It would be nice to make it a bit clearer what happens when staging or testing is not set

website/thaliawebsite/settings.py
```diff
@@ -0,0 +1,604 @@
...
+    if DJANGO_ENV == "testing":
+        return testing
+    if DJANGO_ENV == "production" or staging is _NOT_SET:
+        return production
+    if DJANGO_ENV == "staging":
+        return staging
+    raise Misconfiguration(f"Set DJANGO_ENV to one of: {', '.join(_environments)}")
+
+
+def from_env(
```
lukovdm, line 69: It would be nice to make it a bit clearer what happens when staging or testing is not set

lukovdm (changes requested):

pingiun:
Changed to draft because I want to change the nix configuration to go along with this

README.md
```diff
@@ -12,6 +12,8 @@ The latest Thalia Website built on Django.
 Getting started
 ---------------
 
+If you use Docker, please look at [this part](#docker) of the README.
```
pingiun, line 15: This needs to be removed

website/thaliawebsite/templates/base.html
```diff
@@ -84,24 +77,14 @@
...
-                    <source srcset="{% static 'img/logo.svg' %}" media="(prefers-color-scheme: light)">
-                    <img
-                        src="{% static 'img/logo.svg' %}"
-                        id="logo"
-                        alt="{% trans 'Study Association Thalia' %}"
-                        height="220px"
-                    >
-                {% endif %}
+                <source srcset="{% static 'img/logo-dark.svg' %}" media="(prefers-color-scheme: dark)">
+                <source srcset="{% static 'img/logo-.svg' %}" media="(prefers-color-scheme: light)">
```
pingiun, line 81: Remove dash

Legosteen11 (dismissed):

se-bastiaan (approved):

PR #1460, opened 2020-12-14
Move email send function to utils

### LINKED ISSUE #1305: Use utils.snippets.send_email in everywhere
### Describe the change
Use `utils.snippets.send_email` in all apps. For example, in `registrations` we use our 'own' `_send_email` and `payments` uses the one from `registrations`. They should all use `utils` in fact to centralize this code.

### Motivation
Less code is more better

### Additional context
Maybe we should also check `events` (or other apps)

Tedddoo:
Closes #1305 Use utils.snippets.send_email everywhere

### Summary
I have moved the send_emails function to utils since it is a better position.

### How to test
Try to send a mail when register a new account or send a mail for a payment

### DIFF
--- a/website/registrations/tests/test_emails.py
+++ b/website/registrations/tests/test_emails.py
@@ -20,7 +20,7 @@ class EmailsTest(TestCase):
     def setUp(self):
         translation.activate("en")
 
-    @mock.patch("utils.snippets.send_email")
+    @mock.patch("registrations.emails.send_email")
     def test_send_registration_email_confirmation(self, send_email):
         reg = Registration(
             email="test@example.org",
@@ -45,7 +45,7 @@ class EmailsTest(TestCase):
                 },
             )
 
-    @mock.patch("utils.snippets.send_email")
+    @mock.patch("registrations.emails.send_email")
     def test_send_registration_accepted_message(self, send_email):
         reg = Registration(
             email="test@example.org",
@@ -65,7 +65,7 @@ class EmailsTest(TestCase):
                 {"name": reg.get_full_name(), "fees": floatformat(reg.contribution, 2)},
             )
 
-    @mock.patch("utils.snippets.send_email")
+    @mock.patch("registrations.emails.send_email")
     def test_send_registration_rejected_message(self, send_email):
         reg = Registration(
             email="test@example.org", first_name="John", last_name="Doe", pk=0,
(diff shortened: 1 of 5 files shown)


website/utils/snippets.py
```diff
@@ -178,3 +180,20 @@ def overlaps(check, others, can_equal=True):
             return True
 
     return False
+
+
+def _send_email(to: str, subject: str, body_template: str, context: dict) -> None:
```
pingiun, line 185: Because this is now a function that is "exported", so used in multiple places. I think it's better to change it to `send_email` (withouth the leading underscore)

Tedddoo: I tried so hard, and got so far but in the end it doesn't even matter.
Ik heb het aangepast en nu doet hij het niet meer. Ik heb meerdere van die make tests gedaan en niks maakt het probleem duidelijk.

se-bastiaan: I think your edit was successful tho, the tests are not failing and I see no weird this in the code that suggest it should not work.

website/registrations/tests/test_emails.py
```diff
@@ -12,15 +12,15 @@
 
 from members.models import Member, Profile
 from registrations import emails
-from registrations.emails import _send_email
+from registrations.emails import send_email
```
se-bastiaan, line 15: This test is most likely failing here. This import is incorrect.

se-bastiaan (changes requested):
Some imports are non existent.

website/utils/snippets.py
se-bastiaan: I took the liberty to rebase the code. To get the latest version locally do:
```
git fetch
git checkout utils_email_send
git reset --hard origin/utils_email_send
git pull
```

website/registrations/tests/test_emails.py
se-bastiaan: Okay it is not failing but it is still incorrect I believe?

website/registrations/tests/test_emails.py
```diff
@@ -12,15 +12,15 @@
...
+from utils.snippets import send_email
 from registrations.models import Registration, Renewal
 
 
 class EmailsTest(TestCase):
     def setUp(self):
         translation.activate("en")
 
-    @mock.patch("registrations.emails._send_email")
+    @mock.patch("registrations.emails.send_email")
```
se-bastiaan, line 23: All these mock still reference the non-existing code.

se-bastiaan (changes requested):
.

se-bastiaan (approved):

PR #1468, opened 2020-12-16
Enforce consistent docstyle

### LINKED ISSUE #1294: Use pydocstyle to check documentation
### Describe the change
Add [`pydocstyle`](https://github.com/PyCQA/pydocstyle/) as a linter.

### Motivation
`pydocstyle` is a linter that checks the documentation of code according to [PEP 257](https://www.python.org/dev/peps/pep-0257/). This will improve the overall documentation, because developers are forced to write basic documentation (a tedious, but important part of development).

### Suggested implementation
Add `pydocstyle --match-dir='(?!migrations).*' website` to the linting GitHub Actions workflow.

### Additional context
We need two changes for this to happen.
First all of the existing code needs to be documented using PEP 257:
- [ ] activemembers
- [ ] announcements
- [ ] documents
- [ ] education
- [ ] events
- [ ] mailinglists
- [ ] members
- [ ] merchandise
- [ ] newsletters
- [ ] partners
- [ ] payments
- [ ] photos
- [ ] pizzas
- [ ] pushnotifications
- [ ] registrations
- [ ] singlepages
- [ ] thabloid
- [ ] thaliawebsite
- [ ] utils

And secondly, we need to add it to our CI.
- [ ] Add `pydocstyle` to GitHub Actions workflow

pingiun:
Closes #1294

### Summary
Enforce consistent docstyle

### How to test
No functionality has changed

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -6,17 +6,6 @@ on:
   - pull_request
 
 jobs:
-  codestyle:
-    name: Codestyle
-    runs-on: ubuntu-18.04
-    container: thalia/concrexit-dependencies
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-
-      - name: Run black
-        run: make blackcheck
-
   linting:
     name: Linting
     runs-on: ubuntu-18.04
@@ -27,6 +16,12 @@ jobs:
 
       - name: Run pylint
         run: make pylint
+
+      - name: Run black
+        run: make blackcheck
+
+      - name: Run pydocstyle
+        run: make pydocstyle
 
   tests:
     name: Django Tests
@@ -124,7 +119,7 @@ jobs:
   build-docker:
     name: Build Docker Image
     runs-on: ubuntu-18.04
-    needs: [codestyle, tests, documentation-tests]
+    needs: [linting, tests, documentation-tests]
     env:
       DOCKER_TAG: thalia/concrexit:${{ github.sha }}
     steps:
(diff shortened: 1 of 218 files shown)


.github/workflows/test.yaml
```diff
@@ -6,17 +6,6 @@ on:
...
-    container: thalia/concrexit-dependencies
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-
-      - name: Run black
-        run: make blackcheck
-
   linting:
     name: Linting
```
se-bastiaan, line 10: If this contains style checks then the name should reflect that.

In any case this removes the parrallellism of the codestyle and the linting jobs. And it will show as 1 check in the overview in PRs, I really don't like that personally. I like to see exactly what failed by just glancing over the checks instead of having to open them.

se-bastiaan (changes requested):
Tests are not completing?

pingiun:
> Tests are not completing?

I consolidated codestyle into linting. Not sure why it still shows up here.

.github/workflows/test.yaml
pingiun: For parallelism I could put all three in a seperate check?

pingiun: That is what I have done now

pingiun:
By the way, this explicitely excludes the pydocstyle rules about there not being docstrings. This might incentivise developers to not write docstrings because the CI will yell at them if they do it wrong. Maybe we want something so all new code should contain docstrings? Idk, something to think about

se-bastiaan (approved):

PR #1477, opened 2020-12-21
Add NixOS test

### LINKED ISSUE #1464: Add testing to see if pages load correctly
As #1461 showed, it seems that it is possible for fairly critical page load errors (in this case, `GET /events/`) to slip through testing.

I would like more tests added, to increase robustness and reduce the chance of an edit of templates etc to break pages. In this case, it would be adding tests to see if pages load correctly.

### Motivation
#1461 showed that this can be a real issue. Making sure that testing is comprehensive can prevent issues like this from getting merged in the future.

pingiun:
Closes #1464 

### Summary
NixOS has a testing module where you can write python scripts to test an actual system (running as a VM). I added a test that requests pages from the sitemap and checks if they return a succesful page.

### How to test
We cannot test this yet as GitHub Actions does not have the ability to start up a VM

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -12,7 +12,7 @@ jobs:
       with:
         name: svthalia-concrexit
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
-    - run: nix-build -j 4 release.nix
+    - run: nix-build -j 4 -A github-actions release.nix
     - run: nix-shell --run "echo OK"
     - name: Deploy if this is master or staging
       if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/staging'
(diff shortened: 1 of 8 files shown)


pingiun:
I based this branch on refactor-settings2 (#1459) because then I don't have to change nix things after that is merged

se-bastiaan (commented):
I don't entirely get why this has to be a Nix test, wouldn't it be perfectly fine to be a general test module. The code you wrote is Python already and having tests to test Django in Nix doesn't really make it easier to understand what is happening.

pingiun:
This PR is succeeding on hydra: https://hydra.technicie.nl/jobset/concrexit/1477 (see also the github status checks)

se-bastiaan:
But it shows a 502 error and the other tests fail :p

se-bastiaan (approved):

PR #1481, opened 2020-12-23
Remove push trigger from test workflow

se-bastiaan:
### Summary
This removes push as trigger for PR workflows. I believe we're doing double pipelines for every PR and that is not necesary.

### How to test
1. Cannot really test this right.


pingiun (approved):
Let's try

PR #1487, opened 2021-01-01
Add foreignkey between photos and event

### LINKED ISSUE #1363: Add an optional from albums ForeignKey to events
Not really. But conceptually, it feels weird that our photo albums (which are almost always photos of an event) are not connected to those events. For me it feels like in some way it should be possible to have that connection.

Just an optional Foreign Key for Album to Event (I thought about a one to one field even I think that wouldn't work in some cases maybe)

### Motivation
Maybe in the future (our 7th lustrum maybe) someone would like to write a tool that does something fancy with photos taken at events. Or to send attendees a notification if photos are updated.

And it is really easy to add.

### Additional context
To populate our historical albums, we probably would want to write some script that matches events and albums semi-automated based on the date of both and possibly the name (it is no problem if this does require a little bit of guidance by the user for some cases, or if it is not perfect in all cases)

Jorritboer:
Closes #1363 

### Summary
Added an optional foreign key between photo album and event. In a photo album you can click on event to go to the linked event, and on an event page you can click on the linked albums to go to one of the corresponding photo albums.
Secondly there is also a script to link the old events and albums. This is done by making a csv file which you can then edit and is read to link the albums and events you put in the same row.

### How to test
Test the feature:
1. Go to photo album admin page
2. Add an event

3. Go to the album page 
4. Click on the 'event' button
5. On the event page, click on the 'album' button

Test the script:

1. Make some events and albums on the same day
2. Run the linking script link_album_events
3. Edit the csv file
4. Check if the events and albums are properly linked


JobDoesburg:
This also needs a (management? or migration) script that tries to match existing albums with existing events. For example: match albums with an event if the date is the same, and there is no other event on that day or the event name matches for more than x percent

Jorritboer:
I wanted to ask you what your idea was on how to do that. 
I thought it would be something like a script that we run once so doesn't need to be in the repository.

JobDoesburg:
I think it can be in 

> I wanted to ask you what your idea was on how to do that.
> I thought it would be something like a script that we run once so doesn't need to be in the repository.

I think it shouldn't be a problem to have it in the repository. I personally would create a management command `linkeventstoalbums` or something. Another approach would be to add the script as a manual migration so it will automatically run on migration.

Jorritboer:
I'll add it to this pull request then. 
A migrations sounds like a good idea, I'll take a look what works best. What do you mean with a management command? Something in the makefile?

Jorritboer:
Ah never mind I found it, it's a command like 'download addresses from selected users'.

JobDoesburg:
> Ah never mind I found it, it's a command like 'download addresses from selected users'.

No that is an admin action. A management command is a separate file, like `init_gsuite_users` in `activemembers/management/commands/init_gsuite_users.py`

website/photos/models.py
```diff
@@ -83,6 +84,10 @@ class Album(models.Model, metaclass=ModelTranslateMeta):
         ScheduledMessage, on_delete=models.deletion.SET_NULL, blank=True, null=True
     )
 
+    event = models.OneToOneField(
```
se-bastiaan, line 87: I think this should be a foreignkey instead of a 1-to-1.

se-bastiaan (changes requested):
@Jorritboer I think this needs a rebase and if it Is done it can be marked as "ready for review"

website/events/views.py
```diff
@@ -66,6 +66,8 @@ def get_context_data(self, **kwargs):
 
         context["slide_size"] = settings.THUMBNAIL_SIZES["slide"]
 
+        context["albums"] = event.album_set.all()
```
se-bastiaan, line 69: This should not be necessary, you can just access `event.album_set` in the html file itself :)

Jorritboer:
I have added the linking script. It is kinda whacky by using a csv file (also check the code climate cognitive complexity of 24 vs 5 recommended ;p) but I don't think there is a really different way of doing it that would be better. After all it only really needs to be run once. Nonetheless there are probably some things to be done to optimize the script a bit and deal with some possible issues.

website/photos/management/commands/link_album_events.py
```diff
@@ -0,0 +1,111 @@
...
+            file.write(
+                str(albums[j].pk) + "-" + albums[j].title + "-" + str(albums[j].date)
+            )
+        file.write("\n")
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        events = Event.objects.order_by("start").all()
+        albums = Album.objects.order_by("date").all()
```
JobDoesburg, line 30: Perhaps you should filter on albums that do not have an event linked, so you can rerun the script if required.

website/photos/management/commands/link_album_events.py
```diff
@@ -0,0 +1,111 @@
...
+            elif partitioned_albums[j][0].date == partitioned_events[i][0].start.date():
+                # events and albums on the same day, write both
+                write(partitioned_events[i], partitioned_albums[j], f)
+                i += 1
+                j += 1
+        f.close()
+
+        print("done")
+        print(
+            "Please edit the file such that albums and events that should be linked are next to eachother"
```
JobDoesburg, line 82: Maybe split it into two separate read and write commands

website/photos/management/commands/link_album_events.py
```diff
@@ -0,0 +1,111 @@
...
+            if a.date == last_date:
+                partitioned_albums[len(partitioned_albums) - 1].append(a)
+            else:
+                partitioned_albums.append([a])
+                last_date = a.date
+
+        print("Writing to link_album_events.csv")
+        i = 0  # index of partitioned_events
+        j = 0  # index of partitioned_albums
+        f = open("link_album_events.csv", "w")
```
JobDoesburg, line 54: I think it is better to just print the output to stdout, and read from stdin. Then the user can do the IO via the shell application just as they like, which is more versatile.

Jorritboer: For output that's not really nice right, because you can't print updates or instructions. And then probably nice to do the same for input. 
Maybe just adding an option to supply the file location would be better?

link_album_events.csv
```diff
@@ -0,0 +1,28 @@
+Event (id-title-startdate), Album (id-title-date)
```
JobDoesburg, line 1: This file should be removed

Jorritboer: oops

website/photos/management/commands/link_album_events.py
JobDoesburg: @pingiun what do you think? You know best how to do this on production

se-bastiaan: For our current production environment this would probably work fine, but a file does make it harder to execute. You have to download a file, then change it, then upload the file again. And your server is remote and you usually only have a remote shell unless you start working with scp.

pingiun: I think the best way is to print the csv data output on stdout and print informational messages on stderr. Then the importing step can take input from stdin.

Jorritboer: I didn't see this last comment, I put the information in help

se-bastiaan (approved):

website/events/templates/events/event.html
```diff
@@ -44,6 +44,24 @@ <h1 class="text-center section-title">{{ event.title }}</h1>
                 <div class="col-12 col-md-6 order-1">
                     <table class="table table-borderless">
                         <tbody>
+                        {% if event.album_set.all %}
```
se-bastiaan, line 47: Could this button be moved down to the end of the event information?

website/photos/templates/photos/album.html
```diff
@@ -11,6 +11,12 @@
         <div class="container">
             <h1 class="text-center section-title">
                 {{ album.title }}
+                {% if album.event %}
+                <a class="btn btn-primary btn-second"
+                   href={{ album.event.get_absolute_url }}>
+                   Event
+                </a>
```
se-bastiaan, line 18:
```suggestion
                <a href="{{ album.event.get_absolute_url }}"
                   class="btn btn-primary btn-first"
                   data-toggle="tooltip"
                   data-placement="top"
                   title=""
                   data-original-title="{% trans "Event" %}">
                    <i class="fas fa-calendar"></i>
                </a>
```

se-bastiaan: I'd prefer this to be an icon to keep more space available for the title itself. Moreover, the href was not quoted. And there was some margin to the right that shouldn't be there I think.

Screenshot 2021-04-01 at 20 03 07

se-bastiaan (changes requested):

website/events/templates/events/event.html
Jorritboer: I actually like it here because this will only appear after the event has ended so the rest of the event information is less important then.

se-bastiaan: It's still the information of the event, the link to the photo album is secondary in my opinion since it does not describe the event in any way.

Luca-Castelnuovo:
deploy review

website/events/templates/events/event.html
Luca-Castelnuovo: I also think it's better to place the button below the event information.

Jorritboer: Okay... clear enough I'm in the minority. I have put it all the way down now.

se-bastiaan (approved):

PR #1492, opened 2021-01-04
Added the 'too late to cancel without paying fine'-message to the API…

### LINKED ISSUE #1383: Event (registration) status message in the API
Currently, the event status messages (like 'you cannot cancel your registration without having to pay a fine') are hardcoded and whenever we update them, we must also update the app

Put the message in the API

### Additional context
Also checkout #1381

LCKnol:
…, so it doesn't have to be hardcoded in the app anymore.

Closes #1383 

### Summary
The message that's shown when cancellation isn't possible without paying a fine can now be requested from the API.

### How to test
1. Make sure you're registered for an event for which the registration deadline has already passed 
2. Go to thalia.nu/api/v1/events/1/
2. The message is shown at the bottom: "cancellation isn't possible anymore without..."
3. Go to thalia.nu/events/1/
4. The same message is also shown here

### DIFF
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -183,7 +183,7 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
 
     @property
     def cancel_too_late_message(self):
-        return (
+        return _(
             "Cancellation isn't possible anymore without having to pay "
             "the full costs of €" + str(self.fine) + ". Also note that "
             "you will be unable to re-register. Note: If you have any "
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -183,12 +183,14 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
 
     @property
     def cancel_too_late_message(self):
-        return "Cancellation isn\'t possible anymore without having to pay " \
-                "the full costs of €" + str(self.fine) + ". Also note that " \
-                "you will be unable to re-register. Note: If you have any " \
-                "COVID-19 symptoms you will not have to pay these fees. " \
-                "Let us know via info@thalia.nu that this is this reason " \
-                "for your cancellation."
+        return (
+            "Cancellation isn't possible anymore without having to pay "
+            "the full costs of €" + str(self.fine) + ". Also note that "
+            "you will be unable to re-register. Note: If you have any "
+            "COVID-19 symptoms you will not have to pay these fees. "
+            "Let us know via info@thalia.nu that this is this reason "
+            "for your cancellation."
+        )
 
     @property
     def after_cancel_deadline(self):
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -182,6 +182,15 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
     tpay_allowed = models.BooleanField(_("Allow Thalia Pay"), default=True)
 
     @property
+    def cancel_too_late_message(self):
+        return "Cancellation isn\'t possible anymore without having to pay " \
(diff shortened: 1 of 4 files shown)


website/events/templates/events/event.html
```diff
@@ -217,11 +217,8 @@ <h1 class="text-center section-title">{{ event.title }}</h1>
                                             {% endblocktrans %}
                                         {% endif %}
                                         {% if event.after_cancel_deadline and registration.queue_position == 0 %}
-                                            {% blocktrans trimmed with costs=event.fine %}
-                                                Cancellation isn't possible anymore without having to pay the full costs of €{{ costs }}.
-                                                Also note that you will be unable to re-register.{% endblocktrans %}
-                                            {% blocktrans trimmed %}
-                                                Note: If you have any COVID-19 symptoms you will not have to pay these fees. Let us know via info@thalia.nu that this is this reason for your cancellation.
+                                            {% blocktrans trimmed with no_cancellation=event.cancel_too_late_message%}
+                                                {{ no_cancellation }}
```
se-bastiaan, line 221: Apparently this is not working, since the tests are failing on this text not showing up.

website/events/models/event.py
```diff
@@ -178,6 +178,17 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
         null=True,
     )
 
+    @property
+    def cancel_too_late_message(self):
+        return (
```
se-bastiaan, line 183: I would put the translation in the text here instead of the view, that would make it available in the API too

se-bastiaan (changes requested):

website/events/tests/test_views.py
```diff
@@ -433,9 +433,10 @@ def test_registration_cancel_after_deadline_warning(self):
         self.event.save()
         EventRegistration.objects.create(event=self.event, member=self.member)
         response = self.client.get("/events/1/")
+        print(response.rendered_content)
```
se-bastiaan, line 436: Not sure if this print belongs here? :p

LCKnol: oops, forgot about that!

website/events/models/event.py
se-bastiaan: @LCKnol this is still an improvement I think would be good, if you think it is useless just let me know.

website/events/templates/events/event.html
se-bastiaan: There is no translation here anymore

website/events/models/event.py
se-bastiaan:
```suggestion
        return _(
```

website/events/templates/events/event.html
```diff
@@ -217,11 +217,8 @@ <h1 class="text-center section-title">{{ event.title }}</h1>
...
                                         {% endif %}
                                         {% if event.after_cancel_deadline and not registration.queue_position %}
-                                            {% blocktrans trimmed with costs=event.fine %}
-                                                Cancellation isn't possible anymore without having to pay the full costs of €{{ costs }}.
-                                                Also note that you will be unable to re-register.{% endblocktrans %}
-                                            {% blocktrans trimmed %}
-                                                Note: If you have any COVID-19 symptoms you will not have to pay these fees. Let us know via info@thalia.nu that this is this reason for your cancellation.
+                                            {% blocktrans trimmed with no_cancellation=event.cancel_too_late_message%}
+                                                {{ no_cancellation }}
                                             {% endblocktrans %}
```
se-bastiaan, line 222:
```suggestion
                                            {{ event.cancel_too_late_message }}
```

se-bastiaan (approved):

PR #1121, opened 2020-05-27
Simplify if statement

pingiun:
Closes #1120 

### Summary
Simple fix of codeclimate issue

### How to test
Event permissions should be covered by the test suite

- [ ] Check coverage report

PR #1512, opened 2021-01-25
event registration now saves changes

### LINKED ISSUE #1345: Can't set registration information fields from admin (for non-member registrations)

mtsatsev:
Closes  #1345

### Summary
Organizers of events should be able to edit the members' fields at their events even for non-members. Now they can.

### How to test
1. Add a manual event registration.
1. Edit the information fields of the participants.
1. It works.

### DIFF
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -186,8 +186,8 @@ def update_registration(
         name = registration.name
 
     if (
-        not event_permissions(member, event, name)["update_registration"]
-        or not field_values
+        (not event_permissions(member, event, name)["update_registration"]
+        or not field_values) and is_organiser(member,event)
     ):
         return
 
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -186,8 +186,8 @@ def update_registration(
         name = registration.name
 
     if (
-        (not event_permissions(member, event, name)["update_registration"]
-        or not field_values) and is_organiser(member,event)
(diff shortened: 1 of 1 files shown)


website/events/services.py
```diff
@@ -187,8 +187,8 @@ def update_registration(
 
     if (
         not event_permissions(member, event, name)["update_registration"]
-        or not field_values
-    ):
+        and is_organiser(member, event)
```
se-bastiaan, line 190: When I am not an organiser this will always allow editing the event fields from the user perspective.

se-bastiaan: The change should probably in `event_permissions`. If you look at this PR: https://github.com/svthalia/concrexit/pull/1121/files, you see that the if statements have not the same meaning anymore. 

First, it checked for either an authenticated member or a name. Now it always requires a member first. That is more likely to be the problem here.

This change does not look like the right solution without causing new problems.

se-bastiaan (changes requested):

website/events/services.py
mtsatsev: Ok I will look it up again on Monday.

se-bastiaan:

PR #1523, opened 2021-02-09
Try out github environments

pingiun:
I wanna see how this new GitHub feature works: https://github.blog/changelog/2020-12-15-github-actions-environments-environment-protection-rules-and-environment-secrets-beta/

It seems like it is a solution to every developer on this repo being able to use the secrets


se-bastiaan:

PR #1529, opened 2021-02-15
Add shortlink support

### LINKED ISSUE #1528: Dynamic link to promotion whatsapp group
The board is moving the whatsapp promotion to one group. Whatsapp groups have a maximum of 256 members so there should be multiple groups.

To avoid having to send new links when the old group is full we would like something like whatsapp.thalia.nu or thalia.nu/whatsapp that always links to the newest group. 

The alternative is to deal with multiple links, ugh.

pingiun:
Closes #1528 

### Summary
Add support for short links that redirect to a (longer) link

image
image

image

### How to test
1. Go to shortlink admin
2. Add shortlink with a slug
3. Visit https://website/slug
4. Get redirected depending on the settings


website/thaliawebsite/views.py
```diff
@@ -19,3 +27,11 @@ def dispatch(self, request, *args, **kwargs) -> HttpResponse:
...
             return HttpResponseForbidden("This is not for you")
         raise Exception("Test exception")
+
+
+def short_url(request, slug):
+    link = get_object_or_404(ShortLink, key=slug)
+    if not link.immediate:
+        return render(request, "shortlink.html", {"link_name": link.key, "link_url": link.url})
+
+    return HttpResponseRedirect(link.url)
```
se-bastiaan: This could just as well be a class-based view. Just like everything else in this file.

se-bastiaan (changes requested):
I think it would be better to place these models and the view in a separate app instead of putting this in thaliawebsite. There is a reason the singlepages are in a different place and that this is the first model that is created here.

website/thaliawebsite/admin.py
```diff
@@ -2,5 +2,6 @@
 from django.contrib import admin
 from django.utils.translation import gettext_lazy as _
 
+
```
se-bastiaan, line 5:
```suggestion
```

website/thaliawebsite/settings.py
```diff
@@ -402,6 +402,7 @@ def from_env(
     "registrations.apps.RegistrationsConfig",
     "payments.apps.PaymentsConfig",
     "singlepages.apps.SinglepagesConfig",
+    "shortlink.apps.ShortlinkConfig",
```
se-bastiaan, line 405:
```suggestion
    "shortlinks.apps.ShortlinksConfig",
```

All packages are plural.

website/shortlink/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class ShortLink(models.Model):
+    key = models.CharField(
+        primary_key=True,
+        max_length=32,
+        validators=[validators.validate_slug],
+        verbose_name=_("Name for the url"),
+        help_text=_(f"This is what you use after https://{settings.SITE_DOMAIN}/"),
+    )
```
se-bastiaan, line 15: Since this value is used as a primary key it is impossible to change it. I think it would be better to just force it to be unique, but not use it as pk.

website/shortlink/models.py
```diff
@@ -0,0 +1,33 @@
...
+from django.db import models
+
+from django.utils.translation import gettext_lazy as _
+
+
+class ShortLink(models.Model):
+    key = models.CharField(
+        primary_key=True,
+        max_length=32,
+        validators=[validators.validate_slug],
```
se-bastiaan, line 12: You could also just use a SlugField https://docs.djangoproject.com/en/3.1/ref/models/fields/#slugfield

website/shortlink/models.py
```diff
@@ -0,0 +1,33 @@
...
+        verbose_name=_("Name for the url"),
+        help_text=_(f"This is what you use after https://{settings.SITE_DOMAIN}/"),
+    )
+
+    url = models.TextField(
```
se-bastiaan, line 17: This can just be a URLField https://docs.djangoproject.com/en/3.1/ref/models/fields/#urlfield

website/shortlink/views.py
```diff
@@ -0,0 +1,22 @@
...
+            raise Http404()
+        if self.link.immediate:
+            return HttpResponseRedirect(self.link.url)
+        return super().dispatch(request, *args, **kwargs)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["link_key"] = self.link.key
+        context["link_url"] = self.link.url
+        return context
```
se-bastiaan, line 22:
```suggestion
class ShortLinkView(DetailView):
    model = ShortLink
    slug_field = 'key'
    context_object_name = 'link'
    template_name = "shortlink/confirm.html"

    def render_to_response(self, context, **kwargs):
        if self.object.immediate:
            return HttpResponseRedirect(self.object.url)
        return super().render_to_response(context, **kwargs)
```

website/thaliawebsite/templates/shortlink.html
```diff
@@ -0,0 +1,17 @@
+{% extends "base.html" %}
+{% load static i18n %}
+
+{% block title %}Shortlink for {{ link_key }} — {{ block.super }}{% endblock %}
+{% block opengraph_title %}Shortlink for {{ link_key }} — {{ block.super }}{% endblock %}
```
se-bastiaan, line 5: This template belongs in the shortlink package

website/shortlink/models.py
```diff
@@ -0,0 +1,33 @@
...
+from django.utils.translation import gettext_lazy as _
+
+
+class ShortLink(models.Model):
+    key = models.CharField(
+        primary_key=True,
+        max_length=32,
+        validators=[validators.validate_slug],
+        verbose_name=_("Name for the url"),
+        help_text=_(f"This is what you use after https://{settings.SITE_DOMAIN}/"),
```
se-bastiaan, line 14: Using this setting in the help text will require migrations in a different environment or when the value is changed.

website/shortlink/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class ShortlinkConfig(AppConfig):
```
se-bastiaan, line 4:
```suggestion
class ShortLinkConfig(AppConfig):
```

Consistent naming

se-bastiaan (dismissed):
Overall it would work well I think, there are some things that should still be changed. The template and the key of the model being the important ones.

I also really would like the package to be plurar since that matches most of the other packages. But that is optional.

website/shortlink/models.py
pingiun: Good point, I think it would be that bad if we had no primary key here?

website/shortlinks/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+    immediate = models.BooleanField(
+        default=False,
+        verbose_name=_("redirect without information page"),
+        help_text=_(
+            "Make sure to only do this when users expect an immediate redirect, if "
+            "they expect a Thalia site a redirect will be confusing."
+        ),
+    )
+
```
Jorritboer, line 31: I think a description would be nice. For example for the whatsapp link it could say to email info@thalia.nu if the whatsapp group is full

pingiun: Already feature creep smh

se-bastiaan: No but really, the idea is to have a redirect. If there should be an explanation it would be better to introduce an environment variable that is just used on an explanation page. Not part of the requirements :p

Jorritboer: Yes okay good point. I was a bit distracted by the redirection page and thought we might as well add it there. But it's true that's not what this is about.

Jorritboer (approved):

PR #1531, opened 2021-02-15
Add the option to disable Thalia pay for events

### LINKED ISSUE #1422: Disable Thalia Pay payments for specific events
There might be events where Thalia Pay should not be available (study trip, events where the payment amount is variable like the orientation bbq, events where registration is not first come first serve like wine tasting or committee lunch)

An option in the events admin that sets whether Thalia Pay is allowed for this event. The default should be true. When false, the payment button should not appear.

Probably we should also add something to the Payable interface about this so we can implement the same check in the process_payment code.

### Motivation
Thalia Pay is quite critical and should work as intended always without too much manual hassle.

Manually removing payments that we do not want.

9Y0:
Closes #1422 

### Summary
You can now disable Thalia Pay for specififc events

### How to test
1. Create an event with a price and uncheck `Is Thalia Pay payment allowed for this event`
2. Make sure you have Thalia Pay enabled
3. Notice that there is no longer a `Pay with Thalia Pay` button on the event page


website/events/models/event.py
```diff
@@ -179,6 +179,10 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
         null=True,
     )
 
+    tpay_allowed = models.BooleanField(
+        _("Is Thalia Pay payment allowed for this event"), default=True, blank=False
```
se-bastiaan, line 183: Boolean fields can never be blank. They are either False or True.

se-bastiaan:
deploy review

se-bastiaan:
remove deploy

website/events/models/event.py
```diff
@@ -179,6 +179,10 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
         null=True,
     )
 
+    tpay_allowed = models.BooleanField(
+        _("Is Thalia Pay payment allowed for this event"), default=True
```
se-bastiaan, line 183: image
This text looks a bit weird :p

se-bastiaan (changes requested):
Code looks good! Maybe discuss the text of the checkbox?

website/events/models/event.py
9Y0: Yeah that indeed looks weird.
I think `Allow Thalia Pay payments` would be better. We could also go for `Allow Thalia Payments` but I'm not sure if we want to use that terminology (maybe @JobDoesburg can answer this)

se-bastiaan: I think the first one is best. Since 'Thalia Pay' is kind of the 'thing'.

se-bastiaan: Does that make sense? I don't know :p

JobDoesburg: Or just "Allow Thalia Pay"

se-bastiaan (approved):

JobDoesburg (approved):

Issue #1443, opened 2020-12-07
Allow event registering for open events


mark-boute:
From https://gitlab.science.ru.nl/thalia/concrexit/-/issues/711

### One-sentence description
Allow event registering for open events

### Motivation
So that the personal calendar can contain these events, notifications only get send to these people and open events have an indication how many people are going to go to the event.

### Desired functionality
The normal registration functionality. But it is not registration, more like Facebooks

### Suggested implementation
We should be able to add an option that if there is no registration required that you can still register.

PR #1551, opened 2021-02-22
Add calendar coloring based on registration

### LINKED ISSUE #1116: Use a different colour for waiting list events
In the event overview calendar, the dot in front of the event is colored as if you are registered. And this is confusing.

Use a different colour for events you're on the waiting list for.

### Motivation
Better user experience.

There is not really any alternative

### Additional context
Created based on #1007

ColonelPhantom:
Closes #1116

### Summary

I implemented changes in the calendar view, so that the colours of the event indicators/backgrounds are a lot richer:

- As in #1116, when pending, the event clearly indicates this (by being yellow).
- Events without registration are also marked clearly, by being magenta. It might be good to think if this is better than it being the same color as either a registered or registerable event. #1443 would also solve this.
- Events for which you are registered become green.
- Events with closed registration are red. This is rather hard to distinguish from an unpublished event, which is also red. It might be wise to give another color to one of these.

Here are some screenshots:
screen_2021-02-22-20:56:03
screen_2021-02-22-20:56:09
screen_2021-02-22-20:56:30
screen_2021-02-22-20:57:10

### How to test

- The main functionality is easy to test, simply open the event list/calendar view.
- The change modifies the CalendarJS serializer, I am not sure if this is used elsewhere (like in the app). The flattening I did for the class names (to fix the event indicators in the list view, which are even an issue on current production), might break anything else that depends on it.

### DIFF
--- a/website/events/api/calendarjs/serializers.py
+++ b/website/events/api/calendarjs/serializers.py
@@ -13,19 +13,20 @@ class EventsCalenderJSSerializer(CalenderJSSerializer):
         return reverse("events:event", kwargs={"pk": instance.id})
 
     def _class_names(self, instance):
-        class_names = ["regular-event"]
         if self.context["member"] and services.is_user_registered(
             self.context["member"], instance
         ):
             if not services.user_registration_pending(self.context["member"], instance):
-                class_names.append("pending-registration")
+                return ["regular-event-pending-registration"]
             else:
-                class_names.append("has-registration")
+                return ["regular-event-has-registration"]
         elif not instance.registration_required:
-            class_names.append("no-registration")
+            return ["regular-event-no-registration"]
         elif not instance.registration_allowed:
-            class_names.append("registration-closed")
-        return class_names
+            return ["regular-event-registration-closed"]
+        else:
+            # I think this handles the case that registration is needed, but the user is not registered?
+            return ["regular-event-not-registered"]
 
 
 class UnpublishedEventsCalenderJSSerializer(CalenderJSSerializer):
--- a/website/events/api/calendarjs/serializers.py
+++ b/website/events/api/calendarjs/serializers.py
@@ -16,17 +16,15 @@ class EventsCalenderJSSerializer(CalenderJSSerializer):
         if self.context["member"] and services.is_user_registered(
             self.context["member"], instance
         ):
-            if not services.user_registration_pending(self.context["member"], instance):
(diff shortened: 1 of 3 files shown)


se-bastiaan:
I think your changes to the serializer are fine. What am not sure about if is our members will still  understand all these colours. I am having a hard time understanding it without reading the explanation 😅

JobDoesburg:
> I think your changes to the serializer are fine. What am not sure about if is our members will still understand all these colours. I am having a hard time understanding it without reading the explanation 😅

I agree with Sebas. Maybe it would be good to use grey tones for unpublished events or events that are closed registration (because you cannot attend them), magenta tones for events that don't have registration or that you are registered for, and we can keep the yellow for waiting list events.

ColonelPhantom:
I can definitely see what you mean, haha.

Another option might be to add a short description, although it might also be a good idea to have the same color for events that you are registered for and events that do not require registration. Like I said before, #1443 would also help a lot.

se-bastiaan:
I think it is best to let Job decide the colours for this, I am happy to give my opinion about it :)

JobDoesburg:
I stay with this suggestion for colors (unless someone disagrees)

> Maybe it would be good to use grey tones for unpublished events or events that are closed registration (because you cannot attend them), magenta tones for events that don't have registration or that you are registered for, and we can keep the yellow for waiting list events.

JobDoesburg (changes requested):

JobDoesburg:
So I would say:

- Light grey for unpublished events
- Dark grey for events that are closed registration
- Regular magenta for events that are open (no registration or need registration but still open for registration)
- Darker magenta for events that require registration and that you are actually registered for
- Yellow for events that you are on the waiting list for

ColonelPhantom:
In the end, I came up with this (unfortunately without seeing the suggestions by @JobDoesburg but I did run this by him):
- the original red for unpublished events (so not modified by me at all)
- transparent gray or no indicator for closed registration
- regular gray for open registration
- magenta for registered
- yellow for pending

screen_2021-03-29-21:31:59
screen_2021-03-29-21:32:11
screen_2021-03-29-21:33:40
screen_2021-03-29-21:34:55

ColonelPhantom:
Turns out I forgot to commit and push the color changes, that's done now.

JobDoesburg (approved):

PR #1552, opened 2021-02-24
Implement cache for the payment admin

### LINKED ISSUE #1397: Speed up the PaymentUserAdmin
The PaymentUserAdmin currently does a query for each and every user whether they are allowed to use Thalia Pay and calculate their Tpay balance. This is very very very slow.

We should improve the queries in some way that it is faster. It can all be done by the database so it must be possible. The problem is now that we do the same query many times for each user separately, while we should be able to somehow do this query immediately at once.

pingiun:
Closes #1397 

### Summary
Add a cache that is prefetched to the payment user admin

### How to test
1. Go to payment user admin
2. Is it fast?
Hard to test without enough payment users


pingiun:
Maybe we wanna disable the cache when calculating batches?

JobDoesburg:
We still need to change the admin itself to display the cached fields, right?

pingiun:
> We still need to change the admin itself to display the cached fields, right?

Currently the cached fields are always used (if available). That's why I commented about disabling in certain cases, although maybe it's easier to enable them in the admin instead

JobDoesburg:
> Currently the cached fields are always used (if available). That's why I commented about disabling in certain cases, although maybe it's easier to enable them in the admin instead

Ah right now I look at it again I understand it. I think this looks good, but I'm also curious what our pro-developer @se-bastiaan thinks of this :)

website/payments/signals.py
```diff
@@ -21,3 +21,32 @@ def give_new_users_tpay_permissions(instance, created, **kwargs):
         # after this signal call (triggered by the model's `save()`), a
         # `save_related()` will be called which will override the permissions
         # again.
+
+
+@suspendingreceiver(post_save, sender="payments.BankAccount")
+def clear_enabled_cache(sender, instance, **kwargs):
```
JobDoesburg, line 27: Is there another way to trigger this, for example if we change the global settings?

website/payments/signals.py
```diff
@@ -21,3 +21,32 @@ def give_new_users_tpay_permissions(instance, created, **kwargs):
...
+    cache = PaymentUserCache.objects.filter(user=instance.owner).first()
+    if cache is not None:
+        cache.enabled = None
+        cache.save()
+
+
+@suspendingreceiver(post_save, sender="auth.Permission")
+def clear_allowed_cache(sender, instance, **kwargs):
```
JobDoesburg, line 35: idem

website/payments/signals.py
```diff
@@ -21,3 +21,32 @@ def give_new_users_tpay_permissions(instance, created, **kwargs):
...
+
+@suspendingreceiver(post_save, sender="payments.BankAccount")
+def clear_enabled_cache(sender, instance, **kwargs):
+    cache = PaymentUserCache.objects.filter(user=instance.owner).first()
+    if cache is not None:
+        cache.enabled = None
+        cache.save()
+
+
+@suspendingreceiver(post_save, sender="auth.Permission")
```
JobDoesburg, line 34: This one is not triggered when I use the admin action to allow tpay for a user (which is the default way to allow tpay for a user)

JobDoesburg:
Maybe it would also be good to not only invalidate, but also recalculate the caches when, for example, a new Payment or Bank Account is created. Otherwise, the first time loading the PaymentUserAdmin can still be very slow

website/payments/signals.py
pingiun: that admin just changes the permission right?

JobDoesburg: The admin action calls this: https://github.com/svthalia/concrexit/blob/d2741f10fb6395e58862aaa888ee3c30d2c1b1fe/website/payments/admin.py#L715-L724

se-bastiaan: I'd say that is specifically written to not do a full update on a queryset. So this is weird.

website/payments/signals.py
se-bastiaan: Maybe a management command that is executed on startup?

pingiun:
Closed in favour of #1587

PR #1562, opened 2021-03-02
Add build commit in Nix build

### LINKED ISSUE #1556: Reintroduce build commit environment variable
### Describe the bug
The nix deployment has removed the build commit from the website, because the environment variable isn't set anymore. The commit hash should be remembered somewhere and set as an environment variable

### How to reproduce
1. Go to staging website
2. Check html source
3. Build commit not visible

### Expected behaviour
The build commit hash is shown in the html source

### Additional context
This was removed accidentally because of the migration to nixos

pingiun:
Closes #1556 

### Summary
Reintroduce build commit environment variable

### How to test
1. Deploy via nix
2. Check source of any page
3. See build commit information

### DIFF
--- a/scripts/build-nix.sh
+++ b/scripts/build-nix.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Convenience script to allow you to build concrexit via nix without
+# needing to remember adding the args
+
+if [ -z "$1" ] ; then
+    attribute="machine"
+else
+	attribute=$1
+fi
+
+if [[ $# -gt 1 ]] ; then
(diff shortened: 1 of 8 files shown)


scripts/build-nix.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
wdoeland, line 1: Maybe it would be nice to use this script in `scripts/test-vm.sh`?

wdoeland (approved):

scripts/build-nix.sh
pingiun: good idea

se-bastiaan (approved):

PR #1574, opened 2021-03-15
Allow registration for events that do not require it

### LINKED ISSUE #1443: Allow event registering for open events

ColonelPhantom:
Closes #1443

### Summary
I added the possibility to register for events that do not require it. That should make it possible to get an indication of how many people will be there, but also who, and also makes it possible to make it show up in your own calendar.

### How to test
1. Go to an event that does not require registration.
2. Click on "register".
3. Check that the user is now registered.

I did not test this very extensively, and it might be rough around the edges, so I'd be glad to hear about any deficiencies.

### DIFF
--- a/website/events/templates/events/event.html
+++ b/website/events/templates/events/event.html
@@ -155,15 +155,25 @@
                                     </form>
                                 {% endif %}
                                 {% if permissions.create_registration %}
-                                    <p>
-                                        {% url 'singlepages:event-registration-terms' as terms_url %}
-                                        {% blocktrans trimmed %}
-                                            By registering,
-                                            you confirm that you have read the
-                                            <a target="_blank" href="{{ terms_url }}">terms and conditions</a>,
-                                            that you understand them and that you agree to be bound by them.
-                                        {% endblocktrans %}
-                                    </p>
+                                    {% if event.registration_required %}
+                                        <p>
+                                            {% url 'singlepages:event-registration-terms' as terms_url %}
+                                            {% blocktrans trimmed %}
+                                                By registering,
+                                                you confirm that you have read the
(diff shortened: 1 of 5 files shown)


website/events/templates/events/event.html
```diff
@@ -155,15 +155,25 @@ <h1 class="text-center section-title">{{ event.title }}</h1>
...
+                                                <a target="_blank" href="{{ terms_url }}">terms and conditions</a>,
+                                                that you understand them and that you agree to be bound by them.
+                                            {% endblocktrans %}
+                                        </p>
+                                    {% else %}
+                                        <p>
+                                            {% blocktrans trimmed %}
+                                                Even though registration is not required for this event,
+                                                you can still register to give an indication of who will be there
+                                                as well as mark the event as 'registered' in your calendar.
```
se-bastiaan, line 173: This is a really nice solution to solve the problem I call 'is this even a registration' 👌

se-bastiaan:
So I think I did exactly the same changes you made somewhere in the past. And I did not put it into review because I thought that I must be forgetting something because this change is so damn easy.

But if you have the same idea I do believe we can safely say that this is the way to do it and we only should test some situations before merging this.

I mostly wonder how the API and thus the app will handle the registrations. And what will happen if there is a price set for the event, does the Thalia Pay button show up?

se-bastiaan (approved):

PR #1576, opened 2021-03-15
link event to the album title and date

### LINKED ISSUE #1488: Automatically fill in event title and date after selecting event for a photo album
While solving #1363 I thought it'd be nice if the album's title and date were automatically filled in after selecting the event the albums are for. Just like adding an event to a newsletter.

When you select the event the photos are for it automatically fills in the fields title and date.

### Motivation
Saves 5 seconds to look up what the event was called exactly again and what date it was

Tedddoo:
Closes #1488 

### Summary
When an album is created, it is possible to link an event and use it's title and date instead of filling it in yourself.
This also fixes the multilanguage field for title and it's migrations.

### How to test
Add an album without a title or date filled in and select an event
Save and check if the correct date and title get added

Add an album without a title, date or event, this should give an error message
Add an album without a title and without an event, this should give an error message
Add an album without a date and without an event, this should give an error message


website/photos/models.py
```diff
@@ -122,12 +132,35 @@ def get_absolute_url(self):
         """Get url of Album."""
         return reverse("photos:album", args=[str(self.slug)])
 
+    def clean(self):
+        super().clean()
+        errors = {}
+
+        if (self.title is None or len(self.title) is 0) and self.event is None:
```
SebastiaanSafeguard, line 139:
```suggestion
        if not self.title and not self.event:
```

```suggestion
        if not (self.title or self.event):
```

se-bastiaan, line 139: This check can be made easier, since 

```
self.title = None
bool(self.title) == True
self.title = ""
bool(self.title) == True
```

JobDoesburg:
This one should be merged by @Jorritboer imo as t merges to his branch

se-bastiaan (approved):

PR #1587, opened 2021-03-23
Create view for paymentuser to speed up queries

pingiun:
Closes #1552, #1397

### Summary
Create a view that collects the required data for the paymentuser admin in sql instead of python, now the admin page only requires a couple of queries

### How to test
1. Go to the paymentuser admin page
2. It's not slow


pingiun:
Before i go and fix all the tests i would like to know if this approach makes sense or if i got anything blatenly wrong

se-bastiaan:
I think the SQL query looks good enough to me and with some time it is possible to understand it. There is however one completely different topic I want to touch: do we want to consider that Django's model system may change?

I do not see this as a really big problem honestly, since that would probably mean a major version upgrade and will probably also include migrations itself.

How important is this view to have a working Thalia Pay environment? Is it feasible to use it without the view?

JobDoesburg:
> How important is this view to have a working Thalia Pay environment? Is it feasible to use it without the view?

The view seems to replace the PaymentUser model, so it is heavily integrated in Thalia Pay and we can't go without.

I can read the code indeed, however it does feel clumsy to actually only solve a problem of the admin view here.

pingiun:
> do we want to consider that Django's model system may change?

I've also thought about this, but I think the parts I'm using right now are pretty stable. Also an unmanaged model is officially supported, so the way we're doing this is likely to keep working. 

> however it does feel clumsy to actually only solve a problem of the admin view here

it might also be useful in other cases, i haven't benchmarked other things yet tho. something like statistics are now possible, where before it would just be slow

pingiun:
Oh btw the query only looks very big, the database can easily handle this. queries in the view for individual users take less than a millisecond

se-bastiaan:
> > How important is this view to have a working Thalia Pay environment? Is it feasible to use it without the view?
> 
> The view seems to replace the PaymentUser model, so it is heavily integrated in Thalia Pay and we can't go without.

But what happens if you would revert to the option without the view? Could Thalia Pay keep working if you temporarily disable it?

JobDoesburg:
> > > How important is this view to have a working Thalia Pay environment? Is it feasible to use it without the view?
> 
> > 
> 
> > The view seems to replace the PaymentUser model, so it is heavily integrated in Thalia Pay and we can't go without.
> 
> 
> 
> But what happens if you would revert to the option without the view? Could Thalia Pay keep working if you temporarily disable it? Not that it would happen, but I have lost the overview of the project and do not know exactly what the PaymentUser is used for.
> 
> 
> 
> 

That would require a migration I believe. A PaymentUser is a proxy model of Member and is used throughout the payments app for accessing payer objects. For example, checking if a user has Thalia pay enabled is checked by getting the PaymentUser object.

JobDoesburg:
> it might also be useful in other cases, i haven't benchmarked other things yet tho. something like statistics are now possible, where before it would just be slow

This is very true

JobDoesburg:
I think this view definition is the cleanest solution. I'm only wondering, would it maybe be nice to use a library for this?

I found https://pypi.org/project/django-database-view/ and https://pypi.org/project/django-db-views/ online. 

Maybe this improves readability, but it's also a perhaps nasty dependency...

se-bastiaan:
> 
> 
> I think this view definition is the cleanest solution. I'm only wondering, would it maybe be nice to use a library for this?
> 
> I found https://pypi.org/project/django-database-view/ and https://pypi.org/project/django-db-views/ online.
> 
> Maybe this improves readability, but it's also a perhaps nasty dependency...

I prefer the current solution. It is pretty nice imo. And those dependencies basically still need you to write SQL or a variant of it.

pingiun:
> That would require a migration I believe. A PaymentUser is a proxy model of Member and is used throughout the payments app for accessing payer objects. For example, checking if a user has Thalia pay enabled is checked by getting the PaymentUser object.

I could make it so the previous implementation remains, but that would just make two seperate pieces of logic without a "canonical" one. I think having the logic only in sql is better

se-bastiaan (approved):

JobDoesburg:
While working on the sales app I found django's `Subquery()` with `OuterRef()`. 
Perhaps writing a separate manager that annotates payment users is even cleaner.

JobDoesburg:
I think using `django-queryable-properties` for this would be an even nicer approach!

website/payments/migrations/0017_create_view.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.1.7 on 2021-03-23 19:27
+
+import os.path
+
+from django.db import migrations
+
+def slurp(file):
```
JobDoesburg, line 7: nice

JobDoesburg:
I tried this, but it doesn't really work yet (though we do have the speedup, the answers are all incorrect). I think it's because of the SQL syntax (it's saying things like `1 != True`)

https://github.com/svthalia/concrexit/compare/paymentuser-queryable-properties

JobDoesburg:
So perhaps this PR is just best

se-bastiaan:
Is it not possible to fix that then?

JobDoesburg:
> Is it not possible to fix that then?

Maybe @pingiun can have a look. He knows more about SQL implementations than I do, I think. For the sales app I spent quite some time on this and found out that you run into a lot of problems with postgres and sqlite here.  

I do think this should get priority

JobDoesburg:
I have managed to create queryable properties for the balance and `tpay_enabled`. All that's left is `tpay_allowed` but that will be pretty hard probably, as it works with permissions. 

My proposal is to remove the `tpay_allowed` permission over all: by default everyone can use it. Then we add a single BlacklistedPaymentUser model that is a OneToOneRelation with a regular PaymentUser. Thalia Pay is then prohibited if such a relation exists ('the user is on the blacklist'). This tremendously simplifies our codebase, as we also don't have to keep separate code for giving permissions to new users when they are created.
What do you guys think?

pingiun:
I will close this because @JobDoesburg is working on a better solution that's simpler and more maintainable

PR #1589, opened 2021-03-25
Add current member API endpoint

se-bastiaan:
### Summary

This adds the `/api/v2/members/me` endpoint that was missing.

### How to test
1. Go to /api/docs
2. Click on the view and try it out

### DIFF
--- a/website/members/api/v2/serializers/member.py
+++ b/website/members/api/v2/serializers/member.py
@@ -38,6 +38,13 @@ class MemberSerializer(serializers.ModelSerializer):
             return membership.type
         return None
 
+    def update(self, instance, validated_data):
+        profile_data = validated_data.pop("profile")
+        instance.profile = self.fields["profile"].update(
+            instance=instance.profile, validated_data=profile_data
+        )
+        return instance
+
 
 class MemberListSerializer(MemberSerializer):
     class Meta:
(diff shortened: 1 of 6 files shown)


website/members/api/v2/views.py
```diff
@@ -46,3 +47,8 @@ class MemberDetailView(RetrieveAPIView):
         DjangoModelPermissionsOrAnonReadOnly,
     ]
     required_scopes = ["activemembers:read"]
+
+
+class MemberCurrentView(MemberDetailView):
```
JobDoesburg, line 52: Shouldn't this require the member profile read scope?

se-bastiaan: I was thinking of using the same scope as regular users. But actually a good idea to make it more restrictive since an app might not need access to the full members database. Good catch.

Also, I noted that the scope of this piece of code is wrong anyways.

website/members/api/v2/serializers/member.py
```diff
@@ -47,3 +54,41 @@ class Meta:
             "membership_type",
             "profile",
         )
+
+
+class MemberCurrentSerializer(MemberSerializer):
```
JobDoesburg, line 59: I think MemberDetailSerializer would be better (as this serializes a Member in detail, not necessarily the current member)

JobDoesburg (changes requested):

website/members/api/v2/serializers/member.py
se-bastiaan: No this is always the current member, you will never get this in any other case. This information should never be accidentally exposed in any other part of the API.

JobDoesburg: Well the serializer itself doesn't care to serialize other member objects. But it's fine by me

JobDoesburg (approved):

PR #1595, opened 2021-03-29
Guarantee mandate no. uniqueness

### LINKED ISSUE #1508: Change mandate no. increment to guarantee uniqueness
Consider https://github.com/svthalia/concrexit/blob/56ef628318d5bc4c4bbdc28a6eeac2d6c73cc218/website/payments/views.py#L44

In case an admin might manually remove a mandate (rather than invalidate it), manually change a user's mandate id, the newly generated mandate id might not always be unique. For example, if a user has mandate no. 1234-1 and 1234-3, the next number that will be generated is 1234-3 as well (because we take the `count() + 1`).

Solution: take the max mandate no. if it is of the form `<id>-<i>`, the new mandate no. will be `<id>-<i+1>`, else use the fallback we have already implemented. 

Other solution: do an additional check that the newly generated reference number is in fact really unique.

9Y0:
Closes #1508 

### Summary
The mandate no. generated by the server is now always unique

### How to test
1. Add a direct debit bankaccount
2. Change the mandate no. (in the admin) to `<user pk>-15`
3. Add an extra direct debit bankaccount
4. Notice that the new mandate no. is `<user-pk>-16`

### DIFF
--- a/website/payments/views.py
+++ b/website/payments/views.py
@@ -35,22 +35,9 @@ class BankAccountCreateView(SuccessMessageMixin, CreateView):
     success_url = reverse_lazy("payments:bankaccount-list")
     success_message = _("Bank account saved successfully.")
 
-    def _derive_mandate_no(self) -> str:
-        accounts = (
-            BankAccount.objects.filter(
-                owner=PaymentUser.objects.get(pk=self.request.member.pk)
-            )
-            .exclude(mandate_no=None)
-            .filter(mandate_no__regex=BankAccount.MANDATE_NO_DEFAULT_REGEX)
-        )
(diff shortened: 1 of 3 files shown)


se-bastiaan (dismissed):

website/payments/views.py
```diff
@@ -36,15 +36,17 @@ class BankAccountCreateView(SuccessMessageMixin, CreateView):
     success_message = _("Bank account saved successfully.")
 
     def _derive_mandate_no(self) -> str:
```
JobDoesburg, line 38: Maybe this should be moved to services

se-bastiaan (approved):

PR #1596, opened 2021-03-29
Prevent API from crashing when throttling on the anonymous user

### LINKED ISSUE #1583: Anonymous API calls crash because of throttling
### Describe the bug

Anonymous API calls crash because of throttling.

### How to reproduce
1. Create an API application with server credentials without user attached to it
2. Do an API call
3. Instant crash because the throttling expects a user 

### Expected behaviour
Not crash, but do throttle

se-bastiaan:
Closes #1583

### Summary
If you were using a the API using client credentials it would not work if no user has been attached to the application.

### How to test
1. Create an application with client_credentials
2. Use the credentials
3. Use the API
4. It crashes instantly

### DIFF
--- a/website/thaliawebsite/api/throttling.py
+++ b/website/thaliawebsite/api/throttling.py
@@ -1,0 +1,38 @@
+from rest_framework.throttling import SimpleRateThrottle
+
+
+class AnonRateThrottle(SimpleRateThrottle):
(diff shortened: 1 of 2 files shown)


Luca-Castelnuovo (approved):
Looks wonderful.

Issue #1307, opened 2020-10-06
Cannot remove scheduled notifications


se-bastiaan:
### Describe the bug
Cannot remove scheduled notifications

### How to reproduce
Steps to reproduce the behaviour:
1. Go to a notification
2. Delete it
3. Get a gateway error

### Expected behaviour
Message is successfully removed

se-bastiaan:
I also think that messages are not removed when an event is unpublished but I could not confirm that behaviour

pingiun:
I believe this is similar to #1328 in that the events model uses the antipattern of checking along instance revisions in the clean() method instead of only checking the current state of the instance. 

Although I am not sure how the event model is related to deleting scheduled notifications (afaik there is no connection between the two models). The stack trace of the crash does mention the events model though:
```
Current thread 0x0000700011ff6000 (most recent call first):
  File "/Users/jelle/Library/Caches/pypoetry/virtualenvs/concrexit-8BPMTAC2-py3.8/lib/python3.8/site-packages/django/utils/deconstruct.py", line 16 in __new__
  File "/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/copyreg.py", line 91 in __newobj__
  File "/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/copy.py", line 264 in _reconstruct
  File "/usr/local/Cellar/python@3.8/3.8.6/Frameworks/Python.framework/Versions/3.8/lib/python3.8/copy.py", line 102 in copy
  File "/Users/jelle/Library/Caches/pypoetry/virtualenvs/concrexit-8BPMTAC2-py3.8/lib/python3.8/site-packages/django/db/models/expressions.py", line 348 in copy
  File "/Users/jelle/Library/Caches/pypoetry/virtualenvs/concrexit-8BPMTAC2-py3.8/lib/python3.8/site-packages/django/db/models/expressions.py", line 247 in resolve_expression
  File "/Users/jelle/Library/Caches/pypoetry/virtualenvs/concrexit-8BPMTAC2-py3.8/lib/python3.8/site-packages/django/db/models/lookups.py", line 85 in process_lhs
  File "/Users/jelle/Library/Caches/pypoetry/virtualenvs/concrexit-8BPMTAC2-py3.8/lib/python3.8/site-packages/django/db/models/lookups.py", line 171 in process_lhs
[...]

PR #1453, opened 2020-12-10
Move event changes validation to the admin form

pingiun:
Closes #1307

### Summary
Includes a test to make sure deleting push notifications associated with
this event will not crash python. This commit fixes #1307 by not saving
anything in the __init__ of the event model, as this seems to break
deleting things from foreign keys.

### How to test
Steps to test the changes you made:
1. Create event with a registration and start date more than 1 hour in the future
2. Check that a push notification is created for this event
3. Delete push notification
4. Python doesn't crash

PR #1609, opened 2021-04-05
Allow changing event registration information after publishing

### LINKED ISSUE #1607: Changing registration start after event publishing is always blocked
### Describe the bug

When trying to change the registration start date after publishing the event it will never be possible to change it anymore even though registration _has not_ started yet.

### How to reproduce
1. Create an event with registrations in the _future_
2. Publish the event 
3. Try to change the registration start date
4. See error that says the registration date cannot be changed because registration has already opened

### Expected behaviour

It should actually matter what the date is and if the registration is open for this error to appear.

### Screenshots

Screenshot 2021-04-02 at 21 17 26

9Y0:
Closes #1607 

### Summary
It is now possible to change the price and the registration start and price as long as the registration has not started yet

### How to test
1. Create an event with registrations in the future
2. Publish the event
3. Try to change the registration start date or price
4. Profit

### DIFF
--- a/website/events/admin.py
+++ b/website/events/admin.py
@@ -144,6 +144,11 @@ class EventAdmin(DoNextTranslatedModelAdmin):
         (_("Extra"), {"fields": ("slide", "documents"), "classes": ("collapse",)}),
     )
 
+    def get_form(self, request, obj=None, change=False, **kwargs):
+        form = super().get_form(request, obj, change, **kwargs)
+        form.clean = lambda form: form.instance.clean_changes(form.changed_data)
+        return form
+
     def overview_link(self, obj):
         return format_html(
             '<a href="{link}">{title}</a>',
--- a/website/events/admin.py
+++ b/website/events/admin.py
@@ -144,11 +144,6 @@ class EventAdmin(DoNextTranslatedModelAdmin):
         (_("Extra"), {"fields": ("slide", "documents"), "classes": ("collapse",)}),
     )
 
-    def get_form(self, request, obj=None, change=False, **kwargs):
-        form = super().get_form(request, obj, change, **kwargs)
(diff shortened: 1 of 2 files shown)


website/events/admin.py
```diff
@@ -144,11 +144,6 @@ class EventAdmin(DoNextTranslatedModelAdmin):
         (_("Extra"), {"fields": ("slide", "documents"), "classes": ("collapse",)}),
     )
 
-    def get_form(self, request, obj=None, change=False, **kwargs):
-        form = super().get_form(request, obj, change, **kwargs)
```
se-bastiaan, line 148: Would changes this not reverse the fix @pingiun tried to implement in https://github.com/svthalia/concrexit/pull/1453?

9Y0: In the summary of that PR he mentioned:
> This commit fixes #1307 by not saving anything in the init of the event model, as this seems to break deleting things from foreign keys.

so I assumed that the problem was that we were saving fields in the `__init__` method of `Event`.
If the actual problem was that the `clean` method in the model shouldn't do these checks then you're probably right.
I could try if my approach to checking if you're allowed to change the price/start time would also work in the `clean_changes` method.

pingiun: > If the actual problem was that the clean method in the model shouldn't do these checks then you're probably right.

Conceptually this is the wrong place to check changes, because models should only check their own integrity and not check if changes are correct. 

So if the test I created doesn't fail this isn't technically a wrong solution but it's uglier because you're doing a database lookup in a `clean()` method. 

I think it's best to check changes in the form clean method rather than the model clean

se-bastiaan:
@9Y0 @pingiun does this still need changes?

pingiun (approved):

pingiun:
Very nice solution!

Issue #744, opened 2020-03-12
Thalia Identity guidelines applied in Django backend


thaliawww-service:
In GitLab by @JobDoesburg on Oct 22, 2018, 18:38

### One-sentence description

Thalia Identity guidelines applied in Django backend

### Why?

More 'huisstijl' more better

### Current implementation

Thalia admin is ugly blue-ish.

### Suggested implementation

This could be nice magenta #E62272 and black. Just a CSS thingy I think.

thaliawww-service:
In GitLab by @se-bastiaan on Oct 22, 2018, 18:57

This is technically unfeasible.

thaliawww-service:
In GitLab by @se-bastiaan on Oct 22, 2018, 18:57

closed

thaliawww-service:
In GitLab by @se-bastiaan on Dec 4, 2019, 19:25

reopened

thaliawww-service:
In GitLab by @se-bastiaan on Dec 4, 2019, 19:25

assigned to @DeD1rk

thaliawww-service:
In GitLab by @se-bastiaan on Dec 4, 2019, 19:25

assigned to @JobDoesburg and unassigned @DeD1rk

PR #1614, opened 2021-04-09
Fix colours in the Django admin after 3.2 update

se-bastiaan:
### Summary
Django introduced theming options in version 3.2, however they use some of the same variables 

### How to test
1. Go to the events and pizza admins
2. Check that the theming is correct

### DIFF
--- a/website/pizzas/templates/pizzas/admin/orders.html
+++ b/website/pizzas/templates/pizzas/admin/orders.html
@@ -16,7 +16,6 @@
 
 {% block extrastyle %}
     {{ block.super }}
-    <link rel="stylesheet" type="text/css" href="/static/admin/css/changelists.css">
     {% compress css %}
         <link href="{% static "pizzas/css/admin.scss" %}" rel="stylesheet"
               type="text/x-scss">
--- a/website/pizzas/templates/pizzas/admin/orders.html
+++ b/website/pizzas/templates/pizzas/admin/orders.html
@@ -16,6 +16,7 @@
 
 {% block extrastyle %}
     {{ block.super }}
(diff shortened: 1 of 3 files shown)


JobDoesburg:
Closes #744 🤯

se-bastiaan:
> Closes #744 🤯

No not yet, this just fixes some problematic overrides

ColonelPhantom:
Looks fine to me, the changes all make sense. With one exception: is manually including the stylesheet in the page necessary? When looking at the code in my browser, I can see it being included twice.

But the changes to the HTML itself and the CSS don't raise any red flags, and testing it seems to work OK.

website/pizzas/templates/pizzas/admin/orders.html
```diff
@@ -16,6 +16,7 @@
 
 {% block extrastyle %}
     {{ block.super }}
+    <link rel="stylesheet" type="text/css" href="/static/admin/css/changelists.css">
```
ColonelPhantom, line 19: This extra `link` seems superfluous, since currently the stylesheet gets included twice.

ColonelPhantom (changes requested):
Looks good apart from this!

website/pizzas/templates/pizzas/admin/orders.html
se-bastiaan: hmmm weird, I do not understand where the stylesheet comes from. But good catch!

se-bastiaan: fixed it!

JobDoesburg (approved):

ColonelPhantom (approved):
Thing I mentioned was fixed, so looks good!

PR #1630, opened 2021-04-26
Change lookup field from pk to slug

### LINKED ISSUE #1626: Expose photo album slug in API v2
### Motivation
Currently the only way to get an album from the api is using its `pk`. It would be nice 
if the new app can accept the `members/photos/:slug/` links as deep links, and share
the album using a link that both the website and app can open.

I think the best solution would be to either replace `api/v2/photos/albums/:pk/` with `api/v2/photos/albums/:slug/`,
or add it with a different path or prefix before the slug.

Jorritboer:
Closes #1626 

### Summary
Changed the lookup field for album api from pk to slug

### How to test
1. Go to /api/v2/photos/albums/{slug}/

### DIFF
--- a/website/photos/api/v2/serializers/album.py
+++ b/website/photos/api/v2/serializers/album.py
@@ -12,7 +12,7 @@ class AlbumSerializer(serializers.ModelSerializer):
         """Meta class for the serializer."""
 
         model = Album
-        fields = ("pk", "title", "accessible", "shareable", "cover", "photos")
+        fields = ("slug", "title", "accessible", "shareable", "cover", "photos")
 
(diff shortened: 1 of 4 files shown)


DeD1rk:
deploy review

DeD1rk (changes requested):
The changes you made look good. However, in the AlbumListView (as well as the detail view), the slugs are not included in the albums. I'd also recommend removing the pk from the serializers altogether, since it shouldn't be used anywhere.

DeD1rk:
remove review

DeD1rk (dismissed):

DeD1rk (approved):

PR #1631, opened 2021-04-27
Small bugfixes to the sales admin

JobDoesburg:
### Summary
After merging #1445 I found some minor bugs in the sales admin. This PR fixes them

### How to test
1. The shift title field in the admin is actually readable
2. It is possible to create a shift without setting any managers
3. It is not possible anymore to lock a shift directly when creating it
4. The order of the fields in the shift admin makes more sense

### DIFF
--- a/website/sales/admin/shift_admin.py
+++ b/website/sales/admin/shift_admin.py
@@ -108,20 +108,25 @@ class ShiftAdmin(admin.ModelAdmin):
         "end",
         "active",
         "product_list",
+        "managers",
         "product_sales",
         "num_orders",
         "total_revenue",
         "locked",
-        "managers",
     )
 
     readonly_fields = (
-        "title",
         "active",
         "total_revenue",
         "num_orders",
         "product_sales",
     )
+
+    def get_readonly_fields(self, request, obj=None):
+        fields = super().get_readonly_fields(request, obj)
+        if not obj:
+            fields += ("locked",)
+        return fields
 
     def get_queryset(self, request):
         queryset = super().get_queryset(request)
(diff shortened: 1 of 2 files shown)


website/sales/models/shift.py
```diff
@@ -49,7 +49,10 @@ class Meta:
     )
 
     managers = models.ManyToManyField(
-        MemberGroup, verbose_name=_("managers"), related_name="manager_shifts"
+        MemberGroup,
+        verbose_name=_("managers"),
+        related_name="manager_shifts",
+        blank=True,
```
se-bastiaan, line 55: It does not make much sense to me that a shift cannot have managers?

website/sales/admin/shift_admin.py
```diff
@@ -108,21 +108,26 @@ class ShiftAdmin(admin.ModelAdmin):
...
+        "managers",
         "product_sales",
         "num_orders",
         "total_revenue",
         "locked",
-        "managers",
     )
 
     readonly_fields = (
-        "title",
```
se-bastiaan, line 119: Title field should thus never be read only?

website/sales/models/shift.py
JobDoesburg: I thought it could be useful when, for example, a treasurer wants to open a shift (although I see they could just fill in the board). It could be useful for people that have the override permission

website/sales/admin/shift_admin.py
JobDoesburg: nope (unless the shift is locked, but then the whole change permission is disabled)

website/sales/models/shift.py
se-bastiaan: I don't think we should do that, it's better to always be able to have at least one group of managers. That way there does not have to be a None check and it is clear who is responsible for it.

JobDoesburg: I agree with that in fact

se-bastiaan (approved):

PR #1634, opened 2021-04-29
Manage npm based packages with yarn

### LINKED ISSUE #1610: Update jQuery version
### Describe the change
Update to the latest jQuery version.

### Motivation
```
Medium | CVE-2019-11358 jQuery before 3.4.0, as used in Drupal, Backdrop CMS, and other products, mishandles jQuery.extend(true, {}, ...) because of Object.prototype pollution | 123
-- | -- | --
Medium | CVE-2020-11022 Regex in its jQuery.htmlPrefilter sometimes may introduce XSS | 1
Medium | CVE-2020-11023 Regex in its jQuery.htmlPrefilter sometimes may introduce XSS
```

### Current implementation
We have an older version.

### Suggested implementation
Download the latest version and put it in the source code.

pingiun:
Closes #1610 

### Summary
Updates jquery and makes sure it stays updated via yarn. For the latter part renovate is enabled for javascript.

### How to test
1. Go to any page
2. See no javascript/css loading errors


pingiun:
Only problem is that this makes the repo a bit bloated...

pingiun:
I'm not 100% sure if renovate would commit node_modules after updating or if it only updates the lockfile

se-bastiaan:
> I'm not 100% sure if renovate would commit node_modules after updating or if it only updates the lockfile

I don't think it will.

Wouldn't it make more sense to have a script that can read the version of the current file somehow and update the minified file when necessary? This would add so many thing to the repo that we do not need at all. There are demos in those npm dependencies.

pingiun:
> Wouldn't it make more sense to have a script that can read the version of the current file somehow and update the minified file when necessary?

Not all minified files have version information. We could maybe add the new package.json and yarn.lock files, but leave out node_modules and create a script that copies from node modules to the right files in Github Actions? I don't know exactly if that is doable but it's probably preferrable to this.

> There are demos in those npm dependencies.

That is an upstream issue xddd /jk

se-bastiaan:
> > Wouldn't it make more sense to have a script that can read the version of the current file somehow and update the minified file when necessary?
> 
> Not all minified files have version information. We could maybe add the new package.json and yarn.lock files, but leave out node_modules and create a script that copies from node modules to the right files in Github Actions? I don't know exactly if that is doable but it's probably preferrable to this.
> 
> > There are demos in those npm dependencies.
> 
> That is an upstream issue xddd /jk
> 
> 

Imo anything else is preferable to this. Even the current situation that requires us to update the dependences manually.

It's pretty easy to add a version header to the minified files, but your solution would also work probably. Just need the build code in github actions.

pingiun:
Yeah you're right, let's close this

Issue #1428, opened 2020-11-27
Introduce API v2


se-bastiaan:
### Is your feature request related to a problem? Please describe.
Our current API code is hard to understand and maintain. 

1. We use Django Rest Framework for our API and every implementation we have uses [viewsets](https://www.django-rest-framework.org/api-guide/viewsets/). This is a way to tightly couple the different views and can work wonderful for the [easy cases](https://github.com/svthalia/concrexit/blob/master/website/activemembers/api/viewsets.py). But [in](https://github.com/svthalia/concrexit/blob/master/website/pizzas/api/viewsets.py#L51) [most](https://github.com/svthalia/concrexit/blob/master/website/members/api/viewsets.py#L41) [situations](https://github.com/svthalia/concrexit/blob/master/website/events/api/viewsets/events.py#L49) it only brings extra checks in our code to accommodate different situations. Especially in cases where there are permissions involved.
2. We used [Django token authentication](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) but have now also authentication using OAuth 2. This last implementation should be a first-class citizen in our API but most of the APIs do not have scoping available. Moreover, they are not readable because the API code _expects_ a user.

### Describe the solution you'd like
To be able to solve this I'd like to introduce the Concrexit API v2. This PR is here for comments, what you think about this and maybe to make some design choices.

Thus:
1. The implementation of the API views should happen using [class-based views](https://www.django-rest-framework.org/tutorial/3-class-based-views/), just like most views in the (newer) apps. This is much more similar to the normal views and is thus easier to grasp if you have to write a view for both users (HTML) and applications (API). No more trying to understand viewsets to try to get them to do what you want. This PR contains an example of such APIs
2. Implement scopes and Django permissions for all views. No longer allow TokenAuthentication for the API. This is going to require a custom implementation that checks both the user permissions and the client scopes. A client logging in with client_credentials will get access as read-only anonymous.
3. Throttle requests based on the amount made per day because the API is basically public due to the awesome Swagger documentation (this also applies to API v1)
4. Pagination should always be enabled by default because some views are really slow now (i.e. https://thalia.nu/api/v1/photos/albums/)
5. It is probably smart to split the regular user APIs from the admin APIs. So for organisers that are changing the payments for events or pizzas for example, this will increase duplicated code but reduce API view and serializer complexity. The same endpoints are currently used for both normal users and for admin changes.

### Motivation
Easy and fun.

### Describe alternatives you've considered
[...]

PR #1652, opened 2021-05-09
Rename pizza models and add API v2

se-bastiaan:
Part of #1428 

### Summary
This PR renames some models in the pizzas package to 'food' and thus may break things in the events package.

It also adds the API v2 for the pizzas package, which I called 'food' instead of 'pizzas'. This because we may decide to change the functionality in the future, and offer other foods. Currently that is not possible due to the way the products work, but the endpoint was already prepared to make this possible by adding the event to the url.

The API also returns a list instead of one single food event, which allows us to run multiple food events at the same time in the future. The only real thing to make this possible is changing the views.

### How to test
1. Go to the API docs on /api/docs and select v2.
2. Check the endpoints and try some things out


se-bastiaan:
> the actual app hasn't been renamed yet, right?

No that would require much more changes

Luca-Castelnuovo:
deploy review

Luca-Castelnuovo (commented):
The code looks fine. Can't test it yet because the deploy is broken.

se-bastiaan:
> 
> 
>     * I am also getting a `CSRF token not set` for the POST method of order.
> 
>     * Do we want the URL's of the pizza app to also change?

* How are you executing the request?
* Pretty much out of scope

website/pizzas/api/v2/views.py
```diff
@@ -0,0 +1,124 @@
...
+        IsAuthenticatedOrTokenHasScope,
+        DjangoModelPermissionsOrAnonReadOnly,
+    ]
+    required_scopes = ["food:read"]
+
+
+class FoodEventOrderDetailView(
+    RetrieveAPIView, CreateAPIView, UpdateAPIView, DestroyAPIView
+):
+    """Returns details of an event registration."""
```
DeD1rk, line 73: You didn't replace this docstring

DeD1rk:
deploy review

DeD1rk:
remove review

DeD1rk:
Sadly the review environment doesn't include the changes from #1664. Maybe we could change it to use the post-merge version when automatic merging is possible. Anyway, @se-bastiaan could you merge #1664 into this? Or do we believe that it works alright just based on the code?

JobDoesburg:
> Sadly the review environment doesn't include the changes from #1664. Maybe we could change it to use the post-merge version when automatic merging is possible. Anyway, @se-bastiaan could you merge #1664 into this? Or do we believe that it works alright just based on the code?

I would do a rebase for sure

DeD1rk:
deploy review

website/pizzas/api/v2/urls.py
```diff
@@ -1,3 +1,26 @@
...
+    FoodEventOrderDetailView,
+)
+
+urlpatterns = [
+    path("food/events/", FoodEventListView.as_view(), name="food-events-list"),
+    path(
+        "food/events/<int:pk>/", FoodEventDetailView.as_view(), name="food-event-detail"
+    ),
+    path(
+        "food/events/<int:pk>/order",
```
se-bastiaan, line 17:
```suggestion
        "food/events/<int:pk>/order/",
```

DeD1rk:
remove review

DeD1rk (dismissed):
Nice stuff

DeD1rk (approved):

Issue #1617, opened 2021-04-12
Profile some popular pages and optimize dayab


JobDoesburg:
### Describe the change
A clear and concise description of what the change should be.

### Motivation
A clear and concise motivation why we should invest in this change.

### Current implementation
If relevant, describe how it's done currently.

### Suggested implementation
Provide (a) suggestion(s) for how we could approach this.

### Additional context
Add any other context or screenshots about the change here.

pingiun:
@ColonelPhantom can you take a look at this?

PR #1672, opened 2021-05-14
Add API v2 events admin

se-bastiaan:
Part of #1428

### Summary
This introduces full functionality to manage events via the v2 API.

### How to test
1. Go to the API docs on /api/docs and select v2.
2. Check the endpoints and try some things out


JobDoesburg:
deploy review

DeD1rk:
@JobDoesburg GitHub actions is a little bit broken; it sometimes skips the deploy/remove review actions

JobDoesburg:
> @JobDoesburg GitHub actions is a little bit broken; it sometimes skips the deploy/remove review actions

Yeah I notice that. Don't see how though

website/events/api/v2/admin/views.py
```diff
@@ -0,0 +1,143 @@
...
+    AdminListAPIView,
+    AdminRetrieveAPIView,
+    AdminCreateAPIView,
+    AdminUpdateAPIView,
+    AdminDestroyAPIView,
+)
+import events.api.v2.filters as normal_filters
+
+
+class EventAdminListAPIView(AdminListAPIView):
```
JobDoesburg, line 28: This API view triggers 205 SQL queries on my db. A prefetch on the activemembergroup would be nice

JobDoesburg:
Shouldn't these API views also have a Django model permission permission class?

se-bastiaan:
> 
> 
> Shouldn't these API views also have a Django model permission permission class?

I tihnk you asked this before already, have you seen the AdminXAPIView base classes?

JobDoesburg:
> > Shouldn't these API views also have a Django model permission permission class?
> 
> I tihnk you asked this before already, have you seen the AdminXAPIView base classes?

Alright, yeah I also thought I already asked that but I couldn't find the message in this thread anymore haha

website/events/api/v2/filters.py
```diff
@@ -43,7 +43,7 @@ def filter_queryset(self, request, queryset, view):
         category = request.query_params.get("category", None)
 
         if category:
-            queryset = queryset.filter(category=category)
+            queryset = queryset.filter(category__in=category.split(","))
```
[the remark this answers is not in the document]
se-bastiaan: No you should just not send spaces, that would be stupid

website/events/models/event.py
```diff
@@ -448,7 +442,7 @@ def save(self, **kwargs):
 
             if start_reminder_time > timezone.now():
                 start_reminder.title_en = "Event"
-                start_reminder.body_en = f"'{self.title_en}' starts in 1 hour"
+                start_reminder.body_en = f"'{self.title}' starts in 1 hour"
```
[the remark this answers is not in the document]
se-bastiaan: Which variables? The ones for the reminder still have the language field, those are not from this app.

website/events/api/v2/admin/views.py
se-bastiaan: goddamn good idea

image
image

website/events/api/v2/admin/serializers/event_registration.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+
+    def to_internal_value(self, data):
+        self.fields["member"] = serializers.PrimaryKeyRelatedField(
+            queryset=Member.objects.all()
+        )
+        return super().to_internal_value(data)
+
+    def to_representation(self, instance):
+        self.fields["member"] = MemberSerializer(detailed=False)
```
JobDoesburg, line 31: Going to `localhost:8000/api/v2/admin/events/11/registrations/7/` shows me the ability to `PUT` to a member profile, even if I do not have permissions for this. Maybe the MemberSerializer should have `readonly=True`?

website/events/api/v2/admin/views.py
```diff
@@ -0,0 +1,145 @@
...
+    queryset = Event.objects.prefetch_related("organiser")
```
JobDoesburg, line 29: Still quite a hell lot of queries that are not really necessary

image

Can probably be fixed with prefetching the event registrations as well. But we could also fix it by making the num_participants a queryable property which I'm more in favor of (#1617 which should probably be renamed 'queryable properties for active members and events'). So you can also ignore it here if you want

website/events/api/v2/admin/views.py
```diff
@@ -0,0 +1,145 @@
...
+        "start",
+        "end",
+        "published",
+        "registration_start",
+        "registration_end",
+    )
+
+
+class EventAdminDetailAPIView(
+    AdminRetrieveAPIView, AdminCreateAPIView, AdminUpdateAPIView, AdminDestroyAPIView
```
JobDoesburg, line 50: I think the DetailAPIView shouldn't be a CreateAPIView, only UpdateAPIView. The ListAPIView should be a CreateAPIView in my opinion,

website/events/api/v2/admin/serializers/event.py
```diff
@@ -0,0 +1,75 @@
...
+            "max_participants",
+            "has_fields",
+            "tpay_allowed",
+            "organiser",
+            "slide",
+        )
+
+    description = CleanedHTMLSerializer()
+    organiser = MemberGroupSerializer()
+    price = serializers.FloatField()
```
JobDoesburg, line 74: No DecimalFields, like we use in payments and sales? I think DecimalFields are better, as that's what we're using in our database as well (with `decimal_places=2, max_digits=6`)

JobDoesburg (changes requested):

JobDoesburg:
After this has been merged, I should redo the `sales` API to also use the `AdminAPIViews` and everything related

website/events/api/v2/admin/serializers/event_registration.py
JobDoesburg: Idem for the organizer, btw

website/events/api/v2/admin/serializers/event_registration.py
```diff
@@ -0,0 +1,32 @@
...
+
+    class Meta:
+        model = EventRegistration
+        fields = (
+            "pk",
+            "present",
+            "queue_position",
+            "date",
+            "date_cancelled",
+            "payment",
```
JobDoesburg, line 19: The payment should be read only as payments will be managed by the payments API

website/events/api/v2/admin/views.py
se-bastiaan: I tried adding the `eventregistration_set` to the prefetch but that added a query instead of removing one. Which is understandable since the queries you see here are the registration information field which are completely independent models. Not sure I can prefetch those through the event registrations.

Do agree that queryable properties are probably nice, but really out of scope for this PR.

website/events/api/v2/admin/serializers/event_registration.py
se-bastiaan: That's only because those views assume that the output of a serializer is exactly the same as the input of a serializer. Which I changed in this case, since it accepts a different member id and not a full object. Don't think I can do anything to prevent that. And marking the field as read only could probably make it impossible to edit the field at all.

website/events/api/v2/admin/views.py
se-bastiaan: That would overcomplicate the other view, which shows a list. This one is for single objects and editing those. And it uses a different serializer. Thus I chose to include it here instead.

website/events/api/v2/admin/serializers/event_registration.py
```diff
@@ -0,0 +1,32 @@
...
+            "member",
+            "name",
+        )
```
se-bastiaan, line 22:
```suggestion
        fields = (
            "pk",
            "present",
            "queue_position",
            "date",
            "date_cancelled",
            "payment",
            "member",
            "name",
        )
        read_only_fields = ("payment",)
```

website/events/api/v2/admin/views.py
JobDoesburg: But now, if you want to create a new event, you must do that via another event, also using the `id` of a different event.

I don't think it really makes the other view too complicated

website/events/api/v2/admin/serializers/event_registration.py
se-bastiaan: Looks like it still works when I make them read only :)

website/events/api/v2/admin/views.py
se-bastiaan: I was dumb

website/events/api/v2/admin/serializers/event.py
```diff
@@ -0,0 +1,75 @@
...
+class EventSerializer(serializers.ModelSerializer):
+    """Serializer for events."""
+
+    class Meta:
+        model = Event
+        fields = "__all__"
+
+    description = CleanedHTMLSerializer()
+    price = serializers.FloatField()
+    fine = serializers.FloatField()
```
JobDoesburg, line 25: This should also be DecimaFields

website/events/api/v2/admin/views.py
```diff
@@ -0,0 +1,149 @@
...
+    queryset = EventRegistration.objects.all()
+    permission_classes = [IsOrganiser, IsAuthenticatedOrTokenHasScope]
+    required_scopes = ["events:admin"]
+    event_lookup_field = "event_id"
+
+    def get_queryset(self):
+        return super().get_queryset().filter(event=self.kwargs["event_id"])
+
+
+class EventRegistrationAdminFieldsView(APIView):
```
JobDoesburg, line 96: This should be an AdminAPIView I think?

also maybe the `get` `put` and `patch` methods should call their `super()` methods

JobDoesburg (changes requested):

JobDoesburg:
Looks good, will approve after these 2 changes

website/events/api/v2/admin/views.py
se-bastiaan: There are no super methods. But I had to include the permission mixin indeed.

JobDoesburg (approved):

PR #1673, opened 2021-05-14
Refactor Payables to use a registry

se-bastiaan:
This rewrites the Payable working to use a registry instead of the models directly so that some logic can be outside the entity and we do not get circular dependencies.

It should not actually change any workings, and the only changes made to get the tests to work again were really to register the right models.

The implementation is inspired a lot by the way Django admin sites work, unfortunately decorators do not seem to work on models directly.

### DIFF
--- a/website/events/apps.py
+++ b/website/events/apps.py
@@ -8,3 +8,9 @@ class EventsConfig(AppConfig):
 
     name = "events"
     verbose_name = _("Events")
+
+    def ready(self):
+        # pylint: disable=import-outside-toplevel
+        from .payables import register
+
+        register()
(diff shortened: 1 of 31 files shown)


9Y0 (commented):
Looks good to me, although I would rename all `payments.py` files to `payables.py`. I guess that you chose to name them `payments.py` to distinguish them form the `payments/payables.py` file, but they actually contain `Payable`s, not `Payment`s so in my opinion `payables.py` would be a more intuitive name.

se-bastiaan:
The tests are not actually failing, Github has an outage

9Y0 (dismissed):

9Y0 (approved):

PR #1676, opened 2021-05-16
Use correct total amount calculation when saving orders

### LINKED ISSUE #1675: Sales order payments are not always saved
### Describe the bug
When paying for a Thalia pay order (via the sales payment view, so the QR code flow), the payment is not always stored back to the order. The payment is created properly, but after payment, the foreign key to the payment in the order is not saved. 

### How to reproduce
I am not sure exactly when this happens, at least it happens for the current shift 2 on the current staging environment. It might be because the shift has already been ended.

### Expected behaviour
Store the payment properly

### Additional context
Might be related to https://github.com/svthalia/concrexit/blob/6d0866022afb7fdf3edab34709d4d99e28039d59/website/sales/models/order.py#L123

JobDoesburg:
Closes #1675

### Summary
Previously, we didn't use the general queryable property when checking to save a payment. This PR fixes that, to prevent #1675 from occurring

### How to test
For example, create an order of 3x a €1,20 product ⇒ €3,60 total. Paying this via Thalia pay doesn't work on master, but it does work after this PR

### DIFF
--- a/website/sales/models/order.py
+++ b/website/sales/models/order.py
@@ -118,12 +118,7 @@ class Order(models.Model, Payable):
             return
         if self.shift.start > timezone.now():
             return
-        if (
-            self.payment
-            and float(sum(self.order_items.values_list("total", flat=True)))
-            - (self.discount or 0)
-            != self.payment.amount
-        ):
+        if self.payment and self.total_amount != self.payment.amount:
             return
         if self.payment and not self.payer:
             self.payer = self.payment.paid_by
--- a/website/sales/models/order.py
+++ b/website/sales/models/order.py
@@ -115,11 +115,13 @@ class Order(models.Model, Payable):
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
         if self.shift.locked:
-            return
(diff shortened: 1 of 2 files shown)


website/sales/models/order.py
```diff
@@ -115,16 +115,17 @@ def save(
...
+            raise ValueError("The shift hasn't started yet.")
         if (
             self.payment
-            and float(sum(self.order_items.values_list("total", flat=True)))
-            - (self.discount or 0)
-            != self.payment.amount
+            and self.subtotal - Decimal(self.discount or 0) != self.payment.amount
         ):
-            return
+            # We cannot use self.total_amount as it is a requires a database query and hence will not use any updated values
```
se-bastiaan, line 125: But isn't subtotal also using a database query?

JobDoesburg: Yes, but that query only uses other objects (namely OrderItems). The problem with `total_amount` is that it uses the `discount` field of the same object that we're trying to save here (and so it will use the old value that is still in the database whereas we want to compute the new value).

se-bastiaan (approved):

PR #1684, opened 2021-05-19
Migrate the events admin to use API v2 for ajax requests

se-bastiaan:
Part of #1428 

### Summary
Migrate the events admin to use API v2 for ajax requests

### How to test
Steps to test the changes you made:
1. Go to an event in the admin
2. Check that changing the payments and presence works

PR #1685, opened 2021-05-19
Add Payments API v2

se-bastiaan:
Part of #1428 

### Summary
Add Payments API v2. Currently in draft because I still have to write the test, but would be nice if the approach could already be reviewed.

### How to test
1. Go to the API docs
2. Try the endpoints
3. Check different authentication structures


website/events/payables.py
```diff
@@ -24,12 +25,17 @@ def payment_notes(self):
     def payment_payer(self):
         return self.model.member
 
-    def can_create_payment(self, member):
-        return is_organiser(member, self.model.event)
+    def can_manage_payment(self, member):
+        return is_organiser(member, self.model.event) and member.has_perm(
+            "events.change_eventregistration"
```
JobDoesburg, line 30: or when it has the override permission or when it is superuser. You could use the service that already exists implementing this check

website/events/payables.py
```diff
@@ -24,12 +25,17 @@ def payment_notes(self):
...
+        )
 
     @property
     def tpay_allowed(self):
-        return self.model.event.tpay_allowed
```
JobDoesburg, line 32: I think the purpose of this property is to see whether tpay payments are allowed for the whole payable. The check for whether the user actually _can_ do a tpay payment is implemented in the payments app itself

website/payments/api/v2/admin/serializers/payable_detail.py
```diff
@@ -0,0 +1,31 @@
...
+            self.payer = instance.payment_payer
+            self.topic = instance.payment_topic
+            self.notes = instance.payment_notes
+            self.payment = instance.payment
+
+            if instance.tpay_allowed:
+                self.allowed_payment_types.append(Payment.TPAY)
+
+    allowed_payment_types = ListField(child=CharField())
+    amount = DecimalField(decimal_places=2, max_digits=1000)
```
JobDoesburg, line 27: I think we're using `max_digits=6` everywhere (perhaps we should make a constant for this in the the settings)

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,130 @@
...
+from thaliawebsite.api.v2.admin import (
+    AdminListAPIView,
+    AdminCreateAPIView,
+    AdminRetrieveAPIView,
+    AdminDestroyAPIView,
+    AdminUpdateAPIView,
+)
+
+
+class PaymentListCreateView(AdminListAPIView):
```
JobDoesburg, line 29: This is only a ListView right now, it should also be a CreateView

JobDoesburg, line 29: This is missing a permission class.

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,130 @@
...
+        filters.CreatedAtFilter,
+        filters.PaymentTypeFilter,
+    )
+    ordering_fields = ("created_at",)
+
+
+class PaymentDetailView(AdminRetrieveAPIView, AdminDestroyAPIView):
+    queryset = Payment.objects.all()
+    serializer_class = PaymentSerializer
+    permission_classes = [IsAuthenticatedOrTokenHasScope]
```
JobDoesburg, line 45: This also requires DjangoModelPermissions, or the user must be allowed to manage payments for that payable

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,130 @@
...
+                detail=_("You do not have permission to perform this action.")
+            )
+
+        return payable
+
+
+class PayableDetailView(
+    PayableBaseView, AdminRetrieveAPIView, AdminUpdateAPIView, AdminDestroyAPIView
+):
+    """View that allows you to manipulate the payment for the payable. Permissions of this view are based on the payable."""
```
JobDoesburg, line 81: I don't think I fully understand this. Creating a payment happens by POSTing to a payable? I myself thought we would POST an app name, model name and pk to a general endpoint to create new payments. But this is fine as well I think

website/payments/api/v2/views.py
```diff
@@ -0,0 +1,145 @@
...
+        except (LookupError, NotRegistered) as e:
+            raise serializers.ValidationError(
+                {api_settings.NON_FIELD_ERRORS_KEY: [_("Payable model not found")]}
+            ) from e
+
+        if (
+            not payable.payment_payer
+            or not payable.tpay_allowed
+            or payable.payment_payer != PaymentUser.objects.get(pk=self.request.user.pk)
+        ):
```
JobDoesburg, line 102: I think you should be able to list the details of a non-tpay payment that you made yourself

website/payments/api/v2/views.py
```diff
@@ -0,0 +1,145 @@
...
+            services.delete_payment(
+                payable.model, request.member,
+            )
+            payable.model.save()
+        except PaymentError as e:
+            raise PermissionDenied(detail=str(e))
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def put(self, request, *args, **kwargs):
```
JobDoesburg, line 125: Payments are immutable so I don't think there should be a PUT, just a POST

website/payments/services.py
```diff
@@ -64,14 +72,19 @@ def create_payment(
     return payable.payment
 
 
-def delete_payment(model: Model):
+def delete_payment(model: Model, member: Member = None):
```
JobDoesburg, line 75: A bit unrelated, but maybe we would want to add a LogEntry here

website/sales/payables.py
```diff
@@ -19,8 +19,8 @@ def payment_notes(self):
     def payment_payer(self):
         return self.model.payer
 
-    def can_create_payment(self, member):
-        return True
+    def can_manage_payment(self, member):
+        return member.has_perm("sales.change_order")
```
JobDoesburg, line 23: And they should be organizer, or should have the override permissions

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,130 @@
...
+        filters.PaymentTypeFilter,
+    )
+    ordering_fields = ("created_at",)
+
+
+class PaymentDetailView(AdminRetrieveAPIView, AdminDestroyAPIView):
+    queryset = Payment.objects.all()
+    serializer_class = PaymentSerializer
+    permission_classes = [IsAuthenticatedOrTokenHasScope]
+    required_scopes = ["payments:admin"]
```
JobDoesburg, line 46: What exactly is this scope? It's not in the defined scopes in the settings (or is it something implicit with those AdminViews - I've never seen them)

website/events/payables.py
se-bastiaan: You mean the `is_organiser` check that it is using already?

website/payments/api/v2/admin/serializers/payable_detail.py
se-bastiaan:
```suggestion
    amount = DecimalField(decimal_places=2, max_digits=6)
```
no clue what it actually best would be, 5/6/7 🤷

website/payments/api/v2/admin/views.py
se-bastiaan: That is already part of the base class

website/payments/api/v2/admin/views.py
se-bastiaan: Have not added them to the settings yet. But the admin endpoints have a different scope than the regular endpoints.

website/payments/api/v2/admin/views.py
se-bastiaan: I needed to find a way to use the payables, and also allow for actually CRUD operations for the payments. Just sending the information to /api/v2/payments does not allow you to create a payment manually anymore.

This is a payment that will be created based on the payable, and we need a way to remove them based on the payable as well. Django rest framework does not really support DELETE requests with a body. So I opted to add the information to the URL and actually make this a REST entity. It is now a PUT (thus update) to the payable that will create and add the payment.

website/payments/api/v2/admin/views.py
se-bastiaan: nope, base class

website/payments/api/v2/views.py
se-bastiaan: And when is that exactly?

JobDoesburg: If you would want to check your payment history or something. For me it just feels like `HTTP GET thalia.nu/api/v2/payments` should return your payments.

website/events/payables.py
JobDoesburg: Wut, I'm really blind I guess

website/payments/api/v2/admin/views.py
JobDoesburg: There is no need for DELETE's with a body right? 

Anyway, I do see your idea. It's just different from what I had in mind myself

website/payments/api/v2/views.py
se-bastiaan: good point :D

website/payments/api/v2/views.py
se-bastiaan: oh you meant payments that you were the payer of, I thought you referred to all payments that a user would also be the processor of. I am not sure if this comment is on the right line then, but what you are proposing already works.

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,144 @@
...
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    def patch(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        payable = self.get_payable()
+
+        try:
+            services.create_payment(
```
JobDoesburg, line 132:
```suggestion
            payable.model.payment = services.create_payment(
```

website/payments/api/v2/serializers/payable_detail.py
```diff
@@ -0,0 +1,25 @@
...
+        if instance:
+            self.amount = instance.payment_amount
+            self.topic = instance.payment_topic
+            self.notes = instance.payment_notes
+            self.payment = instance.payment
+
+    amount = DecimalField(decimal_places=2, max_digits=1000)
+    topic = CharField()
+    notes = CharField()
+    payment = CharField()
```
JobDoesburg, line 25: why is this a char field

website/payments/api/v2/views.py
```diff
@@ -0,0 +1,142 @@
...
+from payments.api.v2 import filters
+from payments.api.v2.serializers import PaymentSerializer
+from payments.api.v2.serializers.payable_detail import PayableSerializer
+from payments.exceptions import PaymentError
+from payments.models import Payment, PaymentUser
+from thaliawebsite.api.v2.permissions import IsAuthenticatedOrTokenHasScopeForMethod
+from thaliawebsite.api.v2.serializers import EmptySerializer
+
+
+class PaymentListCreateView(ListAPIView, CreateAPIView):
```
JobDoesburg, line 26: I don't think this should be a CreateAPIView

website/payments/api/v2/views.py
```diff
@@ -0,0 +1,142 @@
...
+        payable = self.get_payable()
+
+        if payable.payment:
+            return Response(
+                data={"detail": _("This object has already been paid for.")},
+                status=status.HTTP_409_CONFLICT,
+            )
+
+        try:
+            services.create_payment(
```
JobDoesburg, line 132:
```suggestion
            payable.model.payment = services.create_payment(
```

JobDoesburg (changes requested):

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,144 @@
+import rest_framework.filters as framework_filters
```
JobDoesburg, line 1: image

Swagger doesn't like it that the admin api views and the regular api views have the same name

JobDoesburg: and it's also a bit confused by it, as it thinks a delete is allowed

image

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,144 @@
...
+        filters.CreatedAtFilter,
+        filters.PaymentTypeFilter,
+    )
+    ordering_fields = ("created_at",)
+
+    def get_serializer_class(self):
+        if self.request.method.lower() == "post":
+            return PaymentCreateSerializer
+        return PaymentSerializer
+
```
JobDoesburg, line 46:
```suggestion

    def get_queryset(self):
        queryset = super().get_queryset()
        queryset = queryset.prefetch_related("paid_by__profile", "paid_by__membership_set", "processed_by__profile", "processed_by__membership_set")
        return queryset

```

JobDoesburg: On my machine, this reduces db traffic from 140 to 28 queries

website/payments/api/v2/admin/serializers/payment.py
```diff
@@ -0,0 +1,52 @@
...
+            "paid_by",
+            "processed_by",
+            "amount",
+            "created_at",
+            "topic",
+            "notes",
+        )
+
+    paid_by = MemberSerializer(detailed=False)
+    processed_by = MemberSerializer(detailed=False)
```
JobDoesburg, line 52: these 2 should be readonly

website/payments/api/v2/admin/serializers/payable_detail.py
```diff
@@ -0,0 +1,33 @@
...
+            self.payment = instance.payment
+
+            if (
+                instance.tpay_allowed
+                and instance.payment_payer
+                and PaymentUser.objects.get(pk=instance.payment_payer.pk).tpay_allowed
+            ):
+                self.allowed_payment_types.append(Payment.TPAY)
+
+    allowed_payment_types = ListField(child=CharField())
```
JobDoesburg, line 28: when I'm testing this, I'm getting an exception:

```
Got AttributeError when attempting to get a value for field `allowed_payment_types` on serializer `PayableSerializer`.
The serializer field might be named incorrectly and not match any attribute or key on the `OrderPayable` instance.
Original exception text was: 'OrderPayable' object has no attribute 'allowed_payment_types'.
```

JobDoesburg (changes requested):

website/payments/api/v2/admin/views.py
se-bastiaan: That is not required, the code in services already does that.

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,144 @@
...
+from thaliawebsite.api.v2.admin import (
+    AdminListAPIView,
+    AdminCreateAPIView,
+    AdminRetrieveAPIView,
+    AdminDestroyAPIView,
+    AdminUpdateAPIView,
+)
+
+
+class PaymentListCreateView(AdminListAPIView, AdminCreateAPIView):
```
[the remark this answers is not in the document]
se-bastiaan: Does that really need to be explicit? It's an admin API and I have tried to make it behave the same as the actual admin.

website/payments/api/v2/views.py
se-bastiaan: Inside the function already

website/payments/api/v2/admin/serializers/payable_detail.py
```diff
@@ -0,0 +1,33 @@
...
+    amount = DecimalField(decimal_places=2, max_digits=6)
+    payer = MemberSerializer(detailed=False)
+    topic = CharField()
+    notes = CharField()
+    payment = CharField()
```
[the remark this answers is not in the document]
JobDoesburg: This is still a CharField here (you did change the other one)

website/payments/api/v2/admin/views.py
```diff
@@ -0,0 +1,154 @@
...
+class PaymentDetailView(AdminRetrieveAPIView, AdminDestroyAPIView):
+    """View that allows you to manage a single payment as admin."""
+
+    queryset = Payment.objects.all()
+    serializer_class = PaymentSerializer
+    permission_classes = [IsAuthenticatedOrTokenHasScope]
+    required_scopes = ["payments:admin"]
+
+
+class PayableDetailView(AdminPermissionsMixin, GenericAPIView):
```
JobDoesburg, line 76: I'm getting error
`'PayableDetailView' should either include a `queryset` attribute, or override the `get_queryset()` method.`

JobDoesburg (changes requested):

website/payments/api/v2/admin/views.py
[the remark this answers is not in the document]
se-bastiaan: I have added some small documentation

website/payments/api/v2/admin/serializers/payable_create.py
```diff
@@ -0,0 +1,12 @@
...
+from rest_framework.serializers import Serializer
+
+from payments.models import Payment
+
+
+class PayableCreateAdminSerializer(Serializer):
+    """Serializer to create a payment from a payable."""
+
+    payment_type = ChoiceField(
+        choices=[Payment.TPAY, Payment.CASH, Payment.CARD, Payment.WIRE]
```
JobDoesburg, line 11: This does allow admins to create Thalia Pay payments for payables (for other users). Do we want to allow that? Right now, we don't...

JobDoesburg: This is my last review comment, I promise

se-bastiaan: Yes we do want this, because this is what is necessary for all the payables to be paid from the app as admin. It replaces the complete 'set payment to NONE/CARD/WIRE etc and then it magically does this work for you. Example of usage is #1684.

And that is also why I added the `can_manage` part to the Payable, it determines if someone can create payments for said Payable, to lock it down.

JobDoesburg: I explicitly mean Thalia Pay payments. Right now, we (deliberately?) do not allow regular admins (like EVC members) to create TPay payments on behalf of others: only users themselves should be able to do that. I do see that Payment.TPAY is an actual payment type, but I don't think we should allow writing it.

se-bastiaan: Ah I think I misunderstood what you meant. You're right, this option should not exist.

JobDoesburg (approved):

PR #1693, opened 2021-05-24
Filter and search for groups in api v2

### LINKED ISSUE #1640: Search and filters for groups in API v2
Search and filtering options for `api/v2/activemembers/groups/`.
I suggest a search parameter for `name`, and filter on `type` and perhaps `since` and `until` as is done for events, or at least some way to only get currently existing results.

### Motivation
This way users will be able to search for a group in the app, and for example show only the ones that currently exist, i.e. those that they might be able to join.

Not doing it, but that would limit functionality of the app.

9Y0:
Closes #1640 

### Summary
- Added a search option on group name
- Added a filter on group type (i.e. board, committee, society)
- Added `start` and `end` parameters to filter on when groups existed
- Removed the `/api/v2/activemembers/committees/`, `/api/v2/activemembers/societies/`, `/api/v2/activemembers/boards/` endpoints since they're replaced by the `type` filter

### How to test
1. Go to `/api/v2/activemembers/groups/`
2. Check out the search and filter options

### DIFF
--- a/website/activemembers/api/v2/filters.py
+++ b/website/activemembers/api/v2/filters.py
@@ -39,7 +39,7 @@ class MemberGroupDateFilter(filters.BaseFilterBackend):
         start, end = extract_date_range(request, allow_empty=True)
 
         if start is not None:
-            queryset = queryset.filter(Q(until__isnull=True) | Q(until_gte=start))
+            queryset = queryset.filter(Q(until__isnull=True) | Q(until__gte=start))
         if end is not None:
             queryset = queryset.filter(Q(since__isnull=True) | Q(since__lte=end))
 
--- a/website/activemembers/api/v2/filters.py
+++ b/website/activemembers/api/v2/filters.py
@@ -1,0 +1,65 @@
+from django.db.models import Q
+from rest_framework import filters
+
+from activemembers.models import Board, Committee, Society
+from utils.snippets import extract_date_range
+
+
+class MemberGroupTypeFilter(filters.BaseFilterBackend):
+    """Allows you to filter by organiser id."""
+
+    def filter_queryset(self, request, queryset, view):
+        type = request.query_params.get("type", None)
+
+        if type:
+            if type == "board":
+                queryset = Board.objects.all()
+            elif type == "committee":
+                queryset = Committee.objects.all()
+            elif type == "society":
+                queryset = Society.objects.all()
+
+        return queryset
+
+    def get_schema_operation_parameters(self, view):
+        return [
+            {
+                "name": "type",
+                "required": False,
+                "in": "query",
+                "description": "Filter by member group type",
+                "schema": {"type": "string", "enum": ["board", "committee", "society"]},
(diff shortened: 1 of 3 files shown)


website/activemembers/api/v2/filters.py
```diff
@@ -0,0 +1,65 @@
...
+from utils.snippets import extract_date_range
+
+
+class MemberGroupTypeFilter(filters.BaseFilterBackend):
+    """Allows you to filter by organiser id."""
+
+    def filter_queryset(self, request, queryset, view):
+        type = request.query_params.get("type", None)
+
+        if type:
```
se-bastiaan, line 14: This if seems superfluous

website/activemembers/api/v2/filters.py
```diff
@@ -0,0 +1,65 @@
...
+            if type == "board":
+                queryset = Board.objects.all()
+            elif type == "committee":
+                queryset = Committee.objects.all()
+            elif type == "society":
+                queryset = Society.objects.all()
```
se-bastiaan, line 20: These querysets should probably use the active_objects prop

DeD1rk:
deploy review

website/activemembers/api/v2/filters.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class MemberGroupTypeFilter(filters.BaseFilterBackend):
+    """Allows you to filter by organiser id."""
+
+    def filter_queryset(self, request, queryset, view):
+        type = request.query_params.get("type", None)
+
+        if type == "board":
+            queryset = Board.active_objects.all()
```
DeD1rk, line 15: These create a new queryset, instead of filtering on the given one.

website/activemembers/api/v2/views.py
```diff
@@ -26,6 +23,12 @@ class MemberGroupListView(ListAPIView):
 
     serializer_class = MemberGroupListSerializer
     queryset = MemberGroup.active_objects.all()
```
DeD1rk, line 25:
```suggestion
    queryset = MemberGroup.objects.all()
```
This way the you can see old committees as well. If you want to filter on 'active' ones, there's the date filter.

9Y0: @se-bastiaan You requested to change it to `active_objects` so could you maybe comment on this?

se-bastiaan: I am not sure we should be exposing non-active committees outside of the admin, it's really more a published/non-published situation where people who are not admins should not have access to those objects

se-bastiaan: Boards are groups that are [always active](https://github.com/svthalia/concrexit/blob/master/website/activemembers/models.py#L164) since they contain extra constraints like that there can only be one board within a certain time period.

So yes, for other types of groups this really means that they should not be shown anymore.

DeD1rk (approved):

PR #1697, opened 2021-05-24
Fix orphaned messages

### LINKED ISSUE #1674: Unpublishing or removing events orphans scheduled messages
### Describe the bug
Unpublishing or removing events orphans scheduled messages.

### How to reproduce
1. Go to events
2. Create a new one, published
3. Check scheduled messages
4. Unpublish the event
5. Check the normal messages
6. See that the message that existed before has moved from scheduled to non-scheduled

### Expected behaviour
The message should actually be deleted.

Jorritboer:
Closes #1674 

### Summary
Fixed orphaned messages after unpublishing or deleting events. 

### How to test
1. Make a new published event
2. Set it to unpublished
3. Check that there are no messages in Messages or Scheduled Messages
4. Set it back to published
5. Delete the event
6. Check that there are no messages in Messages or Scheduled Messages

### DIFF
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -438,7 +438,7 @@ class Event(models.Model, metaclass=ModelTranslateMeta):
                     self.registration_reminder = registration_reminder
                     self.registration_reminder.users.set(Member.current_members.all())
                 elif registration_reminder.pk is not None:
-                    delete_collector.add([self.registration_reminder])
+                    delete_collector.collect([self.registration_reminder])
                     self.registration_reminder = None
 
             start_reminder_time = self.start - timezone.timedelta(hours=1)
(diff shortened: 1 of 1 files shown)


website/events/models/event.py
```diff
@@ -486,9 +486,9 @@ def delete(self, using=None, keep_parents=False):
             self.registration_reminder is not None
             and not self.registration_reminder.sent
         ):
-            collector.add([self.registration_reminder])
+            collector.collect([self.registration_reminder], keep_parents=keep_parents)
         if self.start_reminder is not None and not self.start_reminder.sent:
-            collector.add([self.start_reminder])
+            collector.collect([self.start_reminder], keep_parents=keep_parents)
         if self.has_food_event:
             collector.add([self.food_event])
```
se-bastiaan, line 493: Should this then also be collect?

se-bastiaan:
What is the reason that such an add does not work anymore? Did something change in python?

website/events/models/event.py
Jorritboer: I also checked that. It's not necessary, it works like this. But I can of course change it for consistency

Jorritboer:
I don't understand exactly but it has to do with the fact that ScheduledMessage is a subclass of Message and with add you only delete the subclass, but with collect you delete the class and all super classes.

JobDoesburg (approved):

PR #1706, opened 2021-05-31
Use queryable properties for PaymentUser

### LINKED ISSUE #1397: Speed up the PaymentUserAdmin

JobDoesburg:
Closes #1397 

### Summary
Now, the payments app uses queryable properties for a lot of computations, like the Thalia Pay balance or whether a user has Thalia Pay enabled. This gives a massive speedup as we outsource the computation the database, that does it way more optimal than python itself

### How to test
1. Check out the payment user admin
2. Try to make thalia pay payments


JobDoesburg:
FINALLY

pingiun (approved):
This seems very good. The implementation is simple, modular and fast. It seems you avoided a mistake I made because you didn't even need to change the tests that much.

I don't see any problems with this, it solves the issue in a great way. Nice work!

PR #1708, opened 2021-06-01
Use distinct queryset in PaymentUserAdmin

JobDoesburg:
### Summary
On certain sql systems, querying for whether a bank account exists (in the `tpay_enabled` `queryable-property`) could duplicate the entries, because in the backend a Cartesian product of the two tables was computed. This PR introduces a different way of counting bank accounts that are valid that does not have this problem

### How to test
1. Have a user
2. Make it have 2 bank accounts used for Thalia Pay (one active, one inactive)
3. Check out the PaymentUserAdmin
4. It should contain that user just once (and ignore the old inactive bank account)

### DIFF
--- a/website/payments/models.py
+++ b/website/payments/models.py
@@ -7,8 +7,8 @@ from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import DEFERRED, Q, Sum, BooleanField, DecimalField, Count
-from django.db.models.expressions import Case, When, Value
+from django.db.models import DEFERRED, Q, Sum, BooleanField, DecimalField
+from django.db.models.expressions import Case, When, Value, Exists, OuterRef
 from django.db.models.functions import Coalesce
 from django.urls import reverse
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


website/payments/models.py
```diff
@@ -32,19 +32,17 @@ class Meta:
...
+            "bank_accounts__pk",
+            filter=Q(
+                bank_accounts__valid_from__isnull=False,
+                bank_accounts__valid_from__lte=today,
+            )
+            & (
+                Q(bank_accounts__valid_until__isnull=True)
+                | Q(bank_accounts__valid_until__gt=today)
+            )
+            & Value(settings.THALIA_PAY_ENABLED_PAYMENT_METHOD),
```
pingiun, line 45:
```suggestion
return Case(
            When(
                Exists(
                    BankAccount.objects.filter(owner=OuterRef("pk"))
                    .filter(valid_from__isnull=False, valid_from__lte=today)
                    .filter(valid_until__isnull=True, valid_until__gt=today)
                ),
                then=settings.THALIA_PAY_ENABLED_PAYMENT_METHOD,
            ),
            default=False,
```

JobDoesburg: almost, there's missing a disjunction in your filter criteria

pingiun: ah good catch

pingiun (approved):

PR #1714, opened 2021-06-07
Make membership ManyToOne instead of OneToOne

### LINKED ISSUE #962: Renewals not processed correctly on uniqueness constraint for membership id
In GitLab by @JobDoesburg on Nov 20, 2019, 20:08

### One-sentence description

Renewals not processed correctly on uniqueness constraint for membership id

1. Do a renewal for 1 year
2. Approve it, add a payment, finish it fully etc
3. Do a renewal for the rest of your study
4. Approve it
5. Process a payment
6. Breakage

(the payment gets processed, the membership is correctly added to the user, but the renewal does stick at 'approved' (with a processed payment attached))

### Expected behaviour

Do not crash

JobDoesburg:
Closes #962

### Summary
If a user renews their membership for the second time in the same year (because you do a study-length renewal after a regular renewal), the site would crash after payment, as there would be 2 renewals pointing to the same membership object. This PR fixes that and I think that this is by far the easiest and cleanest solution to the problem.

### How to test

1. Do a renewal for 1 year
2. Approve it, add a payment, finish it fully etc
3. Do a renewal for the rest of your study
4. Approve it
5. Process a payment
6. No breakage


pingiun (approved):
LGTM

PR #1715, opened 2021-06-07
Fix username uniqueness

### LINKED ISSUE #945: Member can be accepted and have paid, but status is not completed
In GitLab by _thaliatechnicie on Oct 30, 2019, 19:21

This ValueError happens sometimes while processing payments for new members or renewals. I'm not quite sure how to reproduce it, but sometimes when setting a registration or renewal to paid. The status of the renewal or registration will afterwards be "Accepted" instead of "Completed", and a 500 error is shown.

Sentry Issue: [CONCREXIT-H](https://sentry.io/organizations/thalia/issues/1172358095/?referrer=gitlab_integration)

```
ValueError: The given username must be set
(19 additional frame(s) were not displayed)
...
  File "registrations/signals.py", line 12, in post_payment_save
    services.process_payment(instance)
  File "registrations/services.py", line 414, in process_payment
    member = _create_member_from_registration(registration)
  File "registrations/services.py", line 286, in _create_member_from_registration
    last_name=registration.last_name
  File "django/contrib/auth/models.py", line 151, in create_user
    return self._create_user(username, email, password, **extra_fields)
  File "django/contrib/auth/models.py", line 140, in _create_user
    raise ValueError('The given username must be set')
```

JobDoesburg:
Closes #945 

### Summary
This PR does 2 things related to the problem in #945:

1. It prevents the bug from occurring in some cases, by making sure that it is not possible to save registrations that share the username with other registrations (even when they are not completed yet)
2. When the bug still occurs, the payment is deleted so that we will never end up in the situation that a registration is paid but not accepted.
3. In all these cases, the payment button admin view will show an error to the user

### How to test
1. Create a registration
2. Approve it, etc.
3. Set the username to a different user's username 
4. That should fail
5. Set the username to the username of a different accepted registration
6. That should fail as well
7. Secretly force override the username value by manually inserting it in the database. Choose a username of a user that already exists
8. Paying will fail and result in an error shown to the user. The payment will be deleted / not be created and the entry will be 'ready for review' again.

### DIFF
--- a/website/payments/admin_views.py
+++ b/website/payments/admin_views.py
@@ -16,6 +16,7 @@ from django.utils.decorators import method_decorator
 from django.utils.http import url_has_allowed_host_and_scheme
 from django.utils.translation import gettext_lazy as _
 from django.views import View
+from sentry_sdk import capture_exception
 
 from payments import services
 from .models import Payment, Batch, PaymentUser
@@ -48,6 +49,7 @@ class PaymentAdminView(View):
             payable_obj.save()
         # pylint: disable=broad-except
         except Exception as e:
+            capture_exception(e)
             messages.error(
                 request,
                 _("Something went wrong paying %s: %s")
--- a/website/payments/admin_views.py
+++ b/website/payments/admin_views.py
@@ -45,6 +45,7 @@ class PaymentAdminView(View):
                 payable_obj, self.request.member, request.POST["type"],
             )
             payable_obj.save()
+        # pylint: disable=broad-except
         except Exception as e:
             messages.error(
                 request,
--- a/website/payments/admin_views.py
+++ b/website/payments/admin_views.py
@@ -44,6 +44,7 @@ class PaymentAdminView(View):
             result = services.create_payment(
                 payable_obj, self.request.member, request.POST["type"],
             )
+            payable_obj.payment = result
             payable_obj.save()
         # pylint: disable=broad-except
         except Exception as e:
--- a/website/payments/admin_views.py
+++ b/website/payments/admin_views.py
@@ -40,10 +40,18 @@ class PaymentAdminView(View):
         payable_model = apps.get_model(app_label=app_label, model_name=model_name)
         payable_obj = payable_model.objects.get(pk=payable)
 
-        result = services.create_payment(
-            payable_obj, self.request.member, request.POST["type"],
-        )
-        payable_obj.save()
+        try:
+            result = services.create_payment(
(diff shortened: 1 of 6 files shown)


JobDoesburg:
Something seems to be wrong... Under certain circumstances the bug is still there

JobDoesburg:
I still don't fully understand this, but there seemed to be some nondeterminism in how python behaved in the payment widget's admin view.  https://github.com/svthalia/concrexit/pull/1715/commits/13f8e3934c1331ab74407f96b62e7fdae7702a5f fixed it

website/payments/admin_views.py
```diff
@@ -40,10 +40,20 @@ def post(self, request, *args, app_label, model_name, payable, **kwargs):
...
-        payable_obj.save()
+        try:
+            result = services.create_payment(
+                payable_obj, self.request.member, request.POST["type"],
+            )
+            payable_obj.payment = result
+            payable_obj.save()
+        # pylint: disable=broad-except
+        except Exception as e:
+            messages.error(
```
se-bastiaan, line 51: Does this propagate the error to the sentry exception logging?

JobDoesburg: Oeh, good catch, it doesn't... What do you think, is it better to just crash the site and send it to sentry? Or do we want to show an error gracefully to the user (because in case of non-unique username, I think we want the latter)

se-bastiaan (changes requested):
I don't understand how this fixes the issue? It only prevents you from getting into a problematic state by reverting a faulty entry.

website/payments/admin_views.py
se-bastiaan: You can just manually add a call to Sentry https://docs.sentry.io/platforms/python/usage/

se-bastiaan:
Not that these changes are wrong btw, but I don't believe it actually solves the issue itself. But since I don't know what that issue is this may be the best way to find out. Just trying to say that I am not sure this really really closes the issue. Even though I am happy to do so.

JobDoesburg:
> I don't understand how this fixes the issue? It only prevents you from getting into a problematic state by reverting a faulty entry.

- By checking the username against other registrations, it basically becomes impossible to have the problem occur, as it will never be possible to set the username field to something that could eventually violate uniqueness
- In case the problem still exists (for example, when adding users manually), it reverts the entry, so that it is clear to the person processing the payment that something went wrong, and prevents you from ending up in a deadlocked situation. Anyway it cannot happen that a registration exists and is paid, but not completed, nor that the user processing does not know this happened.

PR #1728, opened 2021-06-14
Add tpay status API endpoint

### LINKED ISSUE #1530: Thalia Pay: Allow API access to Thalia Pay status of members
It is currently not possible to find out if a user is allowed to use Thalia Pay.

There needs to be a field in the payable object. So events and pizzas for example.

### Motivation
To be able to implement Thalia Pay in the app we really need this.

JobDoesburg:
Closes #1530 

### Summary

### How to test
1. Go to `<hostname>/api/v2/payments/users/me` and inspect the data

### DIFF
--- a/website/payments/api/v2/urls.py
+++ b/website/payments/api/v2/urls.py
@@ -13,7 +13,7 @@ app_name = "payments"
 urlpatterns = [
     path("payments/", PaymentListView.as_view(), name="payments-list"),
     path(
-        "payments/tpay-status/",
+        "payments/users/me/",
         PaymentUserCurrentView.as_view(),
         name="payments-tpay-status",
     ),
--- a/website/payments/api/v2/urls.py
+++ b/website/payments/api/v2/urls.py
@@ -5,12 +5,18 @@ from payments.api.v2.views import (
     PaymentListView,
     PaymentDetailView,
     PayableDetailView,
+    PaymentUserCurrentView,
 )
 
 app_name = "payments"
 
 urlpatterns = [
     path("payments/", PaymentListView.as_view(), name="payments-list"),
+    path(
+        "payments/tpay-status/",
+        PaymentUserCurrentView.as_view(),
+        name="payments-tpay-status",
+    ),
     path("payments/<uuid:pk>/", PaymentDetailView.as_view(), name="payment-detail"),
     path(
         "payments/payables/<str:app_label>/<str:model_name>/<str:payable_pk>/",
--- a/website/payments/api/v2/serializers/payment_user.py
+++ b/website/payments/api/v2/serializers/payment_user.py
@@ -1,0 +1,10 @@
+from rest_framework.fields import DecimalField, BooleanField
+from rest_framework.serializers import Serializer
+
+
+class PaymentUserSerializer(Serializer):
+    """Serializer to show payment users."""
+
+    tpay_balance = DecimalField(decimal_places=2, max_digits=1000)
+    tpay_allowed = BooleanField()
+    tpay_enabled = BooleanField()
--- a/website/payments/api/v2/views.py
+++ b/website/payments/api/v2/views.py
@@ -1,12 +1,10 @@
 import rest_framework.filters as framework_filters
 from django.apps import apps
-from django.http import Http404
 from django.utils.translation import gettext_lazy as _
 from rest_framework import status, serializers
 from rest_framework.exceptions import PermissionDenied, ValidationError
 from rest_framework.generics import (
     ListAPIView,
-    CreateAPIView,
     RetrieveAPIView,
     get_object_or_404,
 )
@@ -17,6 +15,7 @@ from payments import services, payables, NotRegistered
 from payments.api.v2 import filters
 from payments.api.v2.serializers import PaymentSerializer
 from payments.api.v2.serializers.payable_detail import PayableSerializer
+from payments.api.v2.serializers.payment_user import PaymentUserSerializer
 from payments.exceptions import PaymentError
 from payments.models import Payment, PaymentUser
 from thaliawebsite.api.v2.permissions import IsAuthenticatedOrTokenHasScopeForMethod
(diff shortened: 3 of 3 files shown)


website/payments/api/v2/urls.py
```diff
@@ -5,12 +5,18 @@
...
     PayableDetailView,
+    PaymentUserCurrentView,
 )
 
 app_name = "payments"
 
 urlpatterns = [
     path("payments/", PaymentListView.as_view(), name="payments-list"),
+    path(
+        "payments/tpay-status/",
```
se-bastiaan, line 16: I am not really sure this is the best way to do this. It's not an object, it's more an aggregation action and I would like those to be namespaced like `payments/actions/status`. I just have not had found a situation where this was really necessary so far. But maybe we can think of something better, since it could also  be `/payments/users/me`

se-bastiaan:
Is the allowed status not something that should be known per payable?

se-bastiaan (changes requested):
Is the allowed status not something that should be known per payable?

JobDoesburg:
> Is the allowed status not something that should be known per payable?

No in this case, `tpay_allowed` is whether a user is allowed to use tpay (default, True) or that they are blacklisted. Additionally each payable also has a `tpay_allowed_payment_method` which indicates if tpay can be used for that specific payable

se-bastiaan:
> 
> 
> > Is the allowed status not something that should be known per payable?
> 
> No in this case, `tpay_allowed` is whether a user is allowed to use tpay (default, True) or that they are blacklisted. Additionally each payable also has a `tpay_allowed_payment_method` which indicates if tpay can be used for that specific payable

I am not sure your answer is actually 'no'. Because what you say doesn't work for the API without actually implementing it.

website/payments/api/v2/urls.py
JobDoesburg: Shall we do `/payments/users/me` then?

JobDoesburg:
> > > Is the allowed status not something that should be known per payable?
> > 
> > 
> > No in this case, `tpay_allowed` is whether a user is allowed to use tpay (default, True) or that they are blacklisted. Additionally each payable also has a `tpay_allowed_payment_method` which indicates if tpay can be used for that specific payable
> 
> I am not sure your answer is actually 'no'. Because what you say doesn't work for the API without actually implementing it.


https://github.com/svthalia/concrexit/blob/040fb4398f010c326ba1628d7733abd1b85900e7/website/payments/payables.py#L45 is the property that is used by `create_payment`. The payment API doesn't expose that field, but creating the payment will just fail: https://github.com/svthalia/concrexit/blob/040fb4398f010c326ba1628d7733abd1b85900e7/website/payments/api/v2/views.py#L119

se-bastiaan:
> > > > Is the allowed status not something that should be known per payable?
> > > 
> > > 
> > > No in this case, `tpay_allowed` is whether a user is allowed to use tpay (default, True) or that they are blacklisted. Additionally each payable also has a `tpay_allowed_payment_method` which indicates if tpay can be used for that specific payable
> > 
> > 
> > I am not sure your answer is actually 'no'. Because what you say doesn't work for the API without actually implementing it.
> 
> https://github.com/svthalia/concrexit/blob/040fb4398f010c326ba1628d7733abd1b85900e7/website/payments/payables.py#L45
> is the property that is used by `create_payment`. The payment API doesn't expose that field, but creating the payment will just fail:
> 
> https://github.com/svthalia/concrexit/blob/040fb4398f010c326ba1628d7733abd1b85900e7/website/payments/api/v2/views.py#L119

Correct, but that property is also used to determine if the button should be shown.

https://github.com/svthalia/concrexit/blob/master/website/payments/templates/payments/templatetags/payment_button.html#L3

Without providing the attribute that behaviour cannot be replicated in the app.

JobDoesburg:
> > > > > Is the allowed status not something that should be known per payable?
> > > > 
> > > > 
> > > > No in this case, `tpay_allowed` is whether a user is allowed to use tpay (default, True) or that they are blacklisted. Additionally each payable also has a `tpay_allowed_payment_method` which indicates if tpay can be used for that specific payable
> > > 
> > > 
> > > I am not sure your answer is actually 'no'. Because what you say doesn't work for the API without actually implementing it.
> > 
> > 
> > [...]
> > is the property that is used by `create_payment`. The payment API doesn't expose that field, but creating the payment will just fail:
> > [...]
> 
> Correct, but that property is also used to determine if the button should be shown.
> 
> https://github.com/svthalia/concrexit/blob/040fb4398f010c326ba1628d7733abd1b85900e7/website/payments/templates/payments/templatetags/payment_button.html#L3
> 
> Without providing the attribute that behaviour cannot be replicated in the app.

Alright, so the payments API should also provide that attribute yeah I see. But that's a different one than I'm adding in this PR.

JobDoesburg:
> > > > > > Is the allowed status not something that should be known per payable?
> > > > > 
> > > > > 
> > > > > No in this case, `tpay_allowed` is whether a user is allowed to use tpay (default, True) or that they are blacklisted. Additionally each payable also has a `tpay_allowed_payment_method` which indicates if tpay can be used for that specific payable
> > > > 
> > > > 
> > > > I am not sure your answer is actually 'no'. Because what you say doesn't work for the API without actually implementing it.
> > > 
> > > 
> > > [...]
> > > is the property that is used by `create_payment`. The payment API doesn't expose that field, but creating the payment will just fail:
> > > [...]
> > 
> > 
> > Correct, but that property is also used to determine if the button should be shown.
> > https://github.com/svthalia/concrexit/blob/040fb4398f010c326ba1628d7733abd1b85900e7/website/payments/templates/payments/templatetags/payment_button.html#L3
> > 
> > Without providing the attribute that behaviour cannot be replicated in the app.
> 
> Alright, so the payments API should also provide that attribute yeah I see. But that's a different one than I'm adding in this PR.

See #1737

se-bastiaan (approved):

PR #1739, opened 2021-06-20
Fix missing fields in definition for renewal form

### LINKED ISSUE #1738: Renewal of membership send email about references without reason
### Describe the bug
A renewal of a membership to the 'until study end' triggered the email for references.

### How to reproduce
1. Create a registration for a completely new user
2. Create an upgrade for the membership
3. Get the email

### Expected behaviour

No email should be received for references.

### Screenshots
Screenshot 2021-06-20 at 15 20 10
Screenshot 2021-06-20 at 15 20 20

se-bastiaan:
Closes #1738

### Summary
Fix missing fields in definition for renewal form.

### How to test
1. Create a registration for a completely new user
2. Create an upgrade for the membership
3. Get the email

### DIFF
--- a/website/registrations/forms.py
+++ b/website/registrations/forms.py
@@ -151,7 +151,14 @@ class RenewalForm(forms.ModelForm):
 
     class Meta:
         model = Renewal
-        fields = ("member", "length", "contribution", "membership_type")
+        fields = (
+            "member",
+            "length",
+            "contribution",
+            "membership_type",
+            "no_references",
+            "remarks",
+        )
 
 
 class ReferenceForm(forms.ModelForm):


JobDoesburg:
Could you perhaps explain why this caused/solves the problem?

se-bastiaan:
> Could you perhaps explain why this caused/solves the problem?

The fields are not expected by the form, so the value that is set in the view is never saved. And thus `no_references` is `False`. And thus the email is dispatched.

JobDoesburg (approved):

JobDoesburg:
> > Could you perhaps explain why this caused/solves the problem?
> 
> The fields are not expected by the form, so the value that is set in the view is never saved. And thus `no_references` is `False`. And thus the email is dispatched.

Ahh, I see

PR #1745, opened 2021-06-20
Reintroduce membership upgrade

### LINKED ISSUE #1744: Membership 'upgrade' to 'until graduation' is always 30 euros
### Describe the bug
Membership 'upgrade' to 'until graduation' is always 30 euros.

### How to reproduce
1. Make sure you have a membership for the current year
2. Open the renewal page: http://localhost:8000/user/membership/
3. Select the member renewal with 'until graduation'
4. Submit
5. Open the admin and note that the contribution is 30 euros, it should give a discount since you already paid for this year

### Expected behaviour
The amount should include the discount and be 22,50.

### Screenshots
Screenshot 2021-06-20 at 19 05 57
Screenshot 2021-06-20 at 19 06 03

JobDoesburg:
Closes #1744 

### Summary
Current members, when upgrading their membership, will now not have to pay €30 but €22,5

### How to test
1. Have a member with a 1 year membership
2. Let them make a renewal
3. Profit

### DIFF
--- a/website/registrations/models.py
+++ b/website/registrations/models.py
@@ -4,7 +4,7 @@ import uuid
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core import validators
-from django.core.exceptions import ValidationError
+from django.core.exceptions import ValidationError, ObjectDoesNotExist
 from django.core.validators import MinValueValidator, RegexValidator
 from django.db import models
 from django.template.defaultfilters import floatformat
@@ -102,6 +102,19 @@ class Entry(models.Model):
         "members.Membership", on_delete=models.SET_NULL, blank=True, null=True,
     )
 
+    @property
+    def membership_upgrade_discount_applies(self):
+        try:
+            return (
+                self.renewal
+                and self.length == Entry.MEMBERSHIP_STUDY
+                and self.renewal.member.current_membership is not None
+                and self.renewal.member.current_membership.until is not None
+                and self.renewal.member.current_membership.type == Membership.MEMBER
+            )
+        except ObjectDoesNotExist:
+            return False
+
     def save(
         self, force_insert=False, force_update=False, using=None, update_fields=None
     ):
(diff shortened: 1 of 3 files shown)


website/registrations/models.py
```diff
@@ -111,7 +111,8 @@ def save(
         if self.membership_type == Membership.BENEFACTOR:
             self.length = self.MEMBERSHIP_YEAR
         else:
-            self.contribution = settings.MEMBERSHIP_PRICES[self.length]
+            if not self.contribution:
```
pingiun, line 114: This seems to cause the failing test, is it a necessary change?

website/registrations/models.py
```diff
@@ -407,6 +408,19 @@ def save(
     ):
         if self.pk is None:
             self.status = Entry.STATUS_REVIEW
+
+        if (
+            self.contribution is None
+            and self.length == Entry.MEMBERSHIP_STUDY
+            and self.member.current_membership is not None
+            and self.member.current_membership.length - Entry.MEMBERSHIP_YEAR
```
pingiun, line 416: `self.member.current_membership.length - Entry.MEMBERSHIP_YEAR`

is not a boolean check? if this means `!= 0`, please write this explicitely

se-bastiaan: This was the previous check that we had: https://github.com/svthalia/concrexit/blob/ca6abb32946c961fa7d52bc04b1cd34bfb73e88e/website/registrations/services.py#L254


I believe `Entry.MEMBERSHIP_YEAR` is a string.

website/registrations/models.py
```diff
@@ -111,7 +128,14 @@ def save(
         if self.membership_type == Membership.BENEFACTOR:
             self.length = self.MEMBERSHIP_YEAR
         else:
-            self.contribution = settings.MEMBERSHIP_PRICES[self.length]
+            if self.membership_upgrade_discount_applies:
+                # The membership upgrade discount applies
```
se-bastiaan, line 132:
```suggestion
```

Completely redundant

website/registrations/models.py
```diff
@@ -111,7 +128,13 @@ def save(
         if self.membership_type == Membership.BENEFACTOR:
             self.length = self.MEMBERSHIP_YEAR
         else:
-            self.contribution = settings.MEMBERSHIP_PRICES[self.length]
+            if self.membership_upgrade_discount_applies:
+                self.contribution = (
+                    settings.MEMBERSHIP_PRICES["study"]
```
se-bastiaan, line 133: I think that these keys for the lengths are constants somewhere

JobDoesburg: I checked and we use the strings everywhere in the code base (as these are settings, not values in the database)

JobDoesburg: Will address in issue

pingiun (approved):
LGTM

PR #1756, opened 2021-06-21
Do not allow deleting payments

### LINKED ISSUE #1752: PaymentDetailView in admin API allows deleting payments unauthorized
https://github.com/svthalia/concrexit/blob/4ab37961f50e398cc52422cdc1df66f6ab8ff2ee/website/payments/api/v2/admin/views.py#L69

### Describe the bug
Payments sometimes should be undeletable. For example, TPay payments that are in a batch. The PaymentAdmin prevents such deletions. However, the rest framework DestroyAPIView does not respect that.

### How to reproduce
1. Have a payment 
2. Add it to a batch
3. Process the batch
4. Do the API `DELETE` request at `/api/v2/admin/payments/<id>`

### Expected behaviour
Either disable payment deletion at all from the API, or manually implement a check that the payment is not in a processed batch.

JobDoesburg:
Closes #1752 

### Summary
Prevents deleting TPay processed payments

### How to test
1. Have a TPAY payment
2. Put it in the batch
3. Process the batch
4. Try to remove it from the API. It should throw an error

### DIFF
--- a/website/payments/api/v2/admin/views.py
+++ b/website/payments/api/v2/admin/views.py
@@ -74,6 +74,11 @@ class PaymentDetailView(AdminRetrieveAPIView, AdminDestroyAPIView):
     permission_classes = [IsAuthenticatedOrTokenHasScope]
     required_scopes = ["payments:admin"]
 
+    def delete(self, request, *args, **kwargs):
+        if self.get_object().batch and self.get_object().batch.processed:
+            raise PermissionDenied("This payment cannot be deleted.")
+        return super().delete(request, *args, **kwargs)
+
 
 class PayableDetailView(APIView):
     """View that allows you to manipulate the payment for the payable.


pingiun:
Weird error, I selected the "delete selected payments" option

image

pingiun:
Ah sorry I should check the API

JobDoesburg:
> Weird error, I selected the "delete selected payments" option
> 
> image

That is the result of this code in website/payments/admin.py:136 :

```python
        if (
            "payment/" in request.path
            and request.POST
            and request.POST.get("action") == "delete_selected"
        ):
            for payment_id in request.POST.getlist("_selected_action"):
                payment = Payment.objects.get(id=payment_id)
                if payment.batch and payment.batch.processed:
                    return False
``` 

It is there because that admin action with the delete filter also has the problem

pingiun (approved):
LGTM

PR #1757, opened 2021-06-21
Allow deleting/unregistering event registrations from API

### LINKED ISSUE #1753: Event registrations cannot be deleted from admin API
### Describe the bug
Event registrations cannot be deleted from admin API. Users are able to delete event registrations from the admin, so they should also be allowed to do it via the API

### How to reproduce
1. Try to delete an event registration with the API (as admin)
2. You can't

### Expected behaviour
You should be able to do it

JobDoesburg:
Closes #1753

### Summary
Allow deleting event registrations from the events API

### How to test
1. Try to delete an event registration from the event admin API
2. This should just work
3. Note that this is very different from unregistering.
4. Try to unregister for an event by deleting from the regular events API
5. This should just work

### DIFF
--- a/website/events/api/v2/admin/views.py
+++ b/website/events/api/v2/admin/views.py
@@ -84,7 +84,9 @@ class EventRegistrationAdminListView(AdminListAPIView, AdminCreateAPIView):
         return EventRegistration.objects.none()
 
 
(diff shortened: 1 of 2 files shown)


pingiun (approved):
LGTM

PR #1758, opened 2021-06-21
Fix events api registration fields

### LINKED ISSUE #1754: AssertionError: Cannot apply DjangoAdminModelPermissions on a view that does not set `.queryset` or have a `.get_...
Sentry Issue: [CONCREXIT-6H](https://sentry.io/organizations/thalia/issues/2468530240/?referrer=github_integration)

```
AssertionError: Cannot apply DjangoAdminModelPermissions on a view that does not set `.queryset` or have a `.get_queryset()` method.
(7 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 497, in dispatch
    self.initial(request, *args, **kwargs)
  File "rest_framework/views.py", line 415, in initial
    self.check_permissions(request)
  File "rest_framework/views.py", line 332, in check_permissions
    if not permission.has_permission(request, self):
  File "rest_framework/permissions.py", line 229, in has_permission
    queryset = self._queryset(view)
  File "rest_framework/permissions.py", line 205, in _queryset
    assert hasattr(view, 'get_queryset') \
```

Requesting event registration fields via the API results in a crash

JobDoesburg:
Closes #1754 

### Summary
- The events API will not crash anymore when requesting the fields
- The API users proper DRF Responses
- The API throws an error if you cannot update the fields, rather than ignoring

### How to test
1. Have an event with registration fields
2. Try to update them as user
3. Try to update them as organizer

### DIFF
--- a/website/events/api/v2/admin/views.py
+++ b/website/events/api/v2/admin/views.py
@@ -105,7 +105,9 @@ class EventRegistrationAdminFieldsView(AdminPermissionsMixin, APIView):
 
     permission_classes = [IsOrganiser, IsAuthenticatedOrTokenHasScope]
     required_scopes = ["events:admin"]
-    queryset = EventRegistration.objects.filter(event=self.kwargs["event_id"])
+
+    def get_queryset(self):
+        return EventRegistration.objects.filter(event=self.kwargs["event_id"])
 
     def get_object(self):
         event_registration = get_object_or_404(
--- a/website/events/api/v2/admin/views.py
+++ b/website/events/api/v2/admin/views.py
@@ -1,9 +1,9 @@
-import json
-
-from django.http import HttpResponse, Http404
+from django.http import Http404
 from oauth2_provider.contrib.rest_framework import IsAuthenticatedOrTokenHasScope
 from rest_framework import status
+from rest_framework.exceptions import ValidationError, PermissionDenied
 from rest_framework.generics import get_object_or_404
+from rest_framework.response import Response
 from rest_framework.views import APIView
 from rest_framework import filters as framework_filters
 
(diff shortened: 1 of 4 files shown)


website/events/api/v2/admin/views.py
```diff
@@ -105,6 +106,9 @@ class EventRegistrationAdminFieldsView(AdminPermissionsMixin, APIView):
     permission_classes = [IsOrganiser, IsAuthenticatedOrTokenHasScope]
     required_scopes = ["events:admin"]
 
+    def get_queryset(self):
+        return EventRegistration.objects.filter(event=self.kwargs["event_id"])
```
se-bastiaan, line 110:
```suggestion
    queryset = EventRegistration.objects.filter(event=self.kwargs["event_id"])
```

website/events/api/v2/views.py
```diff
@@ -193,38 +191,45 @@ def get_object(self):
...
-                status=status.HTTP_400_BAD_REQUEST,
+            raise ValidationError(
+                f"Missing keys '{', '.join(required_keys)}' in request",
+                status.HTTP_400_BAD_REQUEST,
             )
+
+        if not event_permissions(
+            self.get_object().member, self.get_object().event, self.get_object().name
+        )["update_registration"]:
+            raise PermissionDenied("You cannot update this registration.")
```
se-bastiaan, line 211: Maybe it would be better to have the `services.update_registration` throw this error. Because this check is already in there: https://github.com/svthalia/concrexit/blob/master/website/events/services.py#L205

website/events/api/v2/views.py
```diff
@@ -193,38 +191,45 @@ def get_object(self):
...
+        return Response(
+            data=services.registration_fields(request, registration=self.get_object()),
             status=status.HTTP_200_OK,
         )
 
     def patch(self, request, *args, **kwargs):
+        if not event_permissions(
+            self.get_object().member, self.get_object().event, self.get_object().name
+        )["update_registration"]:
+            raise PermissionDenied("You cannot update this registration.")
```
se-bastiaan, line 226: See comment above

website/events/api/v2/views.py
se-bastiaan: With that I mean catch a RegistrationError and throw a ValidationError based on it, since that can already happen currently.

pingiun (approved):
Seems to work correctly now

PR #1763, opened 2021-06-22
Delete tpay payment if order is modified

JobDoesburg:
### Summary
Right now it is possible to order a pizza, pay it with tpay, change the order to a pizza with a different price, and the payment will not match the order anymore.

### How to test
1. Order a pizza
2. Pay with tpay
3. Change the order
4. The payment should be deleted
5. If the event is over, or the payment is batched, then changing the order should crash

### DIFF
--- a/website/pizzas/views.py
+++ b/website/pizzas/views.py
@@ -66,7 +66,7 @@ def place_order(request):
         if not obj:
             obj = FoodOrder(food_event=event, member=request.member)
         obj.product = product
-        if obj.payment and obj.can_be_changed:
+        if obj.payment:
             delete_payment(obj.payment)
         obj.save()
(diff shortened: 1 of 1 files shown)


website/pizzas/views.py
```diff
@@ -65,5 +66,7 @@ def place_order(request):
         if not obj:
             obj = FoodOrder(food_event=event, member=request.member)
         obj.product = product
+        if obj.payment and obj.can_be_changed:
```
se-bastiaan, line 69: obj.can_be_changed is always true here

se-bastiaan (changes requested):

se-bastiaan (approved):

PR #1764, opened 2021-06-22
Add renewal upgrade discount help text

pingiun:
### Summary
Added a simple help text to the renewals model. 

image

### How to test
1. Go to membership renewals (/user/membership/)
2. Notice help text
3. Don't be confused about expected price

### DIFF
--- a/website/registrations/forms.py
+++ b/website/registrations/forms.py
@@ -148,7 +148,7 @@ class RenewalForm(forms.ModelForm):
                 reverse_lazy("singlepages:privacy-policy")
             )
         )
-        self.fields["length"].help_text = (
+        self.fields["length"].help_text = _(
             "A discount of €7,50 will be applied if you upgrade your year membership to"
             " a membership until graduation. You will only have to pay €22,50 in that case."
         )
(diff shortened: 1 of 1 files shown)


website/registrations/models.py
```diff
@@ -60,7 +60,11 @@ class Entry(models.Model):
     )
 
     length = models.CharField(
-        verbose_name=_("membership length"), choices=MEMBERSHIP_LENGTHS, max_length=20,
+        verbose_name=_("membership length"),
+        choices=MEMBERSHIP_LENGTHS,
+        max_length=20,
+        help_text="A discount of €7,50 will be applied if you upgrade your year membership to a"
```
se-bastiaan, line 66: This help text is global, it should be in de RenewalForm so that it does not show in the registration form.

se-bastiaan (changes requested):

website/registrations/forms.py
```diff
@@ -148,6 +148,10 @@ def __init__(self, *args, **kwargs):
                 reverse_lazy("singlepages:privacy-policy")
             )
         )
+        self.fields["length"].help_text = (
```
se-bastiaan, line 151:
```suggestion
        self.fields["length"].help_text = _(
```

se-bastiaan (approved):
Have not tested but LGTM

PR #1782, opened 2021-06-24
Fix shift statistics

### LINKED ISSUE #1780: Sales order payment view crashes if user does not have Thalia Pay enabled
### Describe the bug
The javascript in the sales order payment view automatically POST's to the form and thus redirects to the Thalia Pay PaymentProcessView. However, if the user doesn't have Thalia Pay enabled, this results in a 403 (which makes sense in itself).

### How to reproduce
1. Make a sales order
2. Have a user without Thalia Pay enabled scan the QR code
3. Profit

### Expected behaviour
The javascript should only POST if the Thalia Pay button is actually there and is not disabled (the user has Thalia Pay enabled).

### LINKED ISSUE #1781: Shift product sales are not calculated correctly
### Describe the bug
For some reason, the product sales for shifts in the sales app are not calculated properly:

### How to reproduce
1. Check staging.thalia.nu, shift 1
2. 
image
3. It should be waaaaay more as there are 200+ orders

### Expected behaviour
Correct calculation

9Y0:
Closes #1781 

### Summary
The sales statistics were incorrect since the results were not grouped by the `product_id` which is now done by using the `order_by()`

### How to test
1. Create a shift
2. Add an order with 5x beer to the shift
3. Add a new order with 3x beer to the shift
4. Go to the shift in the admin and see that there are 8 beer sales

### DIFF
--- a/website/sales/models/shift.py
+++ b/website/sales/models/shift.py
@@ -153,6 +153,7 @@ class Shift(models.Model):
             .values("order_items__product")
             .distinct()
             .annotate(sold=Sum("order_items__amount"))
+            .order_by()
         )
         return {
             item[0]: item[1]
--- a/website/sales/models/shift.py
+++ b/website/sales/models/shift.py
@@ -151,7 +151,6 @@ class Shift(models.Model):
         qs = (
             self.orders.exclude(order_items__isnull=True)
             .values("order_items__product")
-            .distinct()
             .annotate(sold=Sum("order_items__amount"))
             .order_by()
         )
--- a/website/sales/tests/test_models.py
+++ b/website/sales/tests/test_models.py
@@ -453,6 +453,17 @@ class ShiftTest(TestCase):
         o5.payment = create_payment(o5, processed_by=self.member, pay_type=Payment.CASH)
         o5.save()
 
+        # The test was accidentally passing because orders had the same created at,
+        # resulting in the distinct seeing them as the same order
+        o2.created_at += timezone.timedelta(minutes=1)
+        o3.created_at += timezone.timedelta(minutes=2)
+        o4.created_at += timezone.timedelta(minutes=3)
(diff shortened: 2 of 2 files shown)


se-bastiaan:
What is the distinct for exactly? I would say that it is there to prevent the same order from appearing twice in the list. Should we not just include the field to use the distinct on in the call? 

I am sure you also saw the note in the [Django docs](https://docs.djangoproject.com/en/3.2/ref/models/querysets/#distinct) about combining distinct/values/order_by. And I don't know if this should just be working even if the orders have the same date?

9Y0:
I did some testing and using `distinct`/`order_by` gave the following results:
`distinct`, `order_by`:
```sql
SELECT DISTINCT "sales_orderitem"."product_id", SUM("sales_orderitem"."amount") AS "sold"                             ... GROUP BY "sales_orderitem"."product_id"
```

`order_by`:
```sql
SELECT          "sales_orderitem"."product_id", SUM("sales_orderitem"."amount") AS "sold"                             ... GROUP BY "sales_orderitem"."product_id"
```

`distinct`:
```sql
SELECT DISTINCT "sales_orderitem"."product_id", SUM("sales_orderitem"."amount") AS "sold", "sales_order"."created_at" ... GROUP BY "sales_orderitem"."product_id", "sales_order"."created_at"
```
where I redacted some stuff in the middle since it was the same in every query and not interesting. You can see that the `order_by` is making sure that we're only grouping on the fields which we previously mentioned in the `values` call, which is described [here](https://docs.djangoproject.com/en/3.2/topics/db/aggregation/#interaction-with-order-by). Now that I look at it, the `distinct` is probably redundant since we're already grouping by `product_id` so I'll remove it.

Also I'm pretty sure that this
> Should we not just include the field to use the distinct on in the call?

is not supported by SQLite

se-bastiaan (approved):

Issue #1785, opened 2021-06-26
Add sales order API


DeD1rk:
### Describe the solution you'd like
- An api endpoint `api/v2/sales/order/<uuid>/` that returns the order information that is shown on the website when you pay for the url from a qr code.
- It should be possible to pay the order through `api/v2/payments/sales/order/<uuid>`. This might already be possible, I haven't checked.

### Motivation
This way people will be able to pay from the app, instead of through the website where they may even need to log in. I think this is an obvious case where the ease of use of an app is very welcome.

JobDoesburg:
> An api endpoint api/v2/sales/order/<uuid>/ that returns the order information that is shown on the website when you pay for the url from a qr code.

I don't think this needs to exist. The payable will contain a description of the order and thus contains all information a user needs

> It should be possible to pay the order through api/v2/payments/sales/order/<uuid>. This might already be possible, I haven't checked.

This already exists, it's the whole idea behind payments

> This way people will be able to pay from the app, instead of through the website where they may even need to log in.

They should still scan the QR code to identify themselves _at the counter_. The QR code will simply redirect them to the website. It is fine by me if they are redirected to the app to process the payment, but I don't want to introduce a flow of paying for orders

DeD1rk:
> The payable will contain a description of the order and thus contains all information a user needs. (...) This already exists

If thats the case then indeed no new endpoint is needed. The code of the payment page suggested that paying for sales is more intricate than for other payables, but if it just works through the payments api that's fine.

> They should still scan the QR code to identify themselves at the counter. The QR code will simply redirect them to the website. It is fine by me if they are redirected to the app to process the payment

That is exactly the idea.

JobDoesburg:
Well there is one extra thing, namely that the payer is also set after scanning the QR. This is not yet possible via the api.

DeD1rk:
I guess this is possible now after #1787? I'll take a look if I can figure it out :)

JobDoesburg:
I don't think it is possible yet actually. #1787 only gives access to order that are already yours, but _claiming_ an order that doesn't have a payer is not yet possible

DeD1rk:
That seems right. Right now, `v2/sales/order/<uuid>` raises 403 for paid orders and when user orders are not allowed, and 404 if the user is not the payer. That makes sense for user orders, but doesn't let you claim orders. I'm not sure what would be the best way to make claiming possible:

[...]

Issue #1654, opened 2021-05-10
Investigate rebuilding `pizzas` app on top of `sales` app


JobDoesburg:
### Is your feature request related to a problem? Please describe.
- The current pizzas app is a bit old and not written too nicely
- We currently have a `sales` app that offers functionality that is very related to pizzas 

### Describe the solution you'd like
Rebuild `pizzas` on top of `sales`: 
- Pizza's become products, Pizza Events become Shifts with Product Lists, etc.
- There must be a option for product lists to indicate that 'users can order themselves' . This should enable a concrexit-side view to place orders (right now we're using thadmin via the API that only shift managers can use - with this option enabled, users should also be able to place orders themselves)
- We must think about sending reminders for open orders, indicating whether orders are finished or not, etc. 
 
### Motivation
In the end: simpler and better code

** This issue should be a place to discuss possible design decisions **

DeD1rk:
I think it would be nice to do this. Something to think about is how to determine whether someone is allowed to order only 1 (or maybe up to some other number) item (in a shift, from a product list, or of a specific product).

JobDoesburg:
https://github.com/KiOui/TOSTI/blob/2a3519cd1bb0865b3c33f78a9d3e76fcd97605b5/website/orders/models.py#L268-L274 finally something in the `sales` app that we could actually use the TOSTI code for

se-bastiaan:
A field doesn't really count as using the code 😆

I think the motivation of this issue is wrong. It wouldn't really end up as simpler or better code. The pizzas app is not written perfectly, but it is quite nice because it only does 1 single thing. And it does it well.

The real motivation should be rethinking how often we order food and will do this in the future. The functionality is currently lacking, like food at GMs for example. Or having multiple events at the same time with takeout (should probably be the name new) offerings. More focus on removing code duplication and making this part ready for the future.

I suggest removing the solution you'd like, or at least make it clear that this is 1 possible solution.

PR #1787, opened 2021-06-27
User orders

JobDoesburg:
First draft for #1654 and also implements #1785

### Summary
This PR makes it possible for members to place orders via the API for themselves, during specified time windows within a `Shift`. 

In the future we might 

- allow for different product lists for these time windows
- add HTML pages to do this
- add fields to keep track of whether orders have been handed out (and specify per productlistitem if products actually need to be handed out)

to make this replace the pizzas app


DeD1rk:
Could you perhaps move the #1785 changes (setting payer through the api) to a separate PR first? I can imagine that the entire user orders thing is quite huge so may take a while, and I'm hoping to be done with TPay in the app soon (the dialog for sales payments is already there :)

JobDoesburg:
> Could you perhaps move the #1785 changes (setting payer through the api) to a separate PR first? I can imagine that the entire user orders thing is quite huge so may take a while, and I'm hoping to be done with TPay in the app soon (the dialog for sales payments is already there :)

What I could also do is remove the custom product list from here for now and say that self-order periods use the same product list as the main shift. In that case, this PR is basically done now and I don't have to go through the hassle of splitting it into two PRs :)

DeD1rk:
Sounds perfect🤩

pingiun (approved):

PR #1789, opened 2021-06-29
Do not submit form onload if tpay not enabled

### LINKED ISSUE #1780: Sales order payment view crashes if user does not have Thalia Pay enabled

JobDoesburg:
Closes #1780 

### Summary
The javascript in the sales order payment view will not automatically try to POST a payment if the user does not have TPAY enabled

### How to test
1. Have a sales order
2. Don't have tpay enabled
3. Try to scan the QR code
4. It will not crash

### DIFF
--- a/website/sales/templates/sales/order_payment.html
+++ b/website/sales/templates/sales/order_payment.html
@@ -20,8 +20,9 @@
     {{ block.super }}
     <script type="text/javascript">
         window.onload=function() {
-            if (!document.getElementById("tpay-submit-button").disabled)
+            if (!document.getElementById("tpay-submit-button").disabled) {
                 document.forms['tpay-payment-form'].submit();
+            }
         }
     </script>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


website/sales/templates/sales/order_payment.html
```diff
@@ -20,7 +20,8 @@ <h1 class="section-title text-center">{% trans "Pay for order" %}</h1>
     {{ block.super }}
     <script type="text/javascript">
         window.onload=function() {
-            document.forms['tpay-payment-form'].submit();
+            if (!document.getElementById("tpay-submit-button").disabled)
```
se-bastiaan, line 23: Please use `{}` here :)

se-bastiaan (approved):

PR #1790, opened 2021-06-29
Change scopes for sales API

### LINKED ISSUE #1775: Change sales API scope to `sales:admin` instead of `read` and `write`
### Describe the change
Change sales API scope to `sales:admin` instead of `read` and `write`

### Motivation
Follow the same convention as other apps

### Current implementation
`sales:read` and `sales:write` scopes are used

### Suggested implementation
Use a single `sales:admin` scope

JobDoesburg:
Closes #1775 

### Summary
Use a consistent scopes in the sales app as in other apps 

### How to test
1. Check that the sales API still works (every view is still accessible if you have the `sales:admin` scope)

### DIFF
--- a/website/thaliawebsite/settings.py
+++ b/website/thaliawebsite/settings.py
@@ -533,7 +533,8 @@ OAUTH2_PROVIDER = {
         "payments:write": "Write access to payments",
         "payments:admin": "Admin access to payments",
         "sales:read": "Read access to Point of Sale orders",
-        "sales:write": "Write access to Point of Sale orders",
+        "sales:order": "Write access to Point of Sale orders",
+        "sales:admin": "Admin access to Point of Sale orders",
     },
 }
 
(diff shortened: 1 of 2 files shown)


JobDoesburg:
We could decide to split it in `sales:admin:read` and `sales:admin:write` but then we should do it for all other api's too

website/thaliawebsite/settings.py
```diff
@@ -533,7 +533,8 @@ def from_env(
         "payments:write": "Write access to payments",
         "payments:admin": "Admin access to payments",
         "sales:read": "Read access to Point of Sale orders",
-        "sales:write": "Write access to Point of Sale orders",
+        "sales:order": "Write access to Point of Sale orders",
```
se-bastiaan, line 536: Should this not be 'order access?'

se-bastiaan:
> 
> 
> We could decide to split it in `sales:admin:read` and `sales:admin:write` but then we should do it for all other api's too

I had something like that in the beginning, but the API is also checked by the model permissions so I decided that would be a bit toooo much.

se-bastiaan (approved):

PR #1791, opened 2021-06-29
Use color variable for events admin payment dropdown text

### LINKED ISSUE #1777: Payment not visible
### Describe the bug
In the administration page of an event for which you have to pay the payment method is invisible on some browsers. Clicking on the payment does still reveal the menu with the options. See the screenshots below.

### How to reproduce
1. Go to the admin page of an event on Chrome or Firefox.
2. See the payment method

### Screenshots
image
image
Expected:
image

### Additional context
I don't have the problem on Safari, I do have it on Chrome and I heard it's also a problem on Firefox.
Maybe it has something to do with the darkmode. I also don't understand why dark mode doesn't seem to activate on Chrome for me.

JobDoesburg:
Closes #1777

image

### Summary
Changes the payment dropdown in the events admin to use color variables, so the widget is shown correctly in light mode as well

### How to test
1. Visit the events admin in multiple browsers, in dark mode and light mode
2. See the issue for the problematic cases 
3. Don't forget to clear your cache

### DIFF
--- a/website/events/static/events/css/admin.scss
+++ b/website/events/static/events/css/admin.scss
@@ -48,7 +48,6 @@
 
         select[name=payment] {
             background: var(--message-error-bg);
-            color: var(--text-color);
 
             &.paid {
                 background: var(--message-success-bg);
--- a/website/events/static/events/css/admin.scss
+++ b/website/events/static/events/css/admin.scss
@@ -48,7 +48,7 @@
 
         select[name=payment] {
             background: var(--message-error-bg);
-            color: white;
+            color: var(--text-color);
 
             &.paid {
                 background: var(--message-success-bg);


website/events/static/events/css/admin.scss
```diff
@@ -48,7 +48,7 @@
 
         select[name=payment] {
             background: var(--message-error-bg);
-            color: white;
+            color: var(--text-color);
```
se-bastiaan, line 51: This technically does not work due to the variable not existing. However, it appears because the default colour for form element is `var(--body-fg)`. So we can just remove this property.

```suggestion
```

se-bastiaan (changes requested):

website/events/static/events/css/admin.scss
JobDoesburg: Wot hahah, my IDE said it did exist. But sure

se-bastiaan (approved):

PR #1792, opened 2021-06-29
Optional event registration tweaks

### LINKED ISSUE #1771: Disable registration-if-not-required for events in history
### Describe the bug
#1574 introduced the feature that people can indicate their presence for an event, even if registration is not required. However, this also works if the event took place in the past.

### How to reproduce
1. Have an event
2. Don't let it have registrations required
3. Have it be in the past
4. Try to register
5. It is possible!

### Expected behaviour
It should not be possible. Maybe, I think there should even be an option for events to allow this or not.

JobDoesburg:
Closes #1771

### Summary
#1574 enabled a new feature for a lot of events. However, it also 'reopened' event registration for events in the past. This PR fixes that, and makes this whole feature optional, so event organizers can also disable it.

### How to test
1. Have an event
2. You cannot register
3. Make optional registrations allowed
4. You can register
5. Check out other edge cases


se-bastiaan:
Do we really want this to be optional? Don't really understand why we would need that.

JobDoesburg:
> Do we really want this to be optional?
 
Events is the main functionality of our website and I think there could be situations where the opt-in registrations could be confusing. For example, events where ticket sale happens via a different system. Then you could have the situation where people are indicating their presence even though they don't have a ticket. I think for such a visible feature that could really have quite some impact, it should be possible to switch it off

se-bastiaan:
> > Do we really want this to be optional?
> 
> Events is the main functionality of our website and I think there could be situations where the opt-in registrations could be confusing. For example, events where ticket sale happens via a different system. Then you could have the situation where people are indicating their presence even though they don't have a ticket. I think for such a visible feature that could really have quite some impact, it should be possible to switch it off

That is a perfect example that I had not thought of 👍

website/events/models/event.py
```diff
@@ -97,6 +97,18 @@ class Event(models.Model):
         ),
     )
 
+    optional_registrations = models.BooleanField(
+        _("allow optional registrations"),
+        default=False,
```
se-bastiaan, line 102: I think this value should be True by default, else nobody will enable it.

website/events/services.py
```diff
@@ -82,20 +82,20 @@ def event_permissions(member, event, name=None):
 
     perms["create_registration"] = (
         (registration is None or registration.date_cancelled is not None)
-        and (event.registration_allowed or not event.registration_required)
+        and (event.registration_allowed or event.optional_registration_allowed)
```
se-bastiaan, line 85: I was thinking about this, and haven't tested.

But I think this allows registration for events that do require registration while that period hasn't opened yet.

se-bastiaan: Pretty sure it should be

```suggestion
        and (event.registration_allowed or (event.optional_registration_allowed and not event.registration_required))
```

se-bastiaan: maybe even move that logic into the model?

JobDoesburg: you are right

se-bastiaan (approved):

Issue #1766, opened 2021-06-22
Immutable payables


JobDoesburg:
### Describe the change
A common problem, also in our current codebase, is that we often want to prevent changing payables when they are already paid. Think of Food/Sales orders. Right now there is no good way to fix this fundamentally. Sales orders only prevent saving when the amount of the payment and the order don't match (but doesn't update the description, for example). Food orders since #1765 only does the check in the view.

A way to solve this fundamentally would be the following:
- Payables get an additional boolean property `immutable_after_payment` that defaults to False, but can be enabled for models where we want to apply this
- The payments app implements a pre_save_signal on payable models that raises an exception if a model that is `immutable_after_payment` and has a payment is tried to be saved. (Or it tries to delete the payment and will raise the exception when that is not possible anymore - for example because the payment has already been processed).

### Motivation
This is a problem that applies to all payables and could be a generic fix.

### Current implementation
Messy, not really 

### Suggested implementation
See above

se-bastiaan:
It took me a while to understand what the issue is about, your explanation of the problem is really unclear. 
> we often want to prevent changing payables
But why do you want to do that? This is not directly mentioned in the current description.

From what I understand this issue is about the payables that can be changed while already having a payment. Changing the payable does not change the existing payment, which will mean that the values will not be up-to-date.

This is the case when the title of an event changes, or maybe the type of pizza in a food order. You could probably name some of the other cases as well. In the food order case we already check if this may cause problems and block updates, but there are cases where these checks are currently not sufficient.

Therefore we would like to implement some sort of immutability for the models that are registered as a Payable.

---

In my opinion making the models that are registered as a payable immutable makes a lot of sense, I would no implement this as an option but force it for all of them. I am not sure if a signal will correctly block saving the models, have never tried that before.

JobDoesburg:
>  I am not sure if a signal will correctly block saving the models, have never tried that before.
Yes it does, at least for individual save's. Not for bulk creates though.

PR #1793, opened 2021-06-29
Delete payments on FoodOrder update via API

### LINKED ISSUE #1783: Food order can be changed after paying
### Describe the bug
If you order a pizza and pay it, you can still change the product. If you change the product through the api, the payment is not removed.

### How to reproduce
1. Order a pizza
2. Pay with Thalia Pay
3. Change the order through the api
4. Get an expensive pizza for little money

### Expected behaviour
Either changing the order after paying is impossible, or it removes the payment. I think removing the payment (as the website currently seems to do) would be strange, and for event registration we've decided not to enable this.

### Screenshots
image

There are probably no food events any time soon, so a hotfix may not be needed, though it might be good to double-check that similar stuff is not possible with registrations.

JobDoesburg:
Closes #1783

### Summary
Adds a check to the FoodOrderDetail API view that tries to delete a payment if the order is updated

### How to test
1. Have a food order
2. Pay it with Thalia Pay
3. Change to a more expensive pizza
4. The payment is gone (or the update is not accepted)

### DIFF
--- a/website/pizzas/api/v2/views.py
+++ b/website/pizzas/api/v2/views.py
@@ -12,6 +12,8 @@ from rest_framework import filters as framework_filters, status
 from rest_framework.permissions import DjangoModelPermissionsOrAnonReadOnly
 from rest_framework.response import Response
 
+from payments.exceptions import PaymentError
+from payments.services import delete_payment
 from pizzas.api.v2 import filters
 from pizzas.api.v2.serializers import (
     ProductSerializer,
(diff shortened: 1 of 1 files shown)


DeD1rk:
I think something else is the real problem: https://github.com/svthalia/concrexit/blob/6a42032f4ffadf993267b6cacce0ccf7a493df0d/website/pizzas/models.py#L249 
the order model's can_be_changed is true when paid with TPay. Both the website and the api should probably use that property, as in https://github.com/svthalia/concrexit/blob/6a42032f4ffadf993267b6cacce0ccf7a493df0d/website/pizzas/views.py#L54

Although I think this PR does solve the problem in the API, it's probably better to fix the `can_be_changed` for both website and API.

DeD1rk:
Also, do we want it to be possible to change a paid order at all? It feels vague to try to remove the payment ("if you pay you might be able to change your order later, but we're not sure"). I think it would be more logical to just forbid changing orders after paying, unless the "payment can't be removed" case will normally never occur.

JobDoesburg:
The thing is that people at borrels are drunk and they will order a pizza and immediately pay with Thalia Pay, with the consequence that they will not be able to change the order anymore. Better UX would be that they can still try to change their order if it's possible. 

I do agree that we should implement something at the model level. That would be immutable payables, #1772 and #1766.

But for now I think it is a proper fix that implements exactly the same as the html view

DeD1rk:
> people at borrels are drunk 

Fair enough, then the `can_be_changed` is also correct for now.

DeD1rk (approved):

PR #1796, opened 2021-06-30
Added API endpoint for pushnotification categories

KiOui:
Added API endpoint for pushnotification categories.

Testing:

Go to `/api/docs`, select the v2 api docs and see if the endpoint works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,3 @@ result-*
 # direnv
 .envrc
 .env*
-
-# macOS
-.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ result-*
 # direnv
 .envrc
 .env*
+
+# macOS
+.DS_Store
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -37,3 +37,6 @@ result-*
 # direnv
 .envrc
 .env*
+
+# macOS
+.DS_Store
```
JobDoesburg, line 42: Let's do this in a separate PR

JobDoesburg (changes requested):

website/pushnotifications/api/v2/views.py
```diff
@@ -74,6 +74,16 @@ def perform_update(self, serializer):
         serializer.save(user=self.request.user)
 
 
+class CategoryListView(ListAPIView):
+    """Returns an overview of all available categories for push notifications."""
+
+    serializer_class = CategorySerializer
+    queryset = Category.objects.all()
+    required_scopes_per_method = {
```
JobDoesburg, line 82: This does not need to be per method

JobDoesburg (approved):

se-bastiaan (approved):

PR #1803, opened 2021-07-01
Set allowed API versions to fix API v1 default pagination

se-bastiaan:
### Summary
Set allowed API versions to fix API v1 default pagination

### How to test
1. Go to /api/v1/members/
2. See that it is not paginated by default
3. Go to the v2 version and see that there is pagination


JobDoesburg:
The API schema data is broken in this PR
image

image

JobDoesburg (approved):

PR #1843, opened 2021-07-27
Add partners and partner events API v2

### LINKED ISSUE #1428: Introduce API v2

se-bastiaan:
Closes #1428 

### Summary
Adds API v2 for partners apps.

### How to test
1. Go to /api/docs and select v2.
2. Check the partners endpoints

### DIFF
--- a/website/partners/api/v2/serializers/partner_event.py
+++ b/website/partners/api/v2/serializers/partner_event.py
@@ -1,0 +1,16 @@
+from rest_framework import serializers
+
+from partners.models import PartnerEvent
+from thaliawebsite.api.v2.serializers import CleanedHTMLSerializer
+
+
+class PartnerEventSerializer(serializers.ModelSerializer):
+    """Partner events serializer."""
+
+    class Meta:
+        """Meta class for partner events serializer."""
+
+        model = PartnerEvent
+        fields = ("pk", "title", "description", "start", "end", "location", "url")
+
+    description = CleanedHTMLSerializer()
--- a/website/partners/api/v2/filters.py
+++ b/website/partners/api/v2/filters.py
@@ -1,0 +1,35 @@
+from rest_framework import filters
+
+from utils.snippets import extract_date_range
+
+
+class PartnerEventDateFilter(filters.BaseFilterBackend):
+    """Allows you to filter by partner event start/end dates."""
+
+    def filter_queryset(self, request, queryset, view):
+        start, end = extract_date_range(request, allow_empty=True)
+
+        if start is not None:
+            queryset = queryset.filter(end__gte=start)
+        if end is not None:
+            queryset = queryset.filter(start__lte=end)
+
+        return queryset
+
+    def get_schema_operation_parameters(self, view):
+        return [
+            {
+                "name": "start",
+                "required": False,
+                "in": "query",
+                "description": "Filter events by ISO date, starting with this parameter (i.e. 2021-03-30T04:20:00) where `event.end >= value`",
+                "schema": {"type": "string",},
+            },
+            {
+                "name": "end",
+                "required": False,
+                "in": "query",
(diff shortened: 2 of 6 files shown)


DeD1rk:
deploy review

DeD1rk:
remove review

website/partners/api/v2/serializers/partner_event.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class PartnerEventSerializer(serializers.ModelSerializer):
+    """Partner events serializer."""
+
+    class Meta:
+        """Meta class for partner events serializer."""
+
+        model = PartnerEvent
+        fields = ("pk", "title", "description", "start", "end", "location", "url")
```
DeD1rk, line 14: This does not include the partner, which I would expect to be given (even though it might be null or a string). It's not a problem per se but may be something to add.

DeD1rk (changes requested):

website/partners/api/v2/serializers/partner_event.py
DeD1rk: Also note that there's no `/partners/:id` endpoint which is also okay but maybe strange.

se-bastiaan: Partner events are not attached to partners. They are only links to the website of their events. So there is nothing more to add. Would probably be more appropriate to call them 'external events' and just move the model to the events package.

DeD1rk: Still their model does include a partner or 'other partner' string. I do agree moving them to events app would be a good idea.

DeD1rk (approved):

PR #1867, opened 2021-08-09
Remove cache from payable registry

### LINKED ISSUE #1865: Fix payable registry cache using old data
### Describe the bug
Payments are a mess. For example, if you pay for an event registration, delete the payment (through the admin or api), creating a new TPay payment through the api fails with 409 Conflict, there is still a payment in the registration model, but not in the payments api. Furthermore, paying with a different payment type works, but those payments can then not be removed. Also see #1806. I suspect there are many more related problems.

### How to reproduce
Play around with payable things, deleting and recreating them, or deleting and recreating payments.

### Expected behaviour
When a payable is not paid and should be payable with TPay, paying does not fail. Deleting a payment makes the payable not-paid as it was before creating the payment. Deleting or changing a payable is either impossible, or also deletes a payment that belongs to it.

### Additional context
I think it would be a good idea to combine this with #1000. Some test-driven development would make sense for payments, and I think the expected behaviour should be well-testable. Of course the problems may not be entirely within the payments app, but also in the payables defined by other apps.

JobDoesburg:
Closes #1865

### Summary
The payable registry could use old data when querying a payable while a payment was deleted for that object. This lead to weird behavior.

### How to test
Previously:

1. Have a payable object, for example an event registration
2. Add a payment for it (write down the uuid and created_at if you want to see something funny)
3. Delete the payment
4. Inspect that the payment is really really deleted
5. Re-create a payment
6. The same payment with the same uuid and created_at re-appears in the database 🤯 

Now:

1. Have a payable object, for example an event registration
2. Add a payment for it
3. Delete the payment
4. Re-create the payment
5. The payment is re-created and has a different uuid and created_at field

Also:

1. Have a payable object, for example an event registration
2. Add a payment for it
3. Delete the payment
4. Trying to re-create the payment via API will not give errors saying that it is already paid!

### DIFF
--- a/website/payments/payables.py
+++ b/website/payments/payables.py
@@ -52,6 +52,7 @@ class Payable:
 class Payables:
     _registry = {}
 
+    @lru_cache(maxsize=None)
     def _get_key(self, model):
         return f"{model._meta.app_label}_{model._meta.model_name}"
 
--- a/website/payments/payables.py
+++ b/website/payments/payables.py
@@ -52,11 +52,9 @@ class Payable:
 class Payables:
     _registry = {}
 
-    @lru_cache(maxsize=None)
     def _get_key(self, model):
         return f"{model._meta.app_label}_{model._meta.model_name}"
 
-    @lru_cache(maxsize=None)
     def get_payable(self, model: Model) -> Payable:
         if self._get_key(model) not in self._registry:
             raise NotRegistered(f"No Payable registered for {self._get_key(model)}")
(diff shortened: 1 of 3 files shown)


website/payments/payables.py
```diff
@@ -52,11 +52,9 @@ def can_manage_payment(self, member):
 class Payables:
     _registry = {}
 
-    @lru_cache(maxsize=None)
```
se-bastiaan, line 55: This can probably stay?

se-bastiaan (changes requested):

se-bastiaan (dismissed):

se-bastiaan (approved):

PR #1870, opened 2021-08-10
Add user registration to events API v2 for cancelled situations

### LINKED ISSUE #1863: Add cancelled registrations to event in api v2
API v2 doesn't return a cancelled registration when you've cancelled a registration (in the event serializer). So we cannot display stuff like 'Your registration is cancelled (after the deadline)'.

Also return a registration if it's cancelled, and include fields like is_cancelled, is_late_cancellation, etc.

### Motivation
This makes it possible to show nice status texts like in the old app and the website.

Leaving out the messages for cancelled registrations. In that case, you can only see that there is no register button, but not why.

se-bastiaan:
Closes #1863

### Summary
Add user registration to events API v2 for cancelled situations

### How to test
1. Go to /api/docs
2. Check api v2 event registrations for users in the event serializer

### DIFF
--- a/website/events/api/v2/serializers/event.py
+++ b/website/events/api/v2/serializers/event.py
@@ -58,12 +58,20 @@ class EventSerializer(serializers.ModelSerializer):
         try:
             if self.context["request"].member:
                 reg = instance.eventregistration_set.get(
-                    member=self.context["request"].member, date_cancelled=None
+                    member=self.context["request"].member
                 )
                 return EventRegistrationSerializer(
                     reg,
                     context=self.context,
-                    fields=("pk", "present", "queue_position", "date", "payment"),
+                    fields=(
+                        "pk",
+                        "present",
+                        "queue_position",
+                        "is_cancelled",
+                        "is_late_cancellation",
+                        "date",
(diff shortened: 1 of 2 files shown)


JobDoesburg:
Maybe we want to have api filters on this?

se-bastiaan:
> Maybe we want to have api filters on this?

I do not understand your question, what do you want to filter on exactly?

JobDoesburg:
> > Maybe we want to have api filters on this?
> 
> I do not understand your question, what do you want to filter on exactly?

Filter on whether we want to include cancelled registrations or not.

se-bastiaan:
> 
> 
> > > Maybe we want to have api filters on this?
> > 
> > 
> > I do not understand your question, what do you want to filter on exactly?
> 
> Filter on whether we want to include cancelled registrations or not.

@JobDoesburg That has nothing to do with this issue >.< This is about showing the cancelled registration from the user itself. We already have a filter in the admin API: https://github.com/svthalia/concrexit/blob/master/website/events/api/v2/admin/filters.py#L29

DeD1rk (approved):

PR #1875, opened 2021-08-12
Add initial profile fields to registrations

### LINKED ISSUE #1857: Optional profile picture field during registration
Only few people actually have a profile picture on the website. It's nice to see faces of people to recognize them

Give people the opportunity to upload a photo during registration

### Motivation
This will likely result in more faces on the website

Don't do this during registration, but rather redirect people on first login to the 'change profile' page and change the email text to explicitly ask them to upload a photo. 

### Additional context
If I remember correctly, @se-bastiaan came up with this idea

JobDoesburg:
Closes #1857 

### Summary

### How to test


JobDoesburg:
@svthalia/technicie I've been thinking about this, and I think it isn't worth the code complexity. I propose to implement some things to motivate new members to update their profile. For example:

- Change the email confirmation message (that also contains the password when the account is created)
- Redirect users to the profile-edit-page on first login
- Show a (closable) banner for new members with an 'empty' profile

What do you think?

website/registrations/admin.py
```diff
@@ -113,9 +120,19 @@ class RegistrationAdmin(admin.ModelAdmin):
             {"fields": ("student_number", "programme", "starting_year",)},
         ),
     )
+    readonly_fields = ("_photo",)
```
se-bastiaan, line 123: I think it is better to not mix two ways of setting the read-only fields. This simply be defined inside `get_readonly_fields`

website/registrations/services.py
```diff
@@ -264,8 +265,14 @@ def _create_member_from_registration(registration: Registration) -> Member:
         birthday=registration.birthday,
         show_birthday=registration.optin_birthday,
         receive_optin=registration.optin_mailinglist,
+        emergency_contact=registration.emergency_contact,
+        emergency_contact_phone_number=registration.emergency_contact_phone_number,
+        photo=registration.photo,  # TODO change the image path to the correct directory and verify thumbnails
```
se-bastiaan, line 270: You don't need to verify thumbnails, those are created on the first time a specific size of the image is requested.

JobDoesburg:
@se-bastiaan what do you think about this issue in general (see my previous comment)

se-bastiaan:
> @svthalia/technicie I've been thinking about this, and I think it isn't worth the code complexity. I propose to implement some things to motivate new members to update their profile. For example:
> 
>     * Change the email confirmation message (that also contains the password when the account is created)
> 
>     * Redirect users to the profile-edit-page on first login
> 
>     * Show a (closable) banner for new members with an 'empty' profile
> 
> 
> What do you think?

Missed this, but I fully agree. Especially on the photo part. So complex with moving around the file. Not sure what the best solution is. A redirect is probably weird, but the banner and email are a good idea and easy to implement.

pingiun:
> @svthalia/technicie I've been thinking about this, and I think it isn't worth the code complexity. I propose to implement some things to motivate new members to update their profile. For example:
> 
> * Change the email confirmation message (that also contains the password when the account is created)
> * Redirect users to the profile-edit-page on first login
> * Show a (closable) banner for new members with an 'empty' profile
> 
> What do you think?

I think all of those things would be good to implement, nice ideas

JobDoesburg:
Decided to close this

PR #1893, opened 2021-08-27
Add education document notification email

### LINKED ISSUE #1725: Send an email to the Educacie when there is a new document to approve
Currently, the Educacie has to manually check if there are any documents (i.e. exams and summaries) to approve.  

Send the Educacie an email when there is a new document to approve.

### Motivation
As the committee needs to manually check, this only happens a few times a year. If a summary is submitted a few days before an exam and the committee checks after the exam week, students will not be able to access the summary in time.

My first idea was to publish the files automatically and remove bad documents when they are checked. However, Isolde told me that a significant portion of the files is not approved, which means that this solution would result in a lot of poor quality/joke documents on the site.

### Additional context
I discussed this with Isolde and this is the solution that the Educacie prefers.

se-bastiaan:
Closes #1725 

### Summary
This PR introduces a notification email for the Educacie so that they know when new documents have been submitted.

### How to test
1. Go to exam/summary submission
2. Do a new submission
3. Check that an email is sent to the notification address. It is possible that this email address does not exist on staging. Locally the email is shown in the console.

### DIFF
--- a/website/education/emails.py
+++ b/website/education/emails.py
@@ -1,0 +1,34 @@
+"""The emails defined by the education package."""
+import logging
+
+from django.conf import settings
+from django.core import mail
+from django.template import loader
+from django.urls import reverse
+
+logger = logging.getLogger(__name__)
+
+
+def send_document_notification(document):
+    """Send an email to a configured email.
+
+    :param document: The document we are sending the email for
+    """
+    info = (document._meta.app_label, document._meta.model_name)
+    admin_url = reverse("admin:%s_%s_change" % info, args=(document.pk,))
+
+    email_body = loader.render_to_string(
+        "education/email/document_notification.txt",
+        {
+            "document": document.name,
+            "uploader": document.uploader.get_full_name(),
+            "url": settings.BASE_URL + admin_url,
+            "course": f"{document.course.name} ({document.course.course_code})",
+        },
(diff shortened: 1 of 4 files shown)


website/education/emails.py
```diff
@@ -0,0 +1,34 @@
...
+        "education/email/document_notification.txt",
+        {
+            "document": document.name,
+            "uploader": document.uploader.get_full_name(),
+            "url": settings.BASE_URL + admin_url,
+            "course": f"{document.course.name} ({document.course.course_code})",
+        },
+    )
+    mail.EmailMessage(
+        "New document submitted",
```
pingiun, line 30: I think this subject can be more descriptive, what do you think of:

```suggestion
        "Education document ready for review",
```

pingiun:
I think it might be even better if these emails are batched per day, however the Educacie has said that they would like some kind of notifications so let's try this first

se-bastiaan:
> I think it might be even better if these emails are batched per day, however the Educacie has said that they would like some kind of notifications so let's try this first

Yes I agree, but even though I cannot see the document I do not remember a lot of documents uploaded every day. And with a lot I mean amounts that would be annoying to receive emails about.

pingiun (approved):

PR #1913, opened 2021-09-09
Add elaborate access logging

pingiun:
### Summary
Add timing information to the nginx access logs

### How to test
1. Ssh to the server
2. Run command like `nix-shell -p angle-grinder --run "tail -f /var/log/nginx/concrexit.access.log | agrind '* | logfmt | p50(request_time) by upstream_status, request'"`
3. See timing information

If you have nix you can run the VM for this with `./scripts/test-vm.sh`

### DIFF
--- a/nix/configuration.nix
+++ b/nix/configuration.nix
@@ -242,6 +242,22 @@ in
         recommendedOptimisation = true;
         recommendedTlsSettings = true;
 
+        commonHttpConfig = ''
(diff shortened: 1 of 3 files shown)


se-bastiaan:
This won't increase the logs size by gigabytes somehow?

pingiun:
> This won't increase the logs size by gigabytes somehow?

It's not that much more information and we can just setup logrotate to throw away logs faster

se-bastiaan (approved):

Issue #880, opened 2020-03-12
Data minimization on payments and event registrations after 7 years.


thaliawww-service:
In GitLab by @JobDoesburg on Jun 2, 2019, 15:35

### One-sentence description

Data minimization on payments and event registrations after 7 years.

### Motivation

The privacy office of the RU has indicated that there are no grounds for Thalia to keep data of payments or event participation after 7 years (the legal obligatory time to keep this information). So in order to be fully compliant to the AVG/GDPR we should maybe delete this information after 7 years.

### Desired functionality

7 years after an event, information of who attended what activity should be removed and whether they paid, and how. 
It would be nice if we can keep track of the amount of participants at that activity.

### Suggested implementation

Apart from that I can imagine this is not quite easy, I have no idea.

Note that this request is only based on information provided by the Radboud University privacy office. But the Technicie I'm sure also have an grounded opinion about this. So if you guys think it is unnecessary to remove this data because of reasons, and you think it is compliant to the AVG/GDPR that is fine as well! This is just passing on information actually.

thaliawww-service:
In GitLab by @se-bastiaan on Jun 3, 2019, 17:20

> It would be nice if we can keep track of the amount of participants at that activity.

If we replace the member in the registration with a value in the name field we essentially removed the information and kept the amount of registrations. You know how many registrations and whether they were present (a value we could set to `False`) is all that would be left. Especially since we can easily remove payments older than 7 years, which would set the payment status to 'No payment'.

Don't really think it is that hard actually. Just needs careful execution and tests.

thaliawww-service:
In GitLab by @se-bastiaan on Jun 3, 2019, 17:20

changed due date to June 01, 2019

thaliawww-service:
In GitLab by @se-bastiaan on Jun 3, 2019, 17:20

changed due date to June 01, 2022

pingiun:
As we will be using payments more and more, this is becoming a high priority

PR #1935, opened 2021-09-20
Remove event registration information after 7 years

pingiun:
Part of #880

### Summary
Change the `dataminimisation` management command to also delete information about event registrations

### How to test
1. Have very old event (7 years ago)
2. Run `poetry run website/manage.py dataminimisation`
3. See that member foreign key and payment are set to None and name is set to `<removed>`

### DIFF
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -1,5 +1,6 @@
 from collections import OrderedDict
 
+from django.db.models import Q
 from django.utils import timezone
 from django.utils.datetime_safe import date
 from django.utils.translation import gettext_lazy as _
@@ -336,3 +337,16 @@ def generate_category_statistics():
         }
 
     return data
+
+
+def execute_data_minimization(dry_run=False):
+    """Delete information about very old events."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 7))
+
+    queryset = EventRegistration.objects.filter(event__end__lte=deletion_period).filter(
+        Q(payment__isnull=False) | Q(member__isnull=False) | ~Q(name__exact="<removed>")
+    )
+    if not dry_run:
+        queryset.update(payment=None, member=None, name="<removed>")
+    return queryset.all()
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -342,7 +342,7 @@ def generate_category_statistics():
 def execute_data_minimization(dry_run=False):
     """Delete information about very old events."""
     # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
-    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 7))
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 5))
 
     queryset = EventRegistration.objects.filter(event__end__lte=deletion_period).filter(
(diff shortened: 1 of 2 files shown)


website/thaliawebsite/management/commands/dataminimisation.py
```diff
@@ -1,6 +1,7 @@
 from django.core.management.base import BaseCommand
 
-from members import services
+from members import services as members_services
+from events import services as events_services
 
 
 class Command(BaseCommand):
```
se-bastiaan, line 7: If we want this command as general purpose minimisation command it should probably move out of the members app.

pingiun: It is in the `thaliawebsite` app

Jorritboer (dismissed):

JobDoesburg:
In my opinion, we should actually keep the **amount** of registrations for the events. I do think it is worth to keep this information in some way.

website/events/services.py
```diff
@@ -336,3 +337,16 @@ def generate_category_statistics():
...
     return data
+
+
+def execute_data_minimization(dry_run=False):
+    """Delete information about very old events."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 7))
+
+    queryset = EventRegistration.objects.filter(event__end__lte=deletion_period).filter(
+        Q(payment__isnull=False) | Q(member__isnull=False) | ~Q(name__exact="<removed>")
```
JobDoesburg, line 348: Does this work for both free events and paid events?

JobDoesburg (changes requested):

website/events/services.py
JobDoesburg: Also, does this work for non-member registrations?

JobDoesburg: Finally, event registration information fields should probably be deleted too (though one could argue about this)

se-bastiaan: Read the queryset job :p 

Free and paid events are not different, this just gets _all_ registrations from events that are more than 7 years ago.
First filter gets all those.

Second filter gets all the registrations that match one of the following condititions:
1. The payment is not null
2. The member is not null
3. The name (thus non-member) is not `<removed>`

Agree on removing information fields data.

Maybe that this information can be minimised much earlier than 7 years? We don't need to keep it, the payment information is in the payments themselves and the amount of people that went to the event is preserved using this method.

pingiun: yea i think something like 5 years could be good. it is sometimes nice to look back at old events and it's weird to see "<deleted>" everywhere, i think it would inhibit the nostalgia somewhat. so 5 years could strike the balance

JobDoesburg: Wild idea (maybe too complex though): add a "remove data after" field to registration info fields. As the kind of data in there can differ, maybe our retention policy should also be set per case. For example, when visiting an in-house day at the ministry of defense where we ask for a pasport number, we might want to set a deletion date of the day after the event, while for other information we might want a longer period. This way, we also enforce people to think about when to delete the data (and we could have that field be max. 5 years)

se-bastiaan: Kind of makes sense, but not something that should be done in this PR.

JobDoesburg: Agreed

se-bastiaan (approved):

PR #1938, opened 2021-09-20
Fix membership type displaying wrong initial date

### LINKED ISSUE #1929: Membership since displaying wrong date
### Describe the bug
Some members their membership type displays the wrong membership since.
Stating a later year's membership than their first membership (my account displayed 01-09-2020, whilst I have been a member since 01-09-2019 (board fixed this)).

### How to reproduce
1. Go to an account with a wrongly listed membership (e.g. Dirk's https://thalia.nu/members/profile/1902)
2. Scroll down to 'Personal Information'
4. See error in Membership type

### Expected behaviour
The first membership should be displayed

### Screenshots
image

### Additional context
Probable error caused in the rendering template 
https://github.com/svthalia/concrexit/blob/f1cf6819c67cdef916ea9a69e2d12353893fd8f1/website/members/templates/members/user/profile.html#L43
latest_membership should probably be a reference to the first membership instead.

mark-boute:
Closes #1929 

### Summary
Now shows the first since date of the current membership type.
Useful for members who have yearly subscriptions for Thalia.

### How to test
1. Go to '/members/profile'
2. Click on 'see your membership type's since is correct.'
3. Test for multiple configurations of memberships.

### DIFF
--- a/website/members/models/member.py
+++ b/website/members/models/member.py
@@ -118,6 +118,17 @@ class Member(User):
         return self.membership_set.latest("since")
 
     @property
+    def oldest_membership_of_current_type(self):
+        """Get the first membership of the most recent membership's type."""
+        if not self.membership_set.exists():
+            return None
+        return (
+            self.membership_set.filter(type=self.latest_membership.type)
+            .order_by("since")
+            .first()
+        )
+
+    @property
     def earliest_membership(self):
         """Get the earliest membership of this user."""
         if not self.membership_set.exists():
--- a/website/members/templates/members/user/profile.html
+++ b/website/members/templates/members/user/profile.html
@@ -17,7 +17,7 @@
             <div class="row">
                 <div class="pb-4 pb-lg-0 col-12 col-lg-7 text-center">
                     {% if not member.profile.photo %}
-                        <img src="{% static "members/images/default-avatar.jpg" %}"
+                        <img src="{% static 'members/images/default-avatar.jpg' %}"
                              alt="{{ member.profile.display_name }}"/>
                     {% else %}
                         <img src="{{ member.profile.photo.url }}"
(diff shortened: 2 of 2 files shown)


website/members/models/member.py
```diff
@@ -117,6 +117,17 @@ def latest_membership(self):
             return None
         return self.membership_set.latest("since")
 
+    @property
+    def oldest_membership_of_current_type(self):
```
se-bastiaan, line 121: I'd rather see this as part of the membership so that you can get that first connected membership of the same type from any membership, not just the latest one.

Jorritboer: I would say it's alike ```current_membership```, ```latest_membership```, ```earliest_membership```, which are all in Member

Jorritboer (approved):

website/members/models/member.py
se-bastiaan: It really is not, the filter is on the membership that you already have, not based on the user like those are. In all those cases you do not have a membership yet. You want to know 'the latest membership of the user' or the 'current membership of the user'.

Here you want "the first membership of the current membership's type"

website/members/models/member.py
```diff
@@ -117,6 +117,17 @@ def latest_membership(self):
...
+        """Get the first membership of the most recent membership's type."""
+        if not self.membership_set.exists():
+            return None
+        return (
+            self.membership_set.filter(type=self.latest_membership.type)
```
se-bastiaan, line 126: What happens when a user is a member from 2014-09-01 until 2015-08-31, then a benefactor from 2015-09-01 until 2018-08-31, and then a member again starting 2018-09-01?

I don't think it should show `Member (since 1 Sept 2014)` in that case. Since that is not true. And I feel that this code would do that.

mark-boute: It would display `Member (since 1 Sept 2014)` indeed.
I will look into placing this in the membership, and I'll probably change it to be consecutive memberships of the same type.

se-bastiaan: https://github.com/svthalia/concrexit/blob/master/website/activemembers/models.py#L234

PR #1953, opened 2021-10-01
Fix showing inactive partners

### LINKED ISSUE #1910: Inactive partner page is still visible
When a partner is not set to be active, the page can still be loaded, while it should give a 404. See for example: https://thalia.nu/career/partners/exellys/, can be viewed, but it cannot be found on the https://thalia.nu/career/ page.

### How to reproduce
1. Go to specific partner settings and disable "Is active" checkbox
2. Go to partner page: the partner is not visible
3. Go to the specific partner page: the partner is visible

### Expected behaviour
The specific partner page should give a 404.

DeD1rk:
Closes #1910.

Also adds a missing uniqueness constraint on the partners' slugs.

### How to test
1. Create an inactive partner.
2. Go to `career/partners/<slug>/` for that partner.
3. Get a 404.
4. Try to set the same slug for two partners.
5. Fail.


wdoeland (approved):
Looks good, thanks!

PR #1969, opened 2021-10-04
Add hash to payment processing

### LINKED ISSUE #1921: Payable payment amount can change between loading payment processing page and actually paying
The following is possible

1. Open the payment process view, trying to pay a payable
2. The page will show ("you are about to pay €15")
3. Meanwhile, the payable is secretly updated, for example by an update of the thadmin
4. You hit 'confirm'
5. You actually pay a different amount than indicated

Two solutions:

1. Lock a payable when the page is loaded. 
2. [My preferred solution]: pass the payment amount (and description) as fields in the payment processing page form. After confirmation, verify that the amount still matches, and display an error otherwise
   
### Motivation
Users should know what they pay

JobDoesburg:
Closes #1921 

### Summary
Adds a hash of a payable to the payment processing flow. When a payable changes meanwhile, the hash of the payable will also change and thus the payment processing will fail.

### How to test
1. Have a sales order
2. Try to pay for it
3. Meanwhile, change the amount of the order.
4. Confirming the payment will result in an error


se-bastiaan (approved):
LGTM

PR #1974, opened 2021-10-07
Add vacancies to the frontpage

### LINKED ISSUE #1959: Show vacancies more prominently
The vacancy page is not hard to find, but you do have to look for it. 

It would be nice if the vacancies had a bit more of a prominent place on the site so users don't need to search for it. There are multiple ways to do this, this is what I came up with:

- Show some vacancies at the bottom of the home page, under the upcoming events.
- Move the vacancies link in the navbar out of the career section and onto the main nav bar.

### Motivation
It is important that vacancies are seen by users.

See #1845 

### Additional context
See #1845

se-bastiaan:
Closes #1959 

### Summary
This PR adds vacancies to the frontpage by  changing the way we show our events on it as well.
Screenshot 2021-10-06 at 22-47-42 Study Association Thalia
Screenshot 2021-10-06 at 22-30-49 Study Association Thalia

### How to test
1. Make sure you have some events and vacancies in your instance
2. Open the frontpage
3. Check that external vacancies go to a website and that internal vacancies to the right description

### DIFF
--- a/website/partners/templates/partners/frontpage_vacancies.html
+++ b/website/partners/templates/partners/frontpage_vacancies.html
@@ -21,8 +21,8 @@
     <p class="text-center mt-3">
         {% url 'partners:vacancies' as vacancies_index %}
         {% blocktrans trimmed %}
-            <a href="{{ vacancies_index }}">Check out the other vacancies</a>
-            if this does not match your interests.
+            <a href="{{ vacancies_index }}">Check out the vacancies page</a>
+            for more.
         {% endblocktrans %}
     </p>
 </div>


wdoeland:
Looks very nice!

se-bastiaan:
> 
> 
> Imo events are more important than vacancies, so I don't like them being shown side by side

Even tho I share the sentiment that they are not equally important I do think that we do not lose any information about events in this layout. The excerpt usually doesn't contain much useful information.

Moreover, this works a lot better on mobile since it requires less space.

PR #1975, opened 2021-10-10
Enable OIDC in OAuth settings

### LINKED ISSUE #1705: Enable OIDC for OAuth
I want to be able to use OpenID Connect with our OAuth applications and I currently cannot do that.

The toolkit we use has support but needs some config: https://django-oauth-toolkit.readthedocs.io/en/latest/oidc.html

### Motivation
This provides a standardised login flow for clients.

Using normal OAuth.

se-bastiaan:
Closes #1705 

### Summary
Enable OIDC in OAuth settings. This requires some new settings in production to be able to use the RSA signature mode!

### How to test
1. Check that `/user/oauth/.well-known/jwks.json` responds with the set RSA key

### DIFF
--- a/website/thaliawebsite/urls.py
+++ b/website/thaliawebsite/urls.py
@@ -34,8 +34,14 @@ from django.contrib.auth.views import LoginView
 from django.contrib.sitemaps.views import sitemap
 from django.urls import path, re_path
 from django.views.i18n import JavaScriptCatalog
-from oauth2_provider.urls import base_urlpatterns, oidc_urlpatterns
-from oauth2_provider.views import AuthorizedTokensListView, AuthorizedTokenDeleteView
+from oauth2_provider.urls import base_urlpatterns
+from oauth2_provider.views import (
+    AuthorizedTokensListView,
+    AuthorizedTokenDeleteView,
+    ConnectDiscoveryInfoView,
+    JwksInfoView,
+    UserInfoView,
+)
 
 from activemembers.sitemaps import sitemap as activemembers_sitemap
 from documents.sitemaps import sitemap as documents_sitemap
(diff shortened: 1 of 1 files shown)


se-bastiaan:
Any idea how to fix the build of the packages here? @pingiun 
Seems rust compiler is missing?

se-bastiaan:
> 
> 
> > Any idea how to fix the build of the packages here? @pingiun Seems rust compiler is missing?
> 
> We can use the [nixpkgs version](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/python-modules/cryptography/default.nix#L89) of the `cryptography` package, which fixes this by fixing the rust compiler stuff. This does mean that we'll be managing this dependency version via nix instead of poetry

How do we do that?

pingiun:
> > > Any idea how to fix the build of the packages here? @pingiun Seems rust compiler is missing?
> > 
> > 
> > We can use the [nixpkgs version](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/python-modules/cryptography/default.nix#L89) of the `cryptography` package, which fixes this by fixing the rust compiler stuff. This does mean that we'll be managing this dependency version via nix instead of poetry
> 
> How do we do that?

```diff
diff --git a/nix/concrexit.nix b/nix/concrexit.nix
index c4a4f52d..7f4d8a28 100644
--- a/nix/concrexit.nix
+++ b/nix/concrexit.nix
@@ -16,6 +16,7 @@ let
       _self: _super: {
         pillow = pkgs.python3.pkgs.pillow;
         python-magic = pkgs.python3.pkgs.python_magic;
+        cryptography = pkgs.python3.pkgs.cryptography;
         # We don't install uswgi from pypi but instead use the nixpkgs version
         uwsgi = { };
       }
```

PR #1976, opened 2021-10-10
Run test and nix workflows only on master push

### LINKED ISSUE #1967: Do not run CI both on push and on pull_request
### Describe the change
Our CI currently is triggered on both `push` and `pull_request` events. However, this results in CI running twice.

### Motivation
Faster CI, better for the environment etc.

### Current implementation
Two triggers

### Suggested implementation
Change our workflow

### Additional context
afbeelding

se-bastiaan:
Closes #1967

### Summary
Run test and nix workflows only on master push

### How to test
1. See that no push workflow was ran for this PR

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -2,7 +2,7 @@ name: "Build with Nix and deploy if master"
 on:
   push:
     branches:
-      - main
+      - master
   pull_request:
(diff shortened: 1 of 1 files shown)


pingiun:
What's the rationale for this? Why only changes it for the nix action?

PR #1979, opened 2021-10-11
optimize event list queries

ColonelPhantom:
Currently, the calendar is quite slow, since for each event we query twice to check if the logged in user is registered for that particular event. This is obviously really efficient, since it's 2n+1 queries. This PR aims to solve this.

So far I have annotated the registrations; which means that while there are extra queries to check for cancellation it's already a lot more efficient, since events for which the current user is not registered do not require extra queries anymore.

### Summary
Add an annotation to check if the current user is registered

### How to test
1. Go to Calendar
2. Check if all events and registrations appear properly

### DIFF
--- a/website/events/api/calendarjs/views.py
+++ b/website/events/api/calendarjs/views.py
@@ -35,7 +35,7 @@ class CalendarJSEventListView(ListAPIView):
         )
         if self.request.member:
             events = events.annotate(
-                user_reg=Subquery(
+                member_registration=Subquery(
                     EventRegistration.objects.filter(
                         event=OuterRef("id"), member=self.request.member
                     ).values("member")
--- a/website/events/api/calendarjs/views.py
+++ b/website/events/api/calendarjs/views.py
@@ -26,20 +26,18 @@ class CalendarJSEventListView(ListAPIView):
 
     def get_queryset(self):
         start, end = extract_date_range(self.request)
+        objects = Event.objects.filter(
+            end__gte=start, start__lte=end, published=True
+        ).annotate(
+            number_regs=Count('eventregistration', filter=Q(eventregistration__date_cancelled=None))
+        )
         if self.request.member:
-            objects = Event.objects.filter(
-                end__gte=start, start__lte=end, published=True
-            ).annotate(
+            objects = objects.annotate(
                 user_reg=Subquery(
                     EventRegistration.objects.filter(
                         event=OuterRef("id"), member=self.request.member
                     ).values("member")
                 )
-            )
-            # .annotate(user_reg=Count('eventregistration', filter=Q(member=self.request.member)))
-        else:
-            objects = Event.objects.filter(
-                end__gte=start, start__lte=end, published=True
             )
         return objects
 
--- a/website/events/api/calendarjs/views.py
+++ b/website/events/api/calendarjs/views.py
@@ -27,13 +27,20 @@ class CalendarJSEventListView(ListAPIView):
     def get_queryset(self):
         start, end = extract_date_range(self.request)
         if self.request.member:
-            objects = Event.objects.filter(end__gte=start, start__lte=end, published=True)\
-                .annotate(user_reg=Subquery(EventRegistration.objects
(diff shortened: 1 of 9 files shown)


website/events/api/calendarjs/views.py
```diff
@@ -24,7 +26,22 @@ def get_serializer_context(self):
 
     def get_queryset(self):
         start, end = extract_date_range(self.request)
```
se-bastiaan, line 28: Maybe it is possible to use query properties for this? Thats what job used for sales.

https://pypi.org/project/django-queryable-properties/

ColonelPhantom: I could try this for checking the number of registrations, but user_reg is based on services.is_user_registered, so I do not think I can turn that into a property.

ColonelPhantom: For number of registrations, this does not work since looking up the property still does a separate query to count the number of participants.

website/events/api/v1/serializers/events/retrieve.py
```diff
@@ -58,12 +58,10 @@ def _description(self, instance):
         return strip_spaces_between_tags(bleach(instance.description))
 
     def _num_participants(self, instance):
-        if (
-            instance.max_participants
-            and instance.participants.count() > instance.max_participants
-        ):
+        p_count = instance.participants.count()
```
se-bastiaan, line 61:
```suggestion
        participant_count = instance.participants.count()
```

Easier to understand if the names are descriptive.

website/events/api/v2/serializers/event.py
```diff
@@ -78,12 +78,10 @@ def _user_registration(self, instance):
         return None
 
     def _num_participants(self, instance):
-        if (
-            instance.max_participants
-            and instance.participants.count() > instance.max_participants
-        ):
+        p_count = instance.participants.count()
```
se-bastiaan, line 81:
```suggestion
        participant_count = instance.participants.count()
```

website/events/models/event_registration.py
```diff
@@ -63,7 +64,8 @@ def is_registered(self):
     def queue_position(self):
         if self.event.max_participants is not None:
             try:
-                return list(self.event.queue).index(self) + 1
+                queue_ids = [r.member_id for r in self.event.queue]
```
se-bastiaan, line 67: What is `r`?

se-bastiaan: Oh and I kinda fail to see the usefulness of this change :p

website/events/api/calendarjs/views.py
```diff
@@ -24,7 +26,22 @@ def get_serializer_context(self):
...
-        return Event.objects.filter(end__gte=start, start__lte=end, published=True)
+        objects = Event.objects.filter(
```
se-bastiaan, line 29:
```suggestion
        events = Event.objects.filter(
```

website/events/api/calendarjs/views.py
```diff
@@ -24,7 +26,22 @@ def get_serializer_context(self):
...
+            end__gte=start, start__lte=end, published=True
+        ).annotate(
+            number_regs=Count(
+                "eventregistration", filter=Q(eventregistration__date_cancelled=None)
+            )
+        )
+        if self.request.member:
+            objects = objects.annotate(
+                user_reg=Subquery(
```
se-bastiaan, line 38:
```suggestion
                member_registration=Subquery(
```

se-bastiaan (changes requested):
It's not bad to use longer variables names :)

website/events/models/event_registration.py
ColonelPhantom: Before this change, looking up the queue position required looking up the member. This happens because I think str is used, according to the stacktrace. And this requires fetching the member's full name. Just using their id's prevents this, since these are stored inside the registration. (I do assume the member_ids are unique in the queue, which always is true since cancelled registrations are not in there).

se-bastiaan (approved):

PR #2011, opened 2021-11-01
Use API v2 for pizzas  admin

### LINKED ISSUE #1948: Use API v2 for pizza orders
For some reason (#1931) the payment widget of the order overview page is broken. 

Instead of debugging the existing code, I prefer switching API v2 immedaitely.

### Motivation
Better

Debugging

JobDoesburg:
Closes #1948 

### Summary
Changes the javascript code in the custom Food Event Orders Admin to use API v2.

### How to test
1. Have a pizza event
2. Add orders to it
3. Try marking the orders as paid (and unpaid again, change the order, try to delete the payment for an order that was paid a long time ago, try it when the event is way back in the past and another food event is currently active)
4. Verify that API v2 is used and that this all actually works

### DIFF
--- a/website/pizzas/static/pizzas/js/admin.js
+++ b/website/pizzas/static/pizzas/js/admin.js
@@ -1,18 +1,5 @@
 django.jQuery(function () {
     var $ = django.jQuery;
-
-    $(".present-check").change(function () {
-        var checkbox = $(this);
-        var url = checkbox.parent().parent().data("url");
-        var checked = checkbox.prop('checked');
-        request('PATCH', url, { present: checked }, function(result) {
-            checkbox.prop('checked', result.present);
-            $("table").trigger("update");
-        }, function() {
-            checkbox.prop('checked', !checked);
-        });
-    });
-
     var payment_previous;
     $('select[name=payment]').on('focus', function () {
         payment_previous = this.value;
@@ -24,7 +11,6 @@ django.jQuery(function () {
         if (payment_previous === $(this).val()) {
             return;
         }
-        $('table').trigger('update', false).trigger('sortReset');
 
         var successCallback = function(data) {
             if (!data) {
@@ -33,12 +19,10 @@ django.jQuery(function () {
                 select.addClass('paid');
             }
             select.val(data ? data.payment.type: none);
-            $('table').trigger('update', false);
         };
 
         var failCallback = function(xhr) {
             select.val(payment_previous);
-            $('table').trigger('update', false);
 
             if (payment_previous === none) {
                 select.removeClass('paid');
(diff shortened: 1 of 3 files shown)


DeD1rk:
api:v2:admin:food:food-event-order-detail doesn't exist: 
In website/pizzas/api/v2/admin/urls.py:
```
    path(
        "food/events/<int:event_id>/orders/<int:pk>/",
        FoodOrderAdminDetailView.as_view(),
        name="event-registration-detail",
    ),
```
So yeah the name that does exist is just very wrong

website/pizzas/static/pizzas/js/admin.js
```diff
@@ -1,112 +1,60 @@
...
+        if ($(this).val() === none) {
+            $(this).removeClass('paid');
+            request('DELETE', url, null, successCallback, failCallback);
+        } else {
+            $(this).addClass('paid');
+            request('PATCH', url, { payment_type: $(this).val() }, successCallback, failCallback);
         }
-    });
 
-    $('#searchbar').on('input', function() {
```
se-bastiaan, line 92: Where did this go?

se-bastiaan (changes requested):

website/pizzas/static/pizzas/js/admin.js
JobDoesburg: woops, went a bit too hard on the removing code

se-bastiaan: Yeah and now added too much

website/pizzas/static/pizzas/js/admin.js
```diff
@@ -1,100 +1,81 @@
...
-    });
-
     var payment_previous;
     $('select[name=payment]').on('focus', function () {
         payment_previous = this.value;
-    }).change(function() {
+    }).change(function () {
         var select = $(this);
-        var id = $(this).parents('tr').data('id');
+        var url = $(this).parents('tr').data("payable-url");
```
se-bastiaan, line 8:
```suggestion
        var url = $(this).parents('tr').data('payable-url');
```

website/pizzas/static/pizzas/js/admin.js
```diff
@@ -109,4 +90,47 @@ django.jQuery(function() {
...
+            return false;
+        },
+        format: function(s, t, node) {
+            return $(node).data("sortval");
+        },
+        type: "text"
+    });
+
+    $("table").tablesorter({
+        sortList: [[1,0]],
```
se-bastiaan, line 120: Why the sortList?

website/pizzas/static/pizzas/js/admin.js
```diff
@@ -1,100 +1,81 @@
-django.jQuery(function() {
+django.jQuery(function () {
     var $ = django.jQuery;
-    var csrfToken = $('#content-main').data('csrf');
```
se-bastiaan, line 3: If you no longer use this value I expect it to be removed from the HTML

se-bastiaan:
Idea overall is good but I feel like this is leaves traces in the codebase of things we longer need

website/pizzas/static/pizzas/js/admin.js
JobDoesburg: I just copied it from what was there earlier. It states that the table should be sorted on the second column (last name) in ascending order

se-bastiaan: I know what it does haha, but it is added in this PR. It is in the events admin.js but not in the original one for the pizza admin.

se-bastiaan: Deliberate choice not to change this back?

JobDoesburg: I figured it would be better not to change behaviour (sorting on first name is still possible by clicking it)

pingiun (approved):

PeaceDucko (approved):

Issue #2013, opened 2021-11-01
Update implementation of cryptography dependency when crates are built


se-bastiaan:
### Describe the change

TODO: change https://github.com/svthalia/concrexit/blob/master/nix/concrexit.nix#L21 when https://github.com/nix-community/poetry2nix/issues/413 is fixed

### Current implementation
@svthalia/server Please elaborate.

### Suggested implementation
@svthalia/server Please elaborate.

### Additional context
When updating cryptography they added rust build requirements and thus our build started failing when we updated to 35.0.0

pingiun:
> @svthalia/server Please elaborate.

I expect this issue to be solved in poetry2nix at some point in the future and at that point we can remove our override

PR #2014, opened 2021-11-01
Remove todo from code

se-bastiaan:
### Summary
See https://github.com/svthalia/concrexit/issues/2013


pingiun (approved):
I don't think it's bad to have it in the code and in our issues but sure

PR #2021, opened 2021-11-03
Use start date for registration info text in calendar

### LINKED ISSUE #2008: Registration message in calendar wrong
### Describe the bug
In the calendar, when you have to register for an event, it says in the block; You can register on .......
However, this seems to be the date of when the event will take place (or the registration deadline, can't tell for sure).
See screenshots for an example.

### How to reproduce
1. Go to the calendar and look for an event with a registration.
2. Look at the message in the calendar view
3. Compare with the registration date in the event itself.

### Expected behaviour
The message in the calendar view matches the date of when the registrations will open.

### Screenshots
image
image

se-bastiaan:
Closes #2008 

### Summary
This PR changes the messages shown when the registration for an event are not open yet and when registration is optional.

image
image

### How to test
1. Create events with optional registrations and with registrations that open in the future within 2 days and one more in the future.
2. Go to the calendar view.
3. Check the text shown for each calendar item.

### DIFF
--- a/website/events/api/calendarjs/serializers.py
+++ b/website/events/api/calendarjs/serializers.py
@@ -50,7 +50,7 @@ class EventsCalenderJSSerializer(CalenderJSSerializer):
                 return _("You are registered for this event")
         # Optional registration possible
         elif instance.optional_registration_allowed:
-            return _("You can optionally register for this event")
+            return _("Registering for this event is optional")
         # No places left
         elif instance.reached_participants_limit():
             return _("You can put yourself on the waiting list for this event")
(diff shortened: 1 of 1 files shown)


ColonelPhantom:
One idea, do we maybe also want to add e.g. registration closing times or number of remaining spots?

ColonelPhantom (approved):
Changes look good!

website/events/api/calendarjs/serializers.py
```diff
@@ -50,7 +50,7 @@ def _registration_info(self, instance: Event):
                 return _("You are registered for this event")
         # Optional registration possible
         elif instance.optional_registration_allowed:
-            return _("You can optionally register for this event")
+            return _("Registering for this event is optional")
```
pingiun, line 53: The other messages begin with "You ...." maybe it's good to keep this, or change all of the message if you think this copy is better

se-bastiaan: I definitely thought the 'you can optionally register' was weird haha, that's why I changed it. Not really a problem if it is not the same right?

PR #2037, opened 2021-11-17
Add terraform configs for staging and production

### LINKED ISSUE #1561: Switch production to NixOS
### Describe the change
Currently we are running the production version of Concrexit on a Ubuntu server using docker compose and ansible. Our staging environment has been running through Nix(OS) for the past few months and we believe it is stable. We want to switch our production server to running through NixOS. We will not be running green/blue deployments at this time. 

This issue describes what should still be done:

### Checklist before switching
- [x] Fix #1556 
- [ ] create a new server for production on AWS and set up NixOS with our config (via terraform?)
- [ ] Communicate to members about downtime
- [x] Create EBS disks (psql, media)
- [x] Create Terraform scripts

### Checklist while switching
- [ ] take the current server offline by stopping nginx
- [ ] make a backup of all data
- [ ] migrate database
- [ ] migrate media
- [ ] point the elastic IP to the new EC2 production instance
- [ ] test the new server extensively

### Checklist after switching
- [ ] terminate the old server (I think it'd be nice if we still kept it alive for a few weeks)

pingiun:
Closes #1561 

### Summary

Creates an environment where production can run on NixOS

### What's done

Staging should now be able to run with this code. The deployment I created from this PR can be viewed on staging-tf.thalia.nu

My plan is to migrate the data from the current staging to this instance soon after this PR is merged. Then I will create a new PR to use the staging domain instead of staging-tf (terraform)


se-bastiaan:
This is so awesome

infra/modules/concrexit/main.tf
```diff
@@ -0,0 +1,250 @@
...
+  public_key = chomp(tls_private_key.ssh.public_key_openssh)
+}
+
+resource "tls_private_key" "ssh" {
+  algorithm = "RSA"
+  rsa_bits  = "4096"
+}
+
+resource "local_file" "private_key" {
+  filename = "id_rsa"
```
se-bastiaan, line 206: This is a file right?  Does this mean it will recreate the EC2 instance when you run terraform with a changed key?

pingiun: The key is stored in the terraform state, the file is written out to make it possible to do a nix deployment afterwards

se-bastiaan: aah thats the `resource "tls_private_key" "ssh"`, have to make sure that S3 bucket is very not public then :p

pingiun: yea i'm not totally happy with it and we should probably deal with secrets better

infra/stages/production/.terraform.lock.hcl
```diff
@@ -0,0 +1,93 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/aws" {
```
se-bastiaan, line 4: I don't think this should be in git

se-bastiaan (approved):
I think this looks good. Haven't tested obviously.

pingiun:
> obviously.

time to install nix

.github/workflows/nix.yaml
```diff
@@ -12,23 +12,23 @@ jobs:
...
-    - run: nix-build -j 4 --arg version "\"$(git rev-parse HEAD)\"" -A github-actions release.nix
-    - run: nix-shell --run "echo OK"
+        name: svthalia
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN_SVTHALIA }}'
+    - run: nix build --log-format raw --verbose
     - name: Deploy if this is master or staging
       if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/staging'
+      env:
+        TF_VAR_ssh_private_key: ${{ secrets.STAGING_SSH_PRIVATE_KEY }}
+        TF_VAR_ssh_public_key: "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINgNaleedXX3iDZu8KlafjjURMVwXO+YZU1498tWu2m8 deploykey"
```
JobDoesburg, line 22: Not sure what this is, but I think it would be nice to centralize the public key too in secrets?

JobDoesburg:
I really don't know a lot about terraform and nix...

.github/workflows/nix.yaml
pingiun: yea that's fair

default.nix
```diff
@@ -0,0 +1,117 @@
...
+  concrexit-env = poetryPython.python.buildEnv.override { extraLibs = poetryPackages; };
+
+  # We want to enable the python3 plugin for uwsgi, so we override here
+  uwsgi-python = (uwsgi.override { plugins = [ "python3" ]; python3 = concrexit-python; }).overrideAttrs (old: { buildInputs = old.buildInputs ++ [ pkgs.zlib pkgs.expat ]; });
+
+  # Wrapper script that sets the right options for uWSGI
+  concrexit-uwsgi = writeScriptBin "concrexit-uwsgi" ''
+    #! ${pkgs.runtimeShell}
+    cd ${concrexit}/website
+    MANAGE_PY=1 ${concrexit-env}/bin/python ${concrexit}/website/manage.py migrate
```
JobDoesburg, line 55: does this maybe also require `manage.py collectstatic` and `manage.py compress`

pingiun: This is done during built time right here: https://github.com/svthalia/concrexit/blob/0e922a106970c6d6a242979a70554a5b6676b594/default.nix#L89-L90

infra/stages/staging/variables.tf
```diff
@@ -0,0 +1,40 @@
...
+
+variable "ssh_private_key" {
+  description = "The private key used to create the EC2 instance and connect to, public key is used in the Nix OS configuration"
+  type        = string
+}
+
+variable "ssh_public_key" {
+  description = "The public key used to create the EC2 instance, also used in the Nix OS configuration"
+  type        = string
+  default     = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINgNaleedXX3iDZu8KlafjjURMVwXO+YZU1498tWu2m8 deploykey"
```
JobDoesburg, line 39: idem, maybe this should be in the secrets

PR #2046, opened 2021-12-01
Add support for multiple local partners

### LINKED ISSUE #2041: Add support for multiple local partners
We really want to be able to have multiple local partners at Thalia. And the website current does not support this.

We would like the ability to add multiple local partners.

### Motivation
We think it's useful to have multiple local partners so we can show our students what is available in Nijmegen.

Only having one local partner, which does not make sense in the current partner climate.

### Additional context
The local partner is also shown in the newsletter.

wdoeland:
See #2041 

### Summary
Add support for multiple local partners instead of just one.

### How to test
1. Go to 'site admin -> partners'
2. Make sure two or more partners are set as local partner
3. Go to the partner page on the website
4. See two local partners on the website
7. Check if steps 3-4 work with one local partner

### DIFF
--- a/website/partners/models.py
+++ b/website/partners/models.py
@@ -54,8 +54,6 @@ class Partner(models.Model):
         if self.is_main_partner:
             self.is_local_partner = False
             self._reset_main_partner()
-        if self.is_local_partner:
-            self._reset_local_partner()
 
         super().save(**kwargs)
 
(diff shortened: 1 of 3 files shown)


pingiun:
deploy review

pingiun:
remove review

se-bastiaan (approved):

se-bastiaan:
The logos can become very large in usage, maybe we want to make those a bit smaller? But it works, can improve later anyways.

wdoeland:
> The logos can become very large in usage, maybe we want to make those a bit smaller? But it works, can improve later anyways.

We can fix that later, I think.

> Maybe don't close the issue yet tho?

Fixed that!

PR #2051, opened 2021-12-04
Support individual prices for event registrations

JobDoesburg:
### Summary
In some occasions, people attending an event will pay different registration fees then others. For example, at Thalia bouldering, certain members brought their own climbing gear, which resulted in a discount. Currently, concrexit does not support this. 

This PR adds a special_price field to event registrations, which if set, will override the default price for the event. Moreover, it adds a remarks field that can be used to register any specialities or explanation as text remarks.

### How to test
1. Go to the event admin for an event
2. Have a registration
3. Click edit
4. Set a special_price
5. The event admin will show that this person is paying a different amount, and creating a payment will use the different amount too


DeD1rk:
I haven't tried but it looks like this will give every registration a default special price of 0/0.0, not None. I'd say €0 is a valid special price, so the field should be nullable and have None as default right?..

DeD1rk:
And something to also consider is that people could Thalia Pay before they actually get the special price. In that case, should the payment be deleted?

JobDoesburg:
> I haven't tried but it looks like this will give every registration a default special price of 0/0.0, not None. I'd say €0 is a valid special price, so the field should be nullable and have None as default right?..

You are right

website/events/payables.py
```diff
@@ -8,7 +8,11 @@
 class EventRegistrationPayable(Payable):
     @property
     def payment_amount(self):
-        return self.model.event.price
+        return (
+            self.model.event.price
+            if not self.model.special_price
+            else self.model.special_price
+        )
```
se-bastiaan, line 15: I feel like this logic should be part of the event registration, give it a property that returns the right price for that specific registration so that you can use it in all places and not have to repeat the logic.

website/events/templates/events/admin/registrations_table.html
```diff
@@ -51,15 +52,15 @@
                             data-none="no_payment"
                             {% if registration.payment %}class="paid"{% endif %}>
                         <option value="no_payment"
-                                {% if not registration.payment %}selected{% endif %}>{% trans 'None' %}</option>
+                                {% if not registration.payment %}selected{% endif %}>{% trans 'None' %} {% if registration.special_price %}(€ {{ registration.special_price }}) {% endif %}</option>
```
se-bastiaan, line 55: Instead of repeating this price, can we put it in the same td but not in the select? This code is so ugly already :S

website/events/templates/events/admin/registrations_table.html
```diff
@@ -68,6 +69,7 @@
                 {% if cancellations %}
                 <td data-sortval="{{ registration.date_cancelled|date:'c' }}">{{ registration.date_cancelled }}</td>
                 {% endif %}
+                <td data-sortval="{{ registration.remarks }}">{% if registration.remarks %}{{ registration.remarks }}{% endif %}</td>
```
se-bastiaan, line 72: I think you can use a filter that automatically sets the output to '-' or something

se-bastiaan: https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#default-if-none

That doesn't require the if statement

se-bastiaan:
> And something to also consider is that people could Thalia Pay before they actually get the special price. In that case, should the payment be deleted?

When there is a payment already it will not accept changing of the 'special price'. So the payment has to be set to 'none' by the organiser before the price can be changed. Which imo is not really a problem.

Only thing I have to add is that 'special price' sounds like such a Dutch translation. :p

DeD1rk:
> When there is a payment already it will not accept changing of the 'special price'. So the payment has to be set to 'none'

That seems perfect.

website/events/templates/events/admin/registrations_table.html
JobDoesburg: I tried exactly that first, but it looked very ugly visually

JobDoesburg:
> Only thing I have to add is that 'special price' sounds like such a Dutch translation. :p

I agree, but I don't have a better suggestion myself that is short and captures the meaning

website/events/models/event_registration.py
```diff
@@ -103,20 +115,40 @@ def is_late_cancellation(self):
...
                 }
             )
+        if (
+            self.payment
+            and self.special_price
+            and self.special_price != self.payment.amount
+        ):
+            errors.update(
+                {
+                    "special_prize": _(
```
DeD1rk, line 144:
```suggestion
                    "special_price": _(
```

DeD1rk:
Looks good, the api also only adds fields, so we can support the change from the app easily without anything breaking. Is the logic for this also checked when writing from the admin api (disallowing setting a special price when already paid, I'm not sure exactly what was going on with not running the clean method in the api)?

JobDoesburg:
> Looks good, the api also only adds fields, so we can support the change from the app easily without anything breaking. Is the logic for this also checked when writing from the admin api (disallowing setting a special price when already paid (and the permission), I'm not sure exactly what was going on with not running the clean method in the api)?

You are very right, I just checked (https://github.com/encode/django-rest-framework/issues/3144, https://www.django-rest-framework.org/api-guide/validators/#validation-in-rest-framework), and indeed DRF does **NOT** run a full clean before saving the model. In their opinion, that should be a pre-save hook. I think we should fix this globally across our whole code base

DeD1rk:
> I think we should fix this globally across our whole code base

Yes, how do we do that though? migrate all clean()s to pre-save? Or somehow modify the rest framework or all api views?

JobDoesburg:
There are some possibilities:
- change each model's save function to look like this:
``` 
def save(self, **kwargs):
        self.full_clean()
        (...)
```
This will work, but is not that efficient because we're doing double cleaning.

- Do the same thing, but with a global `pre_save` signal. This has the benefit that it can be done with less code, but it is still not really efficient, and it could possibly interfere with other `pre_save` signals.
- Create our own Serializer that does the following:
```
    def validate(self, attrs):
        instance = self.meta.model(**attrs)
        instance.clean()
        return attrs
```
- Fix it in all the Create/Update API views in a similar way
- Maybe other ideas??

DeD1rk:
> Create our own Serializer that does the following:
> ```python
>   def validate(self, attrs):
>       instance = self.meta.model(**attrs)
>       instance.clean()
>       return attrs
```

That seems like a practical solution. We can indeed subclass rest_framework's ModelSerializer, and replace that in all of the serialisers we use.

JobDoesburg:
See #2075

se-bastiaan (approved):

PR #2052, opened 2021-12-04
Add summaries to frontpage, re-show event cards

JobDoesburg:
### Summary
In #1974, the way events were displayed on the front page changed. We did not deploy this, because we did not like the new design, as it did not show the first lines of the description anymore.
This PR fixes that the event cards are displayed 'as normal' again. Moreover, it adds the latest course summaries to the front page, next to the vacancies.

### How to test
1. Visit the home page

### DIFF
--- a/website/education/templates/education/frontpage_summaries.html
+++ b/website/education/templates/education/frontpage_summaries.html
@@ -1,5 +1,5 @@
 {% load i18n bleach_tags alert %}
-<div class="frontpage-vacancies">
+<div class="frontpage-summaries">
     <h1 class="text-center section-title">{% trans "Recent course summaries" %}</h1>
 
     {% if not summaries %}
--- a/website/education/templates/education/frontpage_summaries.html
+++ b/website/education/templates/education/frontpage_summaries.html
@@ -12,8 +12,8 @@
             {% for summary in summaries %}
                 <a class="list-group-item py-3"
                    href="{{ summary.course.get_absolute_url }}">
-                    <span class="vacancy-title mr-3">{{ summary.course.name }}</span>
-                    <span class="vacancy-company" title="{{ summary.name }}">{{ summary.name }}</span>
+                    <span class="summary-course-title mr-3">{{ summary.course.name }}</span>
+                    <span class="summary-name" title="{{ summary.name }}">{{ summary.name }}</span>
                 </a>
             {% endfor %}
         </div>
--- a/website/education/templates/education/frontpage_summaries.html
+++ b/website/education/templates/education/frontpage_summaries.html
@@ -1,0 +1,28 @@
+{% load i18n bleach_tags alert %}
+<div class="frontpage-vacancies">
+    <h1 class="text-center section-title">{% trans "Recent course summaries" %}</h1>
+
+    {% if not summaries %}
+        <div class="mt-4">
+            {% trans 'There are currently no course summaries' as info_text %}
+            {% alert 'info' info_text dismissible=False %}
+        </div>
+    {% else %}
+        <div class="list-group">
+            {% for summary in summaries %}
+                <a class="list-group-item py-3"
+                   href="{{ summary.course.get_absolute_url }}">
+                    <span class="vacancy-title mr-3">{{ summary.course.name }}</span>
+                    <span class="vacancy-company" title="{{ summary.name }}">{{ summary.name }}</span>
+                </a>
+            {% endfor %}
+        </div>
+    {% endif %}
+    <p class="text-center mt-3">
+        {% url 'education:courses' as courses_index %}
+        {% blocktrans trimmed %}
+            <a href="{{ courses_index }}">Check out the courses page</a>
+            for more.
+        {% endblocktrans %}
+    </p>
+</div>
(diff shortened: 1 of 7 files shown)


se-bastiaan:
I know have not really been able to visit Technicie evenings, but I still feel kinda sad that you didn't tell me that you didn't like it and just decided to leave it out instead of just messaging me 'hey sebas, we decided to leave it out and we opened a reverting MR'. Now I had to find that out by asking. Even after 3 people approved my changes.

There were reasons I decided to leave out the descriptions:
1. The first couple of characters usually don't provide more context that the title of the event itself. Usually they repeat the title of the event in the first line even. To illustrate, the current homepage:
Screenshot 2021-12-05 at 16 01 36
2. The mobile version of these cards is quite bad for the user experience I think. The events are already below the fold  and to reach the last one you have to scroll quite the amount. Adding the vacancies below the super large cards means that nobody will ever scroll so far down to reach them, making them basically invisible. Which is also true for this version of the desktop version you want to introduce.

<details>
  <summary>Mobile screenshot</summary>

Screenshot 2021-12-05 at 16-05-00 Study Association Thalia

</details>

<details>
  <summary>Desktop screenshot</summary>

Screenshot 2021-12-05 at 16-16-12 Study Association Thalia

</details>

se-bastiaan:
Completely unrelated to the previous point: how many summaries are actually posted? Maybe the other education documents are also useful to make sure the widget doesn't stay the same for a long time.

website/thaliawebsite/templates/index.html
```diff
@@ -26,11 +26,13 @@ <h1 class="text-center section-title">{% trans "Study Association Thalia" %}</h1
 
     {% render_frontpage_articles %}
 
+    {% render_event_cards %}
+
     <section class="page-section">
         <div class="container">
             <div class="row">
                 <div class="col-12 col-md-6">
-                    {% render_frontpage_events %}
```
se-bastiaan, line 33: Are you keeping this tag on purpose? What about the alumni page?

website/events/templates/events/event_cards.html
```diff
@@ -0,0 +1,48 @@
...
+                                     data-placement="top" title="{% trans "Registered for this event" %}">
+                                        <div class="has-registration"></div>
+                                    </div>
+                                {% else %}
+                                    <div class="event-indication" data-toggle="tooltip"
+                                     data-placement="top" title="{%  trans "Not registered for this event" %}">
+                                        <div class="no-registration"></div>
+                                    </div>
+                                {% endif %}
+                                {% endif %}
```
se-bastiaan, line 29: These texts and indicators have changed much more. See the newer implementation on how to set the right text and colour for each event.

website/education/templates/education/frontpage_summaries.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n bleach_tags alert %}
+<div class="frontpage-vacancies">
```
se-bastiaan, line 2:
```suggestion
<div class="frontpage-summaries">
```

website/education/templates/education/frontpage_summaries.html
```diff
@@ -0,0 +1,28 @@
...
+            {% trans 'There are currently no course summaries' as info_text %}
+            {% alert 'info' info_text dismissible=False %}
+        </div>
+    {% else %}
+        <div class="list-group">
+            {% for summary in summaries %}
+                <a class="list-group-item py-3"
+                   href="{{ summary.course.get_absolute_url }}">
+                    <span class="vacancy-title mr-3">{{ summary.course.name }}</span>
+                    <span class="vacancy-company" title="{{ summary.name }}">{{ summary.name }}</span>
```
se-bastiaan, line 16:
```suggestion
                    <span class="summary-course-title mr-3">{{ summary.course.name }}</span>
                    <span class="summary-name" title="{{ summary.name }}">{{ summary.name }}</span>
```

se-bastiaan (changes requested):
Was this supposed to be a draft? Doesn't feel finished 😅

website/thaliawebsite/templates/index.html
JobDoesburg: I figured alumni are not that interested in summaries, so I kept the layout there

JobDoesburg:
> I know have not really been able to visit Technicie evenings, but I still feel kinda sad that you didn't tell me that you didn't like it and just decided to leave it out instead of just messaging me 'hey sebas, we decided to leave it out and we opened a reverting MR'. 

To be honest, I was personally thinking it was Wouters code and we did tell it him at the Technicie (+ we discussed it with all other people there at the time). So this was just a miscommunication then...

JobDoesburg:
> Completely unrelated to the previous point: how many summaries are actually posted? Maybe the other education documents are also useful to make sure the widget doesn't stay the same for a long time.

I think not too many, but maybe it will change after we make them more visible (kinda similar to what we're doing with vacancies)

website/thaliawebsite/templates/index.html
se-bastiaan: Summaries? huh?

This exact tag is also used on the alumni page to render the alumni-only events there. I changed that to the list to not keep both, but that is a good place for the cards imo.

JobDoesburg: Nvm I was confused. I don't necessarily want to remove the code because it is a nice building block...

se-bastiaan (approved):

PR #2067, opened 2021-12-08
Add short description field to events

### LINKED ISSUE #2056: Add a caption information field to event creation
Posting on Instagram with 20 lines of text is usually not that engaging. The newsletter descriptions are quite long.

As of now you make a title and a (longer) description for an event. The idea is to add a middle ground ,a caption of 2 to 3 sentences. This would be just another text field in between the other event information fields.

### Motivation
This is important to make both the newsletter event descriptions shorter and to allow for more effective promotion on social media.

CatoOri:
Closes #2067 

### Summary
Added the feature to add a short description of events for promotion and the newsbrief
### How to test
1. Go to 'site administration'
2. Go to add event
3. Scroll down to 'detail(Show)'
4. There should be a field to enter a caption


website/newsletters/templates/newsletters/email.html
```diff
@@ -93,7 +93,7 @@ <h2 style="{{ item_h2_style }}">{{ item.title.upper }}</h2>
                 {% if item.newsletterevent %}
                 <tr>
                     <td style="{{ item_tr_style }}">
-                        {{ item.description|safe }}
+                        {{ item.short|safe }}
```
pingiun, line 96: This `item.description` refers to the newsletter description. If you want to fill in the short description from the event in the newsletter, you will need to change it in the javascript here: https://github.com/svthalia/concrexit/blob/507a901d9c3d1b669933c275cafe4b806bba9be5/website/newsletters/static/admin/newsletters/js/forms.js#L28

website/events/services.py
```diff
@@ -294,6 +294,7 @@ def registration_fields(request, member=None, event=None, registration=None, nam
                 "type": field.type,
                 "label": field.name,
                 "description": field.description,
+                "short description": field.short,
```
pingiun, line 297: The API keys shouldn't contain spaces, as this can create some hassle. I propose you call this field "caption" everywhere, except maybe in the user messages

pingiun (changes requested):

pingiun:
I think for the failing GitHub Action, you just need to remove the migration and recreate it again

pingiun (approved):

JobDoesburg (dismissed):

ColonelPhantom (changes requested):
Can you add the caption to API v2 as well? It might also be nice to have it when fetching the list of events as well as a single event, as we probably want the caption to be visible in a potential overview as well. So my suggestion would be adding to the event list for v1 next to the event detail, and add for both the event list and event detail in v2.

Jorritboer (approved):
Looks good. 
We do need to remember to explicitly mention to Walter in the release email this has changed in the newsletter. It's missed easily but could cause some confusion.

ColonelPhantom (approved):
Seems good! It's still not in the list in API v1, but that's not really needed anyway.

I did not look at the resulting newsletter however, but I assume Jorrit did that.

PR #2070, opened 2021-12-09
Add-categories-to-vacancies-api

DeD1rk:
Adds vacancy categories to the api. The categories are added to the vacancy serializer. There is also a filter to get vacancies that have at least one of a comma-separated list of category slugs, and an endpoint for the list of all categories.

## How to test?
Try out the new endpoint,  and the filter.

### DIFF
--- a/website/partners/api/v2/filters.py
+++ b/website/partners/api/v2/filters.py
@@ -66,11 +66,7 @@ class VacancyCategoryFilter(filters.BaseFilterBackend):
         categories = request.query_params.get("categories", None)
 
         if categories:
-            categories_set = VacancyCategory.objects.filter(
-                slug__in=categories.split(",")
-            ).all()
-
-            queryset = queryset.filter(categories__in=categories_set)
+            queryset = queryset.filter(categories__slug__in=categories.split(","))
 
         return queryset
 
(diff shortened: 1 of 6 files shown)


website/partners/api/v2/filters.py
```diff
@@ -56,3 +57,30 @@ def get_schema_operation_parameters(self, view):
...
+
+    def filter_queryset(self, request, queryset, view):
+        categories = request.query_params.get("categories", None)
+
+        if categories:
+            categories_set = VacancyCategory.objects.filter(
+                slug__in=categories.split(",")
+            ).all()
+
+            queryset = queryset.filter(categories__in=categories_set)
```
se-bastiaan, line 73:
```suggestion
            queryset = queryset.filter(categories__slug__in=categories.split(","))
```

This does not work?

DeD1rk: I don't think so, because categories is a Many2Many field, so there is no `categories.slug`. I only looked online though, did not actually try that way, but I'll give it a try now.

DeD1rk: Oh very cool, that does seem to work!

se-bastiaan (dismissed):

JobDoesburg (approved):

PR #2075, opened 2021-12-13
Introduce CleanedModelSerializer

JobDoesburg:
### Summary
Django rest framework's ModelSerializer does not `clean()` it's instances before saving. This introduces a CeanedModelSerializer that does do this.

Note that the sales app does not use these serializers, as the update and create method there already have been overwritten (and using validate actually breaks stuff)

### How to test
1. All API's should be tested...

### DIFF
--- a/website/thaliawebsite/api/v1/cleaned_model_serializer.py
+++ b/website/thaliawebsite/api/v1/cleaned_model_serializer.py
@@ -1,0 +1,11 @@
+from rest_framework import serializers
+
+
+class CleanedModelSerializer(serializers.ModelSerializer):
+    """Custom ModelSerializer that explicitly clean's a model before it is saved."""
+
+    def validate(self, attrs):
+        super().validate(attrs)
+        instance = self.Meta.model(**attrs)
+        instance.clean()
+        return attrs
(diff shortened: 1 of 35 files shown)


website/thaliawebsite/api/v1/cleaned_model_serializer.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import serializers
+
+
+class CleanedModelSerializer(serializers.ModelSerializer):
+    """Custom ModelSerializer that explicitly clean's a model before it is saved."""
+
+    def validate(self, attrs):
```
se-bastiaan, line 7: I am not sure we should add this. v1 has been running a long time and this might break things.

JobDoesburg: Well if it breaks stuff, it breaks on adding stuff that doesn't `clean`. 

What I've learned from this bug (and that it has been here for a long time without us noticing in fact) is that we rely a lot on front end logic. I really think we should clean in v1 too!

se-bastiaan (approved):

PR #2086, opened 2021-12-15
Add dataminimisation for payments, sales and pizzas

### LINKED ISSUE #880: Data minimization on payments and event registrations after 7 years.

### LINKED ISSUE #1653: Minimise sales order data
The new `sales` app stores quite some member credentials that we don't need to keep for a long time

Anonymise the orders (setting the Member field to NULL) after a while (30 days? 1 year? directly when a shift closes)

Note that this also needs an update in the privacy statement!

### Motivation
Data minimization 

Right now this will happen when a member object is deleted, but that takes too long in my opinion. There's no need to keep an order history in my opinion.

### Additional context
- We shouldn't delete the orders themselves, we want to keep them for multiple reasons
- This is no true data minimization, as orders will have payments, and those payments still have a user connected, but that's another issue.

JobDoesburg:
Closes #880 and closes #1653

### Summary
Adds data minimization for payments, sales and pizzas

### How to test
1. Have old data
2. Execute dataminimisation
3. Verify the data is minimized 
4. Verify that things don't crash after minimization

### DIFF
--- a/website/payments/services.py
+++ b/website/payments/services.py
@@ -191,7 +191,7 @@ def send_tpay_batch_processing_emails(batch):
     return len(member_payments)
 
 
-def execute_data_minimization(dry_run=False):
+def execute_data_minimisation(dry_run=False):
     """Anonymizes payments older than 7 years."""
     # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
     deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 7))
--- a/website/payments/services.py
+++ b/website/payments/services.py
@@ -200,5 +200,5 @@ def execute_data_minimization(dry_run=False):
         paid_by__isnull=True
     )
     if not dry_run:
-        queryset.update(paid_by=None)
+        queryset.update(paid_by=None, processed_by=None)
     return queryset
--- a/website/payments/services.py
+++ b/website/payments/services.py
@@ -189,3 +189,16 @@ def send_tpay_batch_processing_emails(batch):
             },
         )
     return len(member_payments)
+
+
+def execute_data_minimization(dry_run=False):
+    """Anonymizes payments older than 7 years."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 7))
+
+    queryset = Payment.objects.filter(created_at__lte=deletion_period).exclude(
+        paid_by__isnull=True
+    )
+    if not dry_run:
+        queryset.update(paid_by=None)
+    return queryset.all()
--- a/website/payments/services.py
+++ b/website/payments/services.py
@@ -201,4 +201,4 @@ def execute_data_minimization(dry_run=False):
     )
     if not dry_run:
         queryset.update(paid_by=None)
-    return queryset.all()
+    return queryset
--- a/website/payments/templates/payments/payment_list.html
+++ b/website/payments/templates/payments/payment_list.html
@@ -92,6 +92,9 @@
                 {% trans "You have not made any payments during this period." as no_payments %}
                 {% alert 'info' no_payments %}
             {% endif %}
+            <span class="mt-3 font-italic">
+                {% trans "Thalia only stores your payments for 7 years." %}
+            </span>
         </div>
     </section>
 {% endblock %}
--- a/website/pizzas/services.py
+++ b/website/pizzas/services.py
@@ -33,12 +33,14 @@ def can_change_order(member, food_event):
     )
 
 
-def execute_data_minimization(dry_run=False):
+def execute_data_minimisation(dry_run=False):
     """Anonymizes pizzas orders older than 3 years."""
     # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
     deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 3))
 
-    queryset = FoodOrder.objects.filter(food_event__end__lte=deletion_period).exclude(name="<removed>")
+    queryset = FoodOrder.objects.filter(food_event__end__lte=deletion_period).exclude(
+        name="<removed>"
+    )
     if not dry_run:
         queryset.update(member=None, name="<removed>")
     return queryset
--- a/website/pizzas/services.py
+++ b/website/pizzas/services.py
@@ -1,3 +1,5 @@
+from django.utils import timezone
+
 from events.services import is_organiser
 from .models import Product, FoodOrder
 
@@ -29,3 +31,22 @@ def can_change_order(member, food_event):
         and member.has_perm("pizzas.change_foodorder")
         and is_organiser(member, food_event.event)
     )
+
+
+def execute_data_minimization(dry_run=False):
+    """Anonymizes pizzas orders older than 3 years."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 3))
+
+    queryset_members = FoodOrder.objects.filter(created_at__lte=deletion_period).filter(
+        payer__isnull=False
+    )
+    queryset_nonmembers = (
+        FoodOrder.objects.filter(created_at__lte=deletion_period)
+        .filter(name__isnull=False)
+        .exclude(name="<deleted>")
+    )
+    if not dry_run:
+        queryset_members.update(payer=None)
+        queryset_nonmembers.update(name="<deleted>")
+    return queryset_members.union(queryset_nonmembers).all()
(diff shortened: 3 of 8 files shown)


JobDoesburg:
@pingiun do you know what's wrong with postgres

pingiun:
> @pingiun do you know what's wrong with postgres

```
======================================================================
243
ERROR: test_data_minimisation (payments.tests.test_services.ServicesTest) [Payments that are 7 years old must be minimised]
244
----------------------------------------------------------------------
245
Traceback (most recent call last):
246
  File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
247
    return self.cursor.execute(sql, params)
248
psycopg2.errors.NumericValueOutOfRange: numeric field overflow
249
DETAIL:  A field with precision 4, scale 2 must round to an absolute value less than 10^2.
```

JobDoesburg:
> > @pingiun do you know what's wrong with postgres
> 
> ```
> ======================================================================
> 243
> ERROR: test_data_minimisation (payments.tests.test_services.ServicesTest) [Payments that are 7 years old must be minimised]
> 244
> ----------------------------------------------------------------------
> 245
> Traceback (most recent call last):
> 246
>   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
> 247
>     return self.cursor.execute(sql, params)
> 248
> psycopg2.errors.NumericValueOutOfRange: numeric field overflow
> 249
> DETAIL:  A field with precision 4, scale 2 must round to an absolute value less than 10^2.
> ```

Yeah I see that, but it doesn't happen locally and I don't see on what field it's actually crashing... It's not the `amount`...

pingiun:
> > > @pingiun do you know what's wrong with postgres
> > 
> > 
> > ```
> > ======================================================================
> > 243
> > ERROR: test_data_minimisation (payments.tests.test_services.ServicesTest) [Payments that are 7 years old must be minimised]
> > 244
> > ----------------------------------------------------------------------
> > 245
> > Traceback (most recent call last):
> > 246
> >   File "/usr/local/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
> > 247
> >     return self.cursor.execute(sql, params)
> > 248
> > psycopg2.errors.NumericValueOutOfRange: numeric field overflow
> > 249
> > DETAIL:  A field with precision 4, scale 2 must round to an absolute value less than 10^2.
> > ```
> 
> Yeah I see that, but it doesn't happen locally and I don't see on what field it's actually crashing... It's not the `amount`...

It doesn't happen locally because sqlite stores rationals as strings I believe

pingiun:
A field with precision 4 scale 2 can only store 4 significant digits i believe, so max 99.99

JobDoesburg:
> A field with precision 4 scale 2 can only store 4 significant digits i believe, so max 99.99

The weird thing is that  we should have 6 digits and 2 decimals...

website/payments/services.py
```diff
@@ -189,3 +189,16 @@ def send_tpay_batch_processing_emails(batch):
             },
         )
     return len(member_payments)
+
+
+def execute_data_minimization(dry_run=False):
```
se-bastiaan, line 194: minimisation or minimization?

I think the member code use the `s`.

website/pizzas/services.py
```diff
@@ -29,3 +31,22 @@ def can_change_order(member, food_event):
...
+    queryset_members = FoodOrder.objects.filter(created_at__lte=deletion_period).filter(
+        payer__isnull=False
+    )
+    queryset_nonmembers = (
+        FoodOrder.objects.filter(created_at__lte=deletion_period)
+        .filter(name__isnull=False)
+        .exclude(name="<deleted>")
+    )
+    if not dry_run:
+        queryset_members.update(payer=None)
```
se-bastiaan, line 50: `payer` does not exist on this model

website/pizzas/services.py
```diff
@@ -29,3 +31,22 @@ def can_change_order(member, food_event):
...
+        queryset_nonmembers.update(name="<deleted>")
+    return queryset_members.union(queryset_nonmembers).all()
```
se-bastiaan, line 52:
```suggestion
    queryset = FoodOrder.objects.filter(created_at__lte=deletion_period).exclude(name="<removed>")
    if not dry_run:
        queryset.update(member=None, name="<removed>")
    return queryset
```

Don't see why you need two queries to do this.

website/pizzas/services.py
```diff
@@ -29,3 +31,22 @@ def can_change_order(member, food_event):
...
         and is_organiser(member, food_event.event)
     )
+
+
+def execute_data_minimization(dry_run=False):
+    """Anonymizes pizzas orders older than 3 years."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 3))
+
+    queryset_members = FoodOrder.objects.filter(created_at__lte=deletion_period).filter(
```
se-bastiaan, line 41: Food orders don't have a `created_at`.

website/pizzas/services.py
```diff
@@ -29,3 +31,22 @@ def can_change_order(member, food_event):
         and member.has_perm("pizzas.change_foodorder")
         and is_organiser(member, food_event.event)
     )
+
+
+def execute_data_minimization(dry_run=False):
+    """Anonymizes pizzas orders older than 3 years."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 3))
```
se-bastiaan, line 39: Maybe this should be a setting or global thing somewhere? Would that make sense?

website/sales/services.py
```diff
@@ -19,3 +21,16 @@ def is_manager(member, shift):
...
     return False
+
+
+def execute_data_minimization(dry_run=False):
+    """Anonymizes orders older than 3 years."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 3))
+
+    queryset = Order.objects.filter(created_at__lte=deletion_period).exclude(
+        member__isnull=True
```
se-bastiaan, line 32:
```suggestion
        payer__isnull=True
```

website/payments/services.py
```diff
@@ -189,3 +189,16 @@ def send_tpay_batch_processing_emails(batch):
...
+    """Anonymizes payments older than 7 years."""
+    # Sometimes years are 366 days of course, but better delete 1 or 2 days early than late
+    deletion_period = timezone.now().date() - timezone.timedelta(days=(365 * 7))
+
+    queryset = Payment.objects.filter(created_at__lte=deletion_period).exclude(
+        paid_by__isnull=True
+    )
+    if not dry_run:
+        queryset.update(paid_by=None)
```
se-bastiaan, line 203:
```suggestion
        queryset.update(paid_by=None, processed_by=None)
```

This is also personal information.

website/payments/services.py
```diff
@@ -189,3 +189,16 @@ def send_tpay_batch_processing_emails(batch):
...
+    return queryset.all()
```
se-bastiaan, line 204:
```suggestion
    return queryset
```

don't think you need that

se-bastiaan (changes requested):
🚒

JobDoesburg:
Another thing I just thought about: Django Admin LogEntry's might need to be anonymized too

website/pizzas/services.py
JobDoesburg: Let me create a new issue for this, because I like the idea!

JobDoesburg: See #2099

website/payments/templates/payments/payment_list.html
```diff
@@ -92,6 +92,9 @@ <h5 class="mb-4">
                 {% trans "You have not made any payments during this period." as no_payments %}
                 {% alert 'info' no_payments %}
             {% endif %}
+            <span class="mt-3 font-italic">
+                {% trans "Thalia only stores your payments for 7 years." %}
```
se-bastiaan, line 96: Maybe add that this is a legal requirement?

se-bastiaan (approved):

website/payments/templates/payments/payment_list.html
JobDoesburg: Meh, they can read it in the privacy policy. The reason I added this here was just to explain why you don't see everything.

pingiun (approved):

PR #2088, opened 2021-12-15
Member since displays wrong date

### LINKED ISSUE #1929: Membership since displaying wrong date

mark-boute:
Closes #1929 

Yes so this is still broken.
Test also needs to be fixed, but querysets are a bit annoying.

I believe my function to be correct, so I guess the error I am currently encountering is either in views.py or in profile.html

Wut:
image

### DIFF
--- a/website/members/models/member.py
+++ b/website/members/models/member.py
@@ -4,6 +4,7 @@ import logging
 import operator
 from datetime import timedelta
 from functools import reduce
+from itertools import groupby
 
 from django.contrib.auth.models import User, UserManager
 from django.db.models import Q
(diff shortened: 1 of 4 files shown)


website/members/models/member.py
```diff
@@ -4,6 +4,7 @@
 import operator
 from datetime import timedelta
 from functools import reduce
+from itertools import groupby
```
mark-boute, line 7: to be removed

se-bastiaan (changes requested):
Why is this a new PR? https://github.com/svthalia/concrexit/pull/1938 is still open.

PR #2092, opened 2021-12-16
Use caption for front page events

pingiun:
### Summary
The shorter captions are now used for the front page articles. I think this will look nicer than long descriptions that are cut off.

Not that this will display N/A for events that were created before captions were mandatory

### How to test
1. Go to frontpage
2. See short descriptions

### DIFF
--- a/website/events/templates/events/event_cards.html
+++ b/website/events/templates/events/event_cards.html
@@ -27,7 +27,7 @@
 
                                 <h6 class="card-date">{{ next_event.event.start|date:"d F Y" }}</h6>
                                 <h5 class="card-title">{{ next_event.event.title }}</h5>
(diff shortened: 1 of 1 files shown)


se-bastiaan (approved):

pingiun:
We could of course plan this PR for the release after the upcoming release. That way the events on the frontpage will mostly have short descriptions

JobDoesburg (approved):

JobDoesburg:
> We could of course plan this PR for the release after the upcoming release. That way the events on the frontpage will mostly have short descriptions

Nah let's just do this right away, worst case I will fix the captions manually

ColonelPhantom (approved):
Maybe we could have detect whether we want the caption or description for now, but nice!

PR #2093, opened 2021-12-16
Use event caption for opengraph description

JobDoesburg:
### Summary
Use event caption for opengraph description

### How to test
1. Visit the event details page
2. Verify that the `og:description` uses the caption

For example, use https://www.opengraph.xyz


se-bastiaan (approved):

se-bastiaan (changes requested):
You should probably add a fallback to the description in this case.

se-bastiaan (approved):

ColonelPhantom (approved):

PR #2094, opened 2021-12-16
Add staging url to deploy workflow

pingiun:
### How to test
1. Go to environments on github
2. Check that url is shown

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -7,7 +7,9 @@ on:
 jobs:
   deployment:
     runs-on: ubuntu-latest
-    environment: staging_environment
+    environment: 
+      name: staging_environment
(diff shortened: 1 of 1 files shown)


JobDoesburg (approved):
Why not just `name: staging`

.github/workflows/deploy.yaml
```diff
@@ -7,7 +7,9 @@ on:
 jobs:
   deployment:
     runs-on: ubuntu-latest
-    environment: staging_environment
+    environment: 
+      name: staging_environment
```
se-bastiaan, line 11:
```suggestion
      name: staging
```
?

pingiun: Yea i'll do that some time but it requires me to recreate the environment in the repo settings and copy all the secrets again

PR #2096, opened 2021-12-16
Refactor terraform

pingiun:
### Summary
Terraform code was a bit messy and didn't allow a clean addition of review environments. Now this is easier to add

### DIFF
--- a/infra/stages/staging/main.tf
+++ b/infra/stages/staging/main.tf
@@ -18,25 +18,10 @@ provider "aws" {
   region  = var.aws_region
 }
 
-module "concrexit_network" {
-  source = "../../modules/concrexit_network"
-  stage  = "staging"
-  tags = {
-    "Category"    = "concrexit",
-    "Owner"       = "technicie",
-    "Environment" = "staging",
-    "Terraform"   = true
-  }
-  customer = var.customer
-}
-
 module "concrexit" {
-  source = "../../modules/concrexit"
-  depends_on = [
-    aws_route53_record.www,
-    aws_route53_record.wildcard
-  ]
-  stage = "staging"
+  source     = "../../modules/concrexit"
+  depends_on = [module.concrexit_dns]
+  stage      = "staging"
   tags = {
     "Category"    = "concrexit",
     "Owner"       = "technicie",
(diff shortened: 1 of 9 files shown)


infra/stages/staging/main.tf
```diff
@@ -18,18 +18,70 @@ provider "aws" {
...
+  # This weird construction is because only when creating the instance is the
+  # ipv6 address known, but the elastic IP is created beforehand.
+  public_ipv4      = module.concrexit_network.public_ipv4
+}
+
+data "aws_route53_zone" "primary" {
+  name = var.domain_name
+}
+
+resource "aws_route53_record" "www" {
```
se-bastiaan, line 62: Feels really weird to have these here, are they not part of concrexit?

se-bastiaan (changes requested):
The idea of the multiple modules is really nice. But you could also put those one layer lower. Like https://github.com/svthalia/discord-bot/tree/master/terraform/modules/authentication/modules/lambda. That way there is less duplication for each stage.

infra/stages/staging/main.tf
pingiun: They will be different for staging/production/review so I could enable them conditionally in the concrexit module or create a new module but this also seemed fine

pingiun:
> The idea of the multiple modules is really nice. But you could also put those one layer lower. Like https://github.com/svthalia/discord-bot/tree/master/terraform/modules/authentication/modules/lambda. That way there is less duplication for each stage.

Good idea!

pingiun:
> The idea of the multiple modules is really nice. But you could also put those one layer lower. Like https://github.com/svthalia/discord-bot/tree/master/terraform/modules/authentication/modules/lambda. That way there is less duplication for each stage.

The concrexit_network module is only used in the staging and production environments and wouldn't be used in the review environments. This is why I want to place the module instantiation in those stages

infra/stages/staging/main.tf
se-bastiaan: Feels like a lot of duplication, since the zone_id, name and ipv4 can just be provided to the module itself.

se-bastiaan:
> > The idea of the multiple modules is really nice. But you could also put those one layer lower. Like https://github.com/svthalia/discord-bot/tree/master/terraform/modules/authentication/modules/lambda. That way there is less duplication for each stage.
> 
> The concrexit_network module is only used in the staging and production environments and wouldn't be used in the review environments. This is why I want to place the module instantiation in those stages

Don't you want the same type of environment for all stages? maybe I am missing something.
It's possible to just enable/disable the modules like you already mentioned in the other comment.

The (or my, cannot find any website explaining this in detail) idea of the stage setups is that it only provides the necessary date to setup an environment for a system. If we conditionally want to enable/disable something that can be an input of the system. And if necessary the whole app can depend on that single setting to (not) do something.

infra/stages/staging/main.tf
se-bastiaan: If anything, this could probably be a `concrexit-domain` module or something. That way I'd be happy.

pingiun: This is also needed because the letsencrypt stuff in the nixos config will only work if the A record exists, so those are given as dependencies to the concrexit module

pingiun:
> Don't you want the same type of environment for all stages? maybe I am missing something.

There's only a limited amount of elastic IPs and VPCs you can make, so I want the review environments to share the same VPC and get a normal dynamic IP just like they do right now

se-bastiaan (approved):

se-bastiaan (approved):

PR #2097, opened 2021-12-16
Introduce PaymentAmountField

### LINKED ISSUE #2091: Payment amount seems to have max value of 99.99
### Describe the change
https://github.com/svthalia/concrexit/pull/2086#issuecomment-995261018 suggests that on production database systems, the maximum value for payment amount is 99.99... This is intended to be 9999.99, though... This should be verified and possibly fixed

### Motivation
99.99 is high, but not uncommonly high for payments

### Current implementation
We use DecimalFields with 6 digits and 2 decimals. But apparently that does not work properly?

### Suggested implementation
- Verify existence of the issue
- Perhaps, subclass DecimalField ourselves to some payment amount field, that auto-sets the number of digits and decimals.

JobDoesburg:
Closes #2091  (hopefully)

### Summary
Introduces a PaymentAmountField to be used for all payment amounts.

### How to test
Try to create payments with the interesting amounts from the test cases.

Note that I couldn't test it properly locally, because I don't run postgres and review environments are broken. And it appears as if the exact validation and error handling really matters here. The good thing is that we use the same number of digits everywhere, so create_payable will not fail at least


JobDoesburg:
deploy review

JobDoesburg:
remove review

JobDoesburg:
deploy review

JobDoesburg:
remove review

se-bastiaan (approved):

PR #2105, opened 2021-12-21
Fix newsletter for up to three local partners

### LINKED ISSUE #2041: Add support for multiple local partners

wdoeland:
Closes #2041

### Summary
Not all local partners were shown in the newsletter. Now they are all shown.

### How to test
1. Create a new newsletter and preview it 'on the site'.
2. Make 0 partners local partner
3. Refresh and see 0 local partners in the newsletter
4. Do steps 2-3 with 0, 1, 2 and 3 partners and see 0, 1, 2 and 3 local partners in the newsletter.

### DIFF
--- a/website/newsletters/templates/newsletters/email.html
+++ b/website/newsletters/templates/newsletters/email.html
@@ -200,7 +200,7 @@
         </table>
     </div>
     {% endif %}
-    {% if local_partners.all %}
+    {% if local_partners.exists %}
     <div style="width: 700px ; margin-top: 30px; margin-left: auto ; margin-right: auto;">
         <table cellspacing="0" cellpadding="0">
             {% if local_partners|length < 2 %}
--- a/website/newsletters/templates/newsletters/email.html
+++ b/website/newsletters/templates/newsletters/email.html
@@ -189,7 +189,7 @@
                 <td align="center" style="color: black; font-size: 12px;">
                     <p style="margin: 10px; font-family: 'Gill Sans', 'Trebuchet MS', sans-serif;">
                         <a href="{{ main_partner.link }}">
-                            <img src="{{ main_partner.logo.url }}" height="95" style="height: 95px;" alt="{{ main_partner.name }}"/>
+                            <img src="{% baseurl %}{{ main_partner.logo.url }}" height="95" style="height: 95px;" alt="{{ main_partner.name }}"/>
                         </a><br/>
                         {% trans "our main partner"|upper %}
                     </p>
@@ -217,7 +217,9 @@
                     {% if local_partner and local_partner.is_active %}
                     <td align="center" style="color: black; font-size: 12px;">
                         <p style="margin: 10px; font-family: 'Gill Sans', 'Trebuchet MS', sans-serif;">
-                            <a href="{{ local_partner.link }}"><img src="{{ local_partner.logo.url }}" height="70" style="height: 70px; margin-bottom: 25px" alt="{{ local_partner.name }}"/></a><br/>
+                            <a href="{{ local_partner.link }}">
+                                <img src="{% baseurl %}{{ local_partner.logo.url }}" height="70" style="height: 70px; margin-bottom: 25px" alt="{{ local_partner.name }}"/>
+                            </a><br/>
                             {% trans "our local partner"|upper %}
                         </p>
                     </td>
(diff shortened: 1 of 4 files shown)


SebastiaanSafeguard (changes requested):
-

se-bastiaan (changes requested):
With this approach any change to the partner logos as uploaded by an administrator could break versions of the newsletter when opened at a later time. I would suggest trying to include them by using base64 images.

wdoeland:
> With this approach any change to the partner logos as uploaded by an administrator could break versions of the newsletter when opened at a later time. I would suggest trying to include them by using base64 images.

A few things on this:
- I think base64 images don't work in a lot of email clients. 
- Partners (and their images) are not deleted but they are just not set to be active anymore.
- I also don't think it's a big issue if an image (with an alt text!) doesn't display properly after a while in a weekly newsletter, does anyone reread these? It's much more important to make sure the email works properly now.

pingiun:
> I think base64 images don't work in a lot of email clients.

I don't think that's true

website/newsletters/templates/newsletters/email.html
```diff
@@ -180,29 +180,16 @@ <h2 style="{{ item_h2_style }}">{{ item.title.upper }}</h2>
...
-                        {% trans "our local partner"|upper %}
-                    </p>
-                </td>
-                {% endif %}
                 {% if main_partner and main_partner.is_active %}
                 <td align="center" style="color: black; font-size: 12px;">
                     <p style="margin: 10px; font-family: 'Gill Sans', 'Trebuchet MS', sans-serif;">
                         <a href="{{ main_partner.link }}">
-                            <img src="{% baseurl %}{% static 'newsletters/nedap-logo.png' %}" height="95" style="height: 95px;" alt="{{ main_partner.name }}"/>
+                            <img src="{{ main_partner.logo.url }}" height="95" style="height: 95px;" alt="{{ main_partner.name }}"/>
```
pingiun, line 192: This should include the baseurl. Relative urls don't work in emails

pingiun:
> * Partners (and their images) are not deleted but they are just not set to be active anymore

When an image is changed for a partner the url can also change. I don't think we delete the old image. What do you think @se-bastiaan should we just merge this?

se-bastiaan:
Other small note: the text version of the newsletter has not been changed yet.

website/newsletters/templates/newsletters/email.html
```diff
@@ -180,29 +180,16 @@ <h2 style="{{ item_h2_style }}">{{ item.title.upper }}</h2>
...
-                        {% trans "our local partner"|upper %}
-                    </p>
-                </td>
-                {% endif %}
                 {% if main_partner and main_partner.is_active %}
                 <td align="center" style="color: black; font-size: 12px;">
                     <p style="margin: 10px; font-family: 'Gill Sans', 'Trebuchet MS', sans-serif;">
                         <a href="{{ main_partner.link }}">
-                            <img src="{% baseurl %}{% static 'newsletters/nedap-logo.png' %}" height="95" style="height: 95px;" alt="{{ main_partner.name }}"/>
+                            <img src="{% baseurl %}{{ main_partner.logo.url }}" height="95" style="height: 95px;" alt="{{ main_partner.name }}"/>
```
se-bastiaan, line 192: Can you use the thumbnail instead of the full image

se-bastiaan: `{% thumbnail main_partner.logo THUMBNAIL_SIZE_MEDIUM fit=False %}`

website/newsletters/templates/newsletters/email.html
```diff
@@ -212,6 +200,36 @@ <h2 style="{{ item_h2_style }}">{{ item.title.upper }}</h2>
         </table>
     </div>
     {% endif %}
+    {% if local_partners.all %}
```
se-bastiaan, line 203:
```suggestion
    {% if local_partners.exists %}
```

Dit is een hele zware query die niet nodig is :)

se-bastiaan (approved):

PR #2118, opened 2021-12-22
Hotfix payment amount 0 for free events

JobDoesburg:
### Summary
Fixes a bug that events cannot be free, because PaymentAmountFields cannot be null

### How to test
1. Make a 'free' event
2. Make sure it clean()'s and it still appears as a free event everywhere on the website

### DIFF
--- a/website/events/models/event.py
+++ b/website/events/models/event.py
@@ -134,12 +134,14 @@ class Event(models.Model):
 
     price = PaymentAmountField(
         verbose_name=_("price"),
+        allow_zero=True,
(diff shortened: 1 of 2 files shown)


se-bastiaan:
@JobDoesburg Is it not possible to allow 0 as value for events somehow?

JobDoesburg:
I agree that this is an easier fix (though my idea behind the previous solution was that it felt more intuitive to have `fine=None` if there is no fine)

pingiun (approved):

PR #2124, opened 2021-12-23
Fix serializer validation

JobDoesburg:
### Summary
Thoroughly fixes model validation after API updates, and changes some things to the EventRegistrationAdminSerializer to make it more convenient.

### How to test
1. Try to POST, PUT and PATCH to the API, bot things that should and should not clean. For example, an Event Registration with both a `member` and `name` field.

### DIFF
--- a/website/events/api/v2/admin/serializers/event_registration.py
+++ b/website/events/api/v2/admin/serializers/event_registration.py
@@ -25,14 +25,36 @@ class EventRegistrationAdminSerializer(CleanedModelSerializer):
             "name",
             "special_price",
             "remarks",
+            "queue_position",
+            "is_cancelled",
+            "is_late_cancellation",
         )
         read_only_fields = ("payment",)
+        optional_fields = ["payment", "member", "name", "special_prize", "remarks"]
 
-    payment = PaymentSerializer()
+    payment = PaymentSerializer(required=False)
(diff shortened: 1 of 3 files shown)


website/events/api/v2/admin/serializers/event_registration.py
```diff
@@ -25,14 +25,36 @@ class Meta:
             "name",
             "special_price",
             "remarks",
+            "queue_position",
```
se-bastiaan, line 28: This is repeated

website/events/api/v2/admin/serializers/event_registration.py
```diff
@@ -25,14 +25,36 @@ class Meta:
...
+    def _is_late_cancellation(self, instance):
+        return instance.is_late_cancellation()
+
+    def _queue_position(self, instance):
+        pos = instance.queue_position
+        return pos if pos and pos > 0 else None
 
-    payment = PaymentSerializer()
+    def _is_cancelled(self, instance):
+        return instance.date_cancelled is not None
```
se-bastiaan, line 53: With `date_cancelled` available this is kind of redundant.

se-bastiaan (commented):
Don't really have time to check if it actually works, but the cleaned model serializer does look good.

pingiun (approved):
Not sure if this should work but I want to test it out

PR #2129, opened 2021-12-24
Remove Docker related configuration

pingiun:
### Summary
Remove Docker related configuration

### How to test
N/A


JobDoesburg:
I would actually propose to keep the docker config as a backup for maintainability of this repo. Deploying a docker is a lot easier (better documentation available online) then the whole NixOS setup (even if we have terraform). I think it is good to keep building Docker images on every release. What do you think?

pingiun:
If things break in the docker stuff we will not notice because we don't use it. And if the dockerfiles need updating (right now to python 3.9 for example) we have no reason to do it because we don't use it. If we ever want to get back on docker we can dig up the files from the repo history

se-bastiaan (dismissed):

se-bastiaan:
> If we ever want to get back on docker we can dig up the files from the repo history

Exactly that.

JobDoesburg (approved):

PR #2131, opened 2021-12-27
Increase nginx client_max_body_size

pingiun:
We used to have this in our previous configuration, and I forgot to add it to the nix configs. This allows members to upload photo albums up to 2GiB


se-bastiaan (approved):
Nice catch, someone already had problems with this?

pingiun:
> Nice catch, someone already had problems with this?

Yep I got a message from Jelte

PR #2133, opened 2021-12-27
Create and update terraform plan for releases

pingiun:
### Summary
Creates a plan file that later can be used to deploy production and do exactly what the user expects

### How to test
1. Dispatch workflow
2. See changes that would be applied to production

### DIFF
--- a/.github/workflows/release-terraform-plan.yaml
+++ b/.github/workflows/release-terraform-plan.yaml
@@ -1,0 +1,33 @@
+name: "Store terraform plan"
+on:
+  release:
+    types: [published]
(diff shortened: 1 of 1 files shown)


se-bastiaan (commented):
What's wrong with creating the plan locally when you execute this?

pingiun:
> What's wrong with creating the plan locally when you execute this?

This is meant to complement #2134

> Kinda feel like this should be in the same PR as #2134

Yea maybe you're right, or we can just merge both

se-bastiaan (approved):

PR #2134, opened 2021-12-27
Add manual GitHub Action for production deployment

pingiun:
### Summary
This action reads in a filename and deploys the terraform plan from that filename. This requires the person doing the deployment to read the terraform plan beforehand, because otherwise they wouldn't know the filename. The action from #2133 can generate and push that plan file


JobDoesburg (dismissed):
LGTM (with my extremely limited terraform and nix knowledge)

PR #2145, opened 2021-12-30
Backport release/40: Add ghostscript to path

pingiun:
### Summary
Add ghostscript to path


pingiun:
Nvm I think it's better to base the 40.2 release of off master

PR #2146, opened 2021-12-31
Fix event registration admin serializer to not throw a 404 on schema generation

se-bastiaan:
### Summary
`/api/docs` is broken because the schema url throws a 404. This is because the 404 is thrown in one of the views during the schema generation step.

### How to test
1. Go to `/api/docs`
2. You have a schema, instead of none


JobDoesburg (approved):

ColonelPhantom (approved):
Fixes the issue :)

PR #2154, opened 2022-01-09
Upgrade to bootstrap 5

### LINKED ISSUE #2139: Upgrade to bootstrap5
### Describe the change
[As django-bootstrap4 seems to go unsupported](https://github.com/svthalia/concrexit/pull/2059#issuecomment-1001773582), and bootstrap 5 is more lightweight, we should upgrade to bootstrap 5.

### Motivation
Bootstrap 5 is more lightweight and more than bootstrap 4.

### Current implementation
Bootstrap 4

### Suggested implementation
django-bootstrap5

JobDoesburg:
Closes #2139 

### Summary
- Upgrades to bootstrap5
- Related: subtle changes to the 'become a member/benefactor' page
- Related: changes some SelectDateWidgets to default input fields with type=date, to get browser-support (which currently is supported by all major browsers)
- Related: Moves the displaying of messages to the base template

Did not yet remove all jQuery

### How to test
1. Make sure all front end pages work properly and look properly


JobDoesburg:
Todo:

- [ ] Fix styling of the file fields image
- [ ] Fix date selection fields image
- [ ] Verify correct functioning of messages
- [ ] Drop jQuery were possible

se-bastiaan:
I instantly noticed a different container size and that the menu does not have the same padding with respect to the login button. I have found that `pr-lg-4` is now `pe-lg-4`

se-bastiaan:
Also found `ml-1` in the thabloid template tags

se-bastiaan:
image

se-bastiaan:
image

se-bastiaan:
The contact page also has a different padding in the table. And no top border. Not sure if that is on purpose and if we care. Just noting down all I see.

se-bastiaan:
Links are now underlined. But I quite like that.

se-bastiaan:
image
Dividers in this menu are gone

JobDoesburg:
I see that I didn't properly remove my cache while checking this PR hahah

se-bastiaan:
> I see that I didn't properly remove my cache while checking this PR hahah

You're welcome :)

website/members/templates/members/user/edit_profile.html
```diff
@@ -10,12 +10,6 @@
         <div class="container">
             <h1 class="section-title">{% trans "edit profile"|capfirst %}</h1>
 
-            {% if messages %}
-                {% for message in messages %}
-                    {% alert message.tags message %}
-                {% endfor %}
-            {% endif %}
```
se-bastiaan, line 17: We no longer need this?

JobDoesburg: I replaced it by {% bootstrap messages %} in the base template

website/registrations/templates/registrations/register_benefactor.html
```diff
@@ -95,14 +95,9 @@ <h1 class="section-title">{% trans "registration" %} {% trans "Benefactor" %}</h
                 {% if tpay_enabled %}
                     <hr/>
 
-                    <div class="mt-3 mb-3">
+                    <div class="">
```
se-bastiaan, line 98: Empty class

website/registrations/templates/registrations/register_member.html
```diff
@@ -105,14 +105,9 @@ <h1 class="section-title">{% trans "registration" %} {% trans "Member" %}</h1>
                 {% if tpay_enabled %}
                     <hr/>
 
-                    <div class="mt-3 mb-3">
+                    <div class="">
```
se-bastiaan, line 108: Empty class

website/thaliawebsite/templates/base.html
```diff
@@ -212,8 +218,6 @@
                 src="{% static "js/jquery.min.js" %}"></script>
         <script type="text/javascript"
                 src="{% static "js/js.cookie.min.js" %}"></script>
-        <script type="text/javascript"
-                src="{% static "js/bootstrap.bundle.min.js" %}"></script>
```
se-bastiaan, line 216: I don't understand why we no longer need javascript

se-bastiaan (changes requested):
Good work so far!

website/thaliawebsite/templates/base.html
```diff
@@ -166,9 +168,13 @@
     {% endblock header_image_container %}
 </header>
 
-{% block pre-body %}{% endblock pre-body %}
+{% block pre-body %}
+{% endblock pre-body %}
```
se-bastiaan, line 172: This is unused

website/members/templates/members/user/edit_profile.html
se-bastiaan: Before:

image

Now:

image


Idea is nice, in practice it doesn't improve the UI ;)
Wouldn't do it in this PR tbh.

website/thaliawebsite/templates/base.html
JobDoesburg: we include bootstrapjs already on top

JobDoesburg:
> I see that I didn't properly remove my cache while checking this PR hahah

I just found the problem, I did properly remove my cache but I didn't re-collected my static folder

JobDoesburg:
TODO:

- [ ] Education page
- [ ] Login page
- [ ] Check vacancies
- [ ] Check thabloid

LCKnol (dismissed):

website/thaliawebsite/templates/base.html
se-bastiaan: Oh I see it now. From the package. Does it have an async attribute? Otherwise it should probably be [best](https://stackoverflow.com/questions/436411/where-should-i-put-script-tags-in-html-markup) at the bottom of the page.

se-bastiaan (changes requested):
The accordions still don't open, neither here nor on the vacancies page.
Screenshot 2022-01-24 at 19 14 27

website/thaliawebsite/templates/base.html
se-bastiaan: Adding to this: I noticed that this means we would start using the jdelivr cdn. I personally like keeping our used bootstrap version in the codebase. Even tho that means it needs manual updates.

Jorritboer:
I think there is a _very_ small difference here. Note that the right white border is not the same size as the top and bottom border. This is on the member list page
New:
image
Old:
image

pingiun (approved):

PR #2187, opened 2022-01-26
Cache static content for maximum time

### LINKED ISSUE #2184: Add correct cache headers
### Describe the bug
We used to apply the `Expires` css header in nginx. When we moved to Nix this was removed. We should maybe check out which headers we should add to make sure /static is properly cached

pingiun:
Closes #2184 

### Summary
Allows the browser to cache the files in /static for 10 years without revalidation. Information at: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

### How to test
1. View any page in the browser network inspection tab
2. Check that files from /static are cached


se-bastiaan:
I know we did this before, but do we really want to cache _all_ the logo files etc so long? For css/js we have the automatically added hash, but those files will never get updated in user browsers.

pingiun:
Good point, maybe we should't put those in /static

se-bastiaan:
We cannot not put those in static I think, that wouldn't work with the staticfiles app.
[This](https://docs.djangoproject.com/en/4.0/ref/contrib/staticfiles/#manifeststaticfilesstorage) looks like a possible way of doing this.

pingiun:
I can also put this on the /static/CACHE directory but having the other static files have some kind of hash would work even better

website/thaliawebsite/settings.py
```diff
@@ -651,6 +651,16 @@ def from_env(
     "compressor.finders.CompressorFinder",
 )
 
+NORMAL_STATICFILES_STORAGE = "django.contrib.staticfiles.storage.StaticFilesStorage"
+MANIFEST_STATICFILES_STORAGE = (
+    "django.contrib.staticfiles.storage.ManifestStaticFilesStorage"
+)
```
se-bastiaan, line 657: This could be shorter we leave out the testing option, since that is the same value as development in `setting`

se-bastiaan: Bit nitpicky tho :p

JobDoesburg:
I'm not sure if 10 years isn't too long... What would happen if we update swagger or bootstrap to a new version?

JobDoesburg (approved):

se-bastiaan (approved):

PR #2189, opened 2022-01-28
Fix announcements close button

se-bastiaan:
### Summary
This fixes both the styling and the workings of the announcements close button. You can test this by creating a global site announcement in the admin that is closable.

### How to test
1. Create an announcement that is closable
2. Click on the close button
3. The announcement is removed from the DOM and it is not shown again on refresh.


pingiun (approved):
Nice fix, good that some jquery was removed too

PR #2190, opened 2022-01-28
Fix alert dimissable workings

### LINKED ISSUE #2188: Visual quirk in the login failure message
### Describe the bug
When you fail to log in on desktop (or probably any wider screen) the message saying this shows up on the left and the login fields are offset to the right. I would expect the message to appear on top, while the login fields were to stay in the same horizontal position.

### How to reproduce
1. Go to the login page
2. Enter invalid credentials

### Expected behaviour
The login fields do not move to the right. The error message is either shown above the login fields or next to them.

### Screenshots
Screenshot from 2022-01-27 16-20-21

### Additional context
Not sure in what other situations these message boxes are used, but it may occur in other places too. Haven't checked.

se-bastiaan:
Closes #2188 

### Summary
This fixes the UI for all alerts and makes sure they are actually dismissible.

### How to test
1. Go to the login page
2. Click on login without credentials
3. Note that the error message is correctly placed and can be closed

### DIFF
--- a/website/thaliawebsite/templates/includes/alert.html
+++ b/website/thaliawebsite/templates/includes/alert.html
@@ -1,9 +1,8 @@
-<div class="alert alert-{{ type }} {{ extra_classes }}">
+<div class="alert alert-{{ type }} fade show {{ extra_classes }}">
     {{ message }}
     {% if dismissible %}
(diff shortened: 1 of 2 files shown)


pingiun (approved):
It works

PR #2197, opened 2022-02-02
Revert "Revert "Fix newsletter for up to three local partners (#2105)""

se-bastiaan:
Reverts svthalia/concrexit#2117


wdoeland (approved):

Jorritboer:
I assume the pictures not showing up is just an issue with my local instance right? If so it's looks good
image

se-bastiaan:
> I assume the pictures not showing up is just an issue with my local instance right? If so it's looks good image

Yes the base url is set to thalia.localhost by default. In settings.py

Jorritboer (approved):
In that case it's alright

PR #2201, opened 2022-02-02
TinyMCE5

JobDoesburg:
### Summary
Updates our tinymce dependency

### How to test
1. View event admin change
2. View an event details page

### DIFF
--- a/website/thaliawebsite/settings.py
+++ b/website/thaliawebsite/settings.py
@@ -682,11 +682,10 @@ THUMBNAIL_SIZES = {
 PHOTO_UPLOAD_SIZE = 2560, 1440
 
 # TinyMCE config
-TINYMCE_JS_URL = "/static/tinymce/js/tinymce/tinymce.min.js"
+TINYMCE_JS_URL = "/static/tinymce/tinymce.min.js"
 
 TINYMCE_DEFAULT_CONFIG = {
     "selector": "textarea",
-    "theme": "modern",
     "plugins": "link image paste code contextmenu",
     "toolbar1": "bold italic underline strikethrough | link unlink | "
     "bullist numlist | undo redo | code",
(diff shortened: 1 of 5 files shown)


website/thaliawebsite/settings.py
```diff
@@ -682,11 +682,10 @@ def from_env(
 PHOTO_UPLOAD_SIZE = 2560, 1440
 
 # TinyMCE config
-TINYMCE_JS_URL = "/static/tinymce/js/tinymce/tinymce.min.js"
+TINYMCE_JS_URL = "/static/tinymce/tinymce.min.js"
 
 TINYMCE_DEFAULT_CONFIG = {
     "selector": "textarea",
-    "theme": "modern",
     "plugins": "link image paste code contextmenu",
```
se-bastiaan, line 689: I don't believe the image plugin is still working

se-bastiaan: Screenshot 2022-02-02 at 20 37 42
Missing the button mostly

pingiun: Before:
image

After:
image

website/thaliawebsite/settings.py
```diff
@@ -682,11 +682,10 @@ def from_env(
 PHOTO_UPLOAD_SIZE = 2560, 1440
 
 # TinyMCE config
-TINYMCE_JS_URL = "/static/tinymce/js/tinymce/tinymce.min.js"
+TINYMCE_JS_URL = "/static/tinymce/tinymce.min.js"
```
se-bastiaan, line 685: I don't believe this file is used

se-bastiaan: wait no it is, but we still have a file in our codebase with an old version

se-bastiaan: and this is the default value basically

pingiun (approved):

PR #2202, opened 2022-02-02
Added option for membership to end in future.

### LINKED ISSUE #312: Allow group memberships to end in the future
In GitLab by @joostrijneveld on Jan 30, 2017, 18:44

### One-sentence description

Allow group memberships to end in the future

### Why?

Because currently memberships of board members never really end until changed after the board year, and it could be possible for committee or society members to end their membership on a date that is already known and it would be nice to not have to worry about this anymore.

### Current implementation

`is_staff` is set to True for everyone who becomes a board member and is never set to False because board memberships don't end (until someone changes the value). This is also the reason why committee membership end dates cannot be in the future, since it sets `is_staff` to False if applicable.

### Desired implementation

Set it to False using a cronjob for example

Pokemaster69:
Closes #312 

### Summary
Added funtion to detect if member is not in commitee anymore and revoke staff permissions if this is the case. created job to run daily to execute this command.

### How to test
1. Give member staff permission that is not in a group
2. In the folder website do ./manage.py revoke_staff
3. Observe on website that staff permissions have been revoked from person in step 1
4. Observe in output that the members of who the staff permissions were revoked are printed.

### DIFF
--- a/infra/nixos/concrexit.nix
+++ b/infra/nixos/concrexit.nix
@@ -245,6 +245,7 @@ in
       sendmembershipnotification.calendar = "*-08-31 06:00:00";
       sendinformationcheck.calendar = "*-10-15 06:00:00";
       revokeoldmandates.calendar = "*-*-* 03:00:00";
+      revoke_staff.calendar = "*-*-* 03:00:00";
     };
 
     services = {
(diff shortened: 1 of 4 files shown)


se-bastiaan (approved):
LGTM!

PR #2203, opened 2022-02-02
Add queued filter on event registrations admin and fix queue_position sorting

### LINKED ISSUE #2071: Add invited/queued filter on event registrations admin
Add a filter to `/api/v2/admin/events/<pk>/registrations/`, named `queued` or something like that, that filters out registrations that are in the queue.

The event admin in the app currently shows registrations that are in queue. This is undesired, but there is no filter in the backend, so no neat solution (hence the high priority).

Jorritboer:
Closes #2071 

### Summary
Made `queue_position` an annotation so filtering/sorting could be done on it for the api.

### How to test
1. Test `/api/v2/admin/events/{id}/registrations/` with `queued` filter
2. Test it with `queued` sorting


website/events/models/event_registration.py
```diff
@@ -110,17 +115,26 @@ def information_fields(self):
...
+        NullIf(
+            Greatest(
+                Count(
+                    "event__eventregistration",
+                    filter=Q(event__eventregistration__date_cancelled=None)
+                    & Q(event__eventregistration__date__lt=F("date"))
+                    | (
+                        Q(event__eventregistration__id__lte=F("id"))
+                        & Q(event__eventregistration__date__exact=F("date"))
+                    ),
```
se-bastiaan, line 130:
```suggestion
                    filter=Q(event__eventregistration__date_cancelled=None)
                    & (Q(event__eventregistration__date__lt=F("date"))
                    | 
                        Q(event__eventregistration__id__lte=F("id"))
                        & Q(event__eventregistration__date__exact=F("date"))
                    ),
```
I believe `&` has precedence over `|`

website/events/api/v2/admin/filters.py
```diff
@@ -48,7 +48,35 @@ def get_schema_operation_parameters(self, view):
...
+
+        return queryset.filter(queue_position=None)
+
+    def get_schema_operation_parameters(self, view):
+        return [
+            {
+                "name": "queued",
+                "required": False,
+                "in": "query",
+                "description": "Filter by event registration if they're in the queue",
```
se-bastiaan, line 79:
```suggestion
                "description": "Filter by event registration queue position",
```

se-bastiaan (changes requested):
Just in case

website/events/models/event_registration.py
Jorritboer: Mm maybe.  This is too complicated, I'll look at this Wednesday

se-bastiaan: You currently have:

`( if the registration is not cancelled 'and' the registration date is before F("date") ) 'or' ( if the event registration has the same registration date 'and' the id is lower than F("id") )`

I suggested:

`if the registration is not cancelled 'and' ( the registration date is before F("date") 'or' ( the registration date is the same 'and' the id is smaller ) )`

Jorritboer: Yes you're right. Very keen eye haha
It would be a very weird bug if this were to cause confusion on production...

se-bastiaan (approved):

PR #2206, opened 2022-02-03
Fix tinymce

### LINKED ISSUE #2205: TinyMCE editor shows superfluous markup options
### Describe the bug
PR #2201 changed the TinyMCE settings to include a whole toolbar. And I am not sure why, because we had a carefully selected list of options available. 

### How to reproduce
1. Go to a page with a TinyMCE HTMLField (that works and is not broken)
2. Note that the toolbar contains a thousand options.

### Expected behaviour
The toolbar should only contain a really small subset of HTML options.

### Additional context
We have deliberately chosen for a subset of HTML options in the past because we don't want anyone using the website actually inserting any fancy styles at all. This is for multiple reasons:
 - People within Thalia do crazy stuff, like fill a description with tables and images instead of readable text that also works on the frontpage, on mobile and in the app.
 - Some HTML tags do not have proper styling, like quotes and some of the header styles
 - In some of the HTML editor situations we cannot even support some of the styles, like in the newsletter. Since it needs inline styling on all elements we want the input (which is directly taken from the events) to be as basic as possible.

Some of this behaviour is also enforced by [our bleach configuration](https://github.com/svthalia/concrexit/blob/master/website/thaliawebsite/templatetags/bleach_tags.py). Some of the options available in the toolbar after this change will just be filtered out and are thus there for no reason.

JobDoesburg:
### Summary
Fixes #2201 and closes #2205 

### How to test
Try to change a HTML field

### DIFF
--- a/website/thaliawebsite/settings.py
+++ b/website/thaliawebsite/settings.py
@@ -683,15 +683,16 @@ PHOTO_UPLOAD_SIZE = 2560, 1440
 
 # TinyMCE config
 TINYMCE_DEFAULT_CONFIG = {
-    "height": 500,
+    "max_height": 500,
     "menubar": False,
-    "plugins": "autolink link image code media paste",
+    "plugins": "autolink autoresize link image code media paste",
     "toolbar": "h2 h3 | bold italic underline strikethrough | image media | link unlink | "
                "bullist numlist | undo redo | code",
     "contextmenu": "bold italic underline strikethrough | link",
     "paste_as_text": True,
     "relative_urls": False,
     "remove_script_host": False,
+    "autoresize_bottom_margin": 50,
 }
 
 BOOTSTRAP5 = {"required_css_class": "required-field"}
--- a/website/thaliawebsite/settings.py
+++ b/website/thaliawebsite/settings.py
@@ -685,13 +685,10 @@ PHOTO_UPLOAD_SIZE = 2560, 1440
 TINYMCE_DEFAULT_CONFIG = {
     "height": 500,
     "menubar": False,
-    "plugins": "advlist,autolink,lists,link,image,charmap,hr,codesample,print,preview,anchor,"
-               "searchreplace,visualblocks,code,fullscreen,insertdatetime,media,table,paste,"
-               "code,help,wordcount",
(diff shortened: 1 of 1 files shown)


website/thaliawebsite/settings.py
```diff
@@ -683,20 +683,19 @@ def from_env(
 
 # TinyMCE config
 TINYMCE_DEFAULT_CONFIG = {
-    "selector": "textarea",
```
se-bastiaan, line 686: This selector is no longer required?

website/thaliawebsite/settings.py
```diff
@@ -683,20 +683,19 @@ def from_env(
...
-    "plugins": "advlist anchor autolink autosave charmap code codesample fullscreen hr image media link lists table",
-    "toolbar1": "styleselect | bold italic underline strikethrough | hr charmap codesample | link unlink | bullist numlist outdent indent | table image media | undo redo | code",
-    "contextmenu": "bold italic underline strikethrough | image",
```
se-bastiaan, line 689: We don't want a context menu anymore?

website/thaliawebsite/settings.py
```diff
@@ -683,20 +683,19 @@ def from_env(
...
+    "height": 500,
     "menubar": False,
-    "inline": False,
-    "statusbar": True,
-    "width": "auto",
-    "height": 240,
```
se-bastiaan, line 694: Auto width no longer required? Is 500 not really large on some admin pages?

website/thaliawebsite/settings.py
JobDoesburg: Nope, selecting it will select _all_ texture's on a page. The new library will automatically select based on the field id

website/thaliawebsite/settings.py
JobDoesburg: Auto width is not required, it will automatically fill. 
I figured 240px was quite small (which made sense because earlier, we had both dutch and English fields, but now we only have one).

website/thaliawebsite/settings.py
JobDoesburg: I re-added it

se-bastiaan (approved):

se-bastiaan (approved):

PR #2210, opened 2022-02-09
Album title and date on same line

### LINKED ISSUE #2191: Improve page layout for photo albums
The layout on this page is not optimal
image

Align the date on the right side, at the same line as the album title

### Motivation
Better layout

Keeping it this way

stefanvi:
Closes #2191 

### Summary
I added the date as a span in the h1 tag of the album title. Accounting for the case where the button is not visible and the date should be aligned to the very right edge.

### How to test
1. Go to 'The album page'
2. Click on an album (might need to be created first in the admin)
3. See that the date is now on the same line as the title

### DIFF
--- a/website/photos/templates/photos/album.html
+++ b/website/photos/templates/photos/album.html
@@ -9,24 +9,24 @@
 {% block body %}
     <section id="photos-album" class="page-section">
         <div class="container">
-                <h1 style="float: left" class="section-title">
-                    {{ album.title }}
-                </h1>
-                <h1 style="float: right; padding-top: 20px">
-                    {{ album.date|date:"d-m-Y" }}
-                    {% if album.event %}
-                        <a href="{{ album.event.get_absolute_url }}"
-                           class="btn btn-primary btn-first"
-                           data-bs-toggle="tooltip"
-                           data-bs-placement="top"
-                           title=""
-                           data-original-title="{% trans "Event" %}">
-                           <i class="fas fa-calendar"></i>
-                        </a>
-                    {% endif %}
-                </h1>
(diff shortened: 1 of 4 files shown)


website/photos/templates/photos/album.html
```diff
@@ -12,18 +12,21 @@
...
+                    <a href="{{ album.event.get_absolute_url }}"
+                       style="top: auto"
+                       class="btn btn-primary btn-first"
+                       data-bs-toggle="tooltip"
+                       data-bs-placement="top"
+                       title=""
+                       data-original-title="{% trans "Event" %}">
+                        <i class="fas fa-calendar"></i>
+                    </a>
+                    <span style="position: absolute; right: 50px">{{ album.date|date:"d-m-Y" }}</span>
```
se-bastiaan, line 24: If we would style every element with inline CSS it would get messy very quickly. Aka: this should be in a SCSS file :)

stefanvi:
I've made quite some changes. Importantly, h1 is now a flex container and the btn-first and btn-second were removed. Now if you want to align something to the right of the title, you add the `first-right` class to the first element that should be right aligned. All following items will be right aligned.

This is used to solve the original problem of right aligning the date and button of a photo album on the same line. I also applied it to the calendar.

website/photos/templates/photos/album.html
stefanvi: Fair point. Current solution should be a lot less messy and no longer requires hard coding things like `btn-first`! Let me know what you think

pingiun (approved):

ColonelPhantom (approved):

PR #2221, opened 2022-02-16
Fix grid item image for chromium

JobDoesburg:
### Summary
Fixes that images for grid items did not properly fill, in chromium browsers


se-bastiaan (changes requested):

pingiun (approved):
Worked in FF and chrome

PR #2222, opened 2022-02-16
Fix international office url in student well-being page

JobDoesburg:
### Summary
Fix international office url in student well-being page


pingiun (approved):
I'll take your word for it

PR #2226, opened 2022-02-16
Display random current summaries instead of always the latest

JobDoesburg:
### Summary
Display random current summaries instead of always the latest

### DIFF
--- a/website/education/templatetags/frontpage_summaries.py
+++ b/website/education/templatetags/frontpage_summaries.py
@@ -1,10 +1,6 @@
 from django import template
-from django.utils import timezone
-from django.utils.translation import gettext as _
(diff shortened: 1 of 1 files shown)


pingiun:
The header will still say "recent"

JobDoesburg:
> The header will still say "recent"

Same for vacancies

se-bastiaan:
That was a deliberate choice so that all partners are happy :) 

Maybe it would be nice if the summaries are actually recent? Choose some random options from the 25 last uploaded documents or something?

JobDoesburg:
I did filter on courses that 'exist', but maybe I could choose the most recent document per course or something

PR #2228, opened 2022-02-16
Add endpoint for unpublished external events to CalendarJS API

ColonelPhantom:
Fixes the bug that CalendarJS did not show unpublished internal events, and external events had a wrong link (to the internal event with the same ID).

### Summary

### How to test

### DIFF
--- a/website/events/api/calendarjs/serializers.py
+++ b/website/events/api/calendarjs/serializers.py
@@ -87,6 +87,21 @@ class UnpublishedEventsCalenderJSSerializer(CalenderJSSerializer):
     def _registration_info(self, instance):
         return "Unpublished event"
 
+class UnpublishedExternalEventCalendarJSSerializer(CalenderJSSerializer):
+    """See CalenderJSSerializer, customised classes."""
+
+    class Meta(CalenderJSSerializer.Meta):
+        model = Event
+
+    def _class_names(self, instance):
+        return ["unpublished-event"]
+
+    def _url(self, instance):
+        return reverse("admin:events_externalevent_change", kwargs={"object_id": instance.id})
+
+    def _registration_info(self, instance):
+        return "Unpublished event"
+
 
 class ExternalEventCalendarJSSerializer(CalenderJSSerializer):
     """External event calender serializer."""
--- a/website/events/api/calendarjs/urls.py
+++ b/website/events/api/calendarjs/urls.py
@@ -5,6 +5,7 @@ from events.api.calendarjs.views import (
     CalendarJSEventListView,
     CalendarJSUnpublishedEventListView,
     CalendarJSExternalEventListView,
+    CalendarJSUnpublishedExternalEventListView,
 )
 
 app_name = "events"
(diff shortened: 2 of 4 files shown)


website/events/api/calendarjs/serializers.py
```diff
@@ -87,6 +87,21 @@ def _url(self, instance):
...
+    """See CalenderJSSerializer, customised classes."""
+
+    class Meta(CalenderJSSerializer.Meta):
+        model = Event
+
+    def _class_names(self, instance):
+        return ["unpublished-event"]
+
+    def _url(self, instance):
+        return reverse("admin:events_externalevent_change", kwargs={"object_id": instance.id})
```
se-bastiaan, line 100: Should this not use pk?

website/events/api/calendarjs/urls.py
```diff
@@ -25,4 +26,10 @@
         CalendarJSUnpublishedEventListView.as_view(),
         name="calendarjs-unpublished",
     ),
+    path(
+        "external/unpublished/",
```
se-bastiaan, line 30:
```suggestion
        "external-events/unpublished/",
```

website/events/api/calendarjs/views.py
```diff
@@ -56,7 +57,7 @@ def get_queryset(self):
 class CalendarJSUnpublishedEventListView(ListAPIView):
     """Define a custom route that outputs the correctly formatted external events information for CalendarJS, unpublished events only."""
 
-    queryset = ExternalEvent.objects.filter(published=False)
+    queryset = Event.objects.filter(published=False)
```
se-bastiaan, line 60: This is the actual required bugfix @pingiun

se-bastiaan: I think this something with the rebase that went wrong when fixing conflicts

website/events/api/calendarjs/serializers.py
ColonelPhantom: pk gave an error

website/events/api/calendarjs/views.py
se-bastiaan: The rest is technically not required to fix the bug and get back to original functionality

website/events/api/calendarjs/urls.py
ColonelPhantom: I simply followed the conventions set by the other endpoints, and there is no `external-events/` endpoint, only `external/`

website/events/api/calendarjs/views.py
pingiun: I think we should only have this one line fix then

ColonelPhantom: Correct that this is the fix, but I did not want to remove unpublished external events from the calendar, hence the rest of the PR.

pingiun: The rest of the PR is untested code, I just want to merge the fix so we can release on time

se-bastiaan (changes requested):
I think it would be better to fix the bug that we are having and then introduce the full new endpoint in a later release

Maybe even merge the unpublished events into the regular feeds.

ColonelPhantom:
Have we dealt with this issue of not being able to see unpublished externals events yet? If not, do we still want to do this by bringing this PR up to date?

T8902:
Since this pr is dead for over a year ill just close it

PR #2234, opened 2022-02-23
Release fixes

pingiun:
Last release we had an issue where the terraform deploy action couldn't run. I believe this fixes that issue


JobDoesburg (approved):
With my very limited terraform knowledge I think this works well

PR #2236, opened 2022-02-24
Refactor templates

JobDoesburg:
Closes #2227 

### Summary
Refactors templates to use small_page or simple_page. Note that templates with multiple sections are not migrated.

### How to test
1. Make sure that all HTML pages look good and all JS still works on them

PR #2237, opened 2022-02-24
Refactor templates

JobDoesburg:
Closes #2272

### Summary
Refactors templates to use small_page or simple_page. Note that templates with multiple sections are not migrated.

### How to test
1. Make sure that all HTML pages look good and all JS still works on them

### DIFF
--- a/website/members/templates/members/statistics.html
+++ b/website/members/templates/members/statistics.html
@@ -1,76 +1,63 @@
-{% extends "base.html" %}
+{% extends "simple_page.html" %}
 {% load i18n static compress %}
 
 {% block title %}{% trans "Statistics" %} — {{ block.super }}{% endblock %}
 {% block opengraph_title %}{% trans "Statistics" %} — {{ block.super }}{% endblock %}
 
-{% block body %}
-    <section class="page-section" id="members-statistics" data-statistics="{{ statistics }}">
-        <div class="container">
-            <h1 class="section-title">{% trans "Statistics" %}</h1>
+{% block page_title %}{% trans "Statistics" %}{% endblock %}
+{% block section_tags %}id="members-statistics" data-statistics="{{ statistics }}"{% endblock %}
 
-            <h2 class="mb-4">{% trans "Total amount of Thalia members" %}: {{ total_members }}</h2>
-
-            <div class="row">
-                <div class="col-6">
-                    <canvas id="members-type-chart"
-                            data-data="{{ member_type_distribution }}"
-                            data-title="Membership Types (currently)"
-                            data-chart-type="bar">
-                    </canvas>
-                </div>
-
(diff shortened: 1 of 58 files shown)


pingiun:
[website/singlepages/templates/singlepages/become_active.html](https://github.com/svthalia/concrexit/pull/2236/files/5ce2794e06bfb0215ea13bbdfe32fa51bf3bcd34#diff-7ce82231149aa02853c5fa5772e5644fff6ed348093439146a3ef59d0d80be6c)
[website/education/templates/education/books.html](https://github.com/svthalia/concrexit/pull/2236/files/5ce2794e06bfb0215ea13bbdfe32fa51bf3bcd34#diff-cf8d04d94fa898be4d0e4c38ccf10e3b908ffc5d1bb596e04812ff9f85d0e3cc), [website/payments/templates/payments/bankaccount_list.html](https://github.com/svthalia/concrexit/pull/2236/files/5ce2794e06bfb0215ea13bbdfe32fa51bf3bcd34#diff-2d43bdba64084b35501e822afeff5f6f28b109b3bd206508368ba9a7009822b6)

Look better as a small_page imo

JobDoesburg:
- For the bank account, I don't agree. It would require the payments overview and 'add bankaccount form' page to be small too in my opinion, which looks weird.
- For book sales and become active, I'm in doubt. The small page imo works best for forms or simple user interaction like a success message. Not necessarily for just info pages.

se-bastiaan:
I don't think this PR is about changing the UI, rather about refactoring the templates. So which page looks better with which template sounds irrelevant to me. :)

In general I like a better template, but I am not entirely sure here. It feels like hiding the actual implementation of each page in a base template. I don't like that, it makes understanding the UI of the page much harder. And all because of two classes that are on a tag.

website/members/templates/members/statistics.html
```diff
@@ -1,76 +1,63 @@
...
-                <div class="col-6">
-                    <canvas id="active-members-chart"
-                            data-data="{{ active_members }}"
-                            data-title="Active vs Non-active members"
-                            data-chart-type="pie">
-                    </canvas>
-                </div>
-            </div>
+{% block page_title %}{% trans "Statistics" %}{% endblock %}
+{% block section_tags %}id="members-statistics" data-statistics="{{ statistics }}"{% endblock %}
```
se-bastiaan, line 8: This seems to be the only place where we are adding more than just the id to the HTML. Which makes me think that the statistics could be inserted another way and the block just be `section_id` or `page_id` by placing the id on the body or something.

pingiun (dismissed):

website/members/templates/members/statistics.html
JobDoesburg: we decided to not change this in this PR

PR #2239, opened 2022-02-24
Copy product name for order items

### LINKED ISSUE #1655: Copy product information to payable instead of using a foreign key
Right now, it is possible to change the description of an order even after it has been paid, by just changing the product. Only the price is stored separately and cannot be cascade-changed

When a `Product` is used in an order that is paid, the product should be immutable

### Motivation
Consistency of data

Copying fields, that just creates more database columns that we don't really need

### Additional context
It should be fine to delete products (and then, the product can just be SET NULL)

JobDoesburg:
Closes #1655

### Summary
Copies the product name for to order items, to allow changing product names

### How to test
1. Have an order with products
2. Change a product name or delete the product
3. The product name should still exist for the order item
4. Also test the API


pingiun (changes requested):
Seems good if the migration also copies the name

pingiun (approved):

PR #2245, opened 2022-03-01
Add promotion module

nvoers:
### Summary
Added proper permission management. Added restrictions on publish date (minimum one week from now). Remade emails; from one per request to one weekly email. Added status update actions.

### How to test
Steps to test the changes made

publish date:
1. Go to promotion requests
2. Click on add promotion request
3. Fill in date field

update actions:
1. Go to promotion requests
2. Select one or multiple requests
3. Select action to apply to requests


JobDoesburg:
Can you rebase on master :)

ColonelPhantom:
I didn't get this running, due to some dependency issue which I don't think is the fault of this PR :sweat_smile: 

Anyway, I think my review was only requested due to some other version.

website/promotion/admin.py
```diff
@@ -16,8 +16,49 @@ class PromotionRequestAdmin(admin.ModelAdmin):
...
+    mark_finished.short_description = ("Mark requests as finished")
+
+    def mark_published(self, request, queryset):
+        """Change the status of the event to published."""
+        self._change_published(request, queryset, PromotionRequest.PUBLISHED)
+
+    mark_published.short_description = ("Mark requests as published")
+
+    @staticmethod
+    def _change_published(request, queryset, status):
```
JobDoesburg, line 52: `request` param is actually unused here

website/promotion/apps.py
```diff
@@ -13,4 +13,4 @@ class PromotionConfig(AppConfig):
     def ready(self):
         """Import the signals when the app is ready."""
         # pylint: disable=unused-import,import-outside-toplevel
-        from . import signals
+        #from . import signals
```
JobDoesburg, line 16: I don't think this is what you want

website/promotion/admin.py
```diff
@@ -16,8 +16,49 @@ class PromotionRequestAdmin(admin.ModelAdmin):
...
+        queryset.update(status=status)
 
 
 @admin.register(PromotionChannel)
 class PromotionChannelAdmin(ModelAdmin):
-    pass
+    list_display = (
+        "name",
+    )
```
JobDoesburg, line 60: What does this change?

website/promotion/emails.py
```diff
@@ -1,16 +1,55 @@
...
+
+logger = logging.getLogger(__name__)
+
+def send_weekly_overview():
+
+    new_requests = services.get_new_requests(timezone.localtime())
+    upcoming_requests = services.get_upcoming_requests(timezone.localdate())
+
+    from_email = settings.PROMO_REQUEST_NOTIFICATION_ADDRESS
+    text_template = get_template("requests/weekly_overview.txt")
```
JobDoesburg, line 22: please use utils.send_email for sending emails and don't try things like setting language and translations or catching SMTP errors yourself in here.

website/promotion/forms.py
```diff
@@ -0,0 +1,33 @@
...
+            "channel",
+            "assigned_to",
+            "status",
+            "drive_folder",
+            "remarks",
+        ]
+    
+    def clean_publish_date(self):
+        publish_date = self.cleaned_data.get("publish_date")
+        create_time_minimum = publish_date - timezone.timedelta(weeks=1)
```
JobDoesburg, line 22: Let's make this timedelta a setting, because it feels like something that might change in the future

website/promotion/models.py
```diff
@@ -78,6 +79,15 @@ def __str__(self):
         if self.event:
             return _("Promotion request for ") + str(self.event)
         return _("Promotion request ") + str(self.pk)
+    
+    def clean(self):
+        super().clean()
+        errors = {}
+        if self.publish_date is None:
```
JobDoesburg, line 86: setting null and blank to False already does this check for you

website/promotion/services.py
```diff
@@ -0,0 +1,13 @@
+from django.utils import timezone
+from promotion.models import PromotionRequest
+
+def get_new_requests(end_date):
```
JobDoesburg, line 4: It would be nicer to implement this as a custom ModelManager, like in here for example https://docs.djangoproject.com/en/4.0/topics/db/managers/

JobDoesburg (changes requested):
I'm not sure if this is really what we want. People will not be able to create a request for less than a week, and I think if that's not allowed, committees will just skip the promo. So in the end I think this will result in a _less_ promotion instead of _better_ promo...

website/promotion/emails.py
```diff
@@ -1,16 +1,30 @@
...
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
-
 from utils.snippets import send_email
+from django.template.loader import get_template
+from promotion.models import PromotionRequest
+
+logger = logging.getLogger(__name__)
+
+def send_weekly_overview():
```
JobDoesburg, line 12: To send this every week, a management command calling this function should be created, and a cronjob should be added to https://github.com/svthalia/concrexit/blob/c9762db146168569870d2936179e0a92041bb3a4/infra/nixos/concrexit.nix#L230-L250

JobDoesburg (changes requested):
Looks good, apart from this one comment

JobDoesburg (approved):

nvoers:
@JobDoesburg yeah all fine!

PR #2254, opened 2022-03-09
Make price serializers match PaymentAmountField

ColonelPhantom:
@TippieCodes ran into an issue with the API crashing on serializing certain fields, apparently a bogus high fine.

I found out that this is because the serializer has a lower `max_digits` than `PaymentAmountField`.

I fixed this both for fines, and the other `DecimalField` serializers I could find that were related to `PaymentAmountField`s.

### How to test
1. Create an event with a huge fine (6 digits or more)
2. Open the event list in the API
3. Observe that it does not crash


JobDoesburg:
I would be in favor of creating our own PaymentAmountSerializerField

ColonelPhantom:
I added the custom `PaymentAmountSerializer`. There are some remaining usages of `DecimalField`, we may want to check if those are still appropriate (e.g. for contribution, which I guess is not really Thalia Pay but is a payment). I asked `ripgrep` to make me a list of where (excluding migrations and inheritance):
```
website/sales/api/v2/admin/serializers/shift.py
43:    total_revenue = serializers.DecimalField(

website/registrations/forms.py
152:    contribution = forms.DecimalField(

website/payments/api/v2/serializers/payable_detail.py
1:from rest_framework.fields import CharField, DecimalField, BooleanField
10:    amount = DecimalField(decimal_places=2, max_digits=1000, source="payment_amount")

website/payments/api/v2/admin/serializers/payable_detail.py
1:from rest_framework.fields import CharField, ListField, DecimalField
20:    amount = DecimalField(decimal_places=2, max_digits=1000, source="payment_amount")

website/payments/api/v2/serializers/payment_user.py
1:from rest_framework.fields import DecimalField, BooleanField
8:    tpay_balance = DecimalField(decimal_places=2, max_digits=1000)

```

JobDoesburg (approved):

PR #2268, opened 2022-03-30
Add april fools prank

pingiun:
### Summary
Ssht don't tell anyone, this will rotate the whole website on april fools

### How to test
1. Wait until april fools
2. See the whole website is slightly rotated

### DIFF
--- a/website/thaliawebsite/templates/base.html
+++ b/website/thaliawebsite/templates/base.html
@@ -40,6 +40,13 @@
             <link href="{% static "css/main.scss" %}" rel="stylesheet" type="text/x-scss">
         {% endcompress %}
     {% endblock %}
+    {% if APRIL_FOOLS %}
+    <style>
+        body {
+            transform: rotate(.5deg);
+        }
+    </style>
+    {% endif %}
 
     {% block js_head %}
     {% endblock %}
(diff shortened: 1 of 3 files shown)


website/thaliawebsite/templates/base.html
```diff
@@ -40,6 +40,13 @@
             <link href="{% static "css/main.scss" %}" rel="stylesheet" type="text/x-scss">
         {% endcompress %}
     {% endblock %}
+    {% if APRIL_FOOLS %}
+    <style>
+        body {
+            transform: rotate(.5deg);
```
se-bastiaan, line 46: May I suggest

```
import random
from django import template

register = template.Library()

@register.simple_tag
def random_float(a, b=None):
    if b is None:
        a, b = 0, a
    return round(random.uniform(a, b), 2)
```

with 

`{% load random_float %}`

and 

`rotate({% random_float 2 -2 %}deg);`

pingiun: no, because this will be removed in the release after the next one anyway

pingiun: oops sorry, didn't see what you were actually suggesting

pingiun: i think it's fine the way it is though, very subtle

se-bastiaan (changes requested):
Timezones bah bah bah

se-bastiaan (approved):

PR #2269, opened 2022-03-30
Remove aprilfools

pingiun:
Merge this after april fools has happened and everyone laughed a lot at the prank

### DIFF
--- a/website/thaliawebsite/templates/base.html
+++ b/website/thaliawebsite/templates/base.html
@@ -40,13 +40,6 @@
             <link href="{% static "css/main.scss" %}" rel="stylesheet" type="text/x-scss">
         {% endcompress %}
     {% endblock %}
-    {% if APRIL_FOOLS %}
-    <style>
-        body {
-            transform: rotate(.5deg);
-        }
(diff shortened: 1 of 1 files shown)


JobDoesburg:
maybe we want to keep the context preprocessor?

pingiun:
> maybe we want to keep the context preprocessor?

good idea!

website/thaliawebsite/settings.py
```diff
@@ -472,7 +472,6 @@ def from_env(
                 "django.contrib.messages.context_processors.messages",
                 "announcements.context_processors.announcements",
                 "thaliawebsite.context_processors.thumbnail_sizes",
-                "thaliawebsite.context_processors.aprilfools",
```
JobDoesburg, line 475: I think it would also be fine to keep this (but also fine to remove it, so I don't care too much actually)

pingiun: Lets also keep this, it makes it less confusing to eventually use again

JobDoesburg (approved):

PR #2285, opened 2022-04-13
Configure logrotate

pingiun:
### Summary

Rotate logs every day and only keep logs for 5 days.

### How to test
1. Ssh to the server
2. `ls -al /var/log/nginx`
3. See if there are .1 .2 files etc

### DIFF
--- a/infra/nixos/concrexit.nix
+++ b/infra/nixos/concrexit.nix
@@ -343,8 +343,24 @@ in
             "\"\"" = {
               default = !cfg.allowUnknownHost;
               locations."/".return = "444";
+              extraConfig = ''
+                access_log off;
+              '';
             };
           };
+      };
+
+      logrotate = {
+        enable = true;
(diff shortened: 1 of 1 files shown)


infra/nixos/concrexit.nix
```diff
@@ -343,10 +343,26 @@ in
...
             };
           };
       };
 
+      logrotate = {
+        enable = true;
+        extraConfig = ''
+          /var/log/nginx/*.log {
+            daily
+            rotate 5
```
JobDoesburg, line 358: Maybe 5 days is a bit too short? Technicie meets weekly, so having a 2 or at least 1 week log rotation doesn't sound weird to me (even though we hardly ever use them)

pingiun: I agree

JobDoesburg (approved):

PR #2296, opened 2022-04-15
Add data retention period to event information fields

### LINKED ISSUE #1949: Automatically remove registration information fields
Wild idea (maybe too complex though): add a "remove data after" field to registration info fields. As the kind of data in there can differ, maybe our retention policy should also be set per case. For example, when visiting an in-house day at the ministry of defense where we ask for a pasport number, we might want to set a deletion date of the day after the event, while for other information we might want a longer period. This way, we also enforce people to think about when to delete the data (and we could have that field be max. 5 years)

_Originally posted by @JobDoesburg in https://github.com/svthalia/concrexit/pull/1935#discussion_r716887942_

JobDoesburg:
Closes #1949

### Summary
Adds a data retention period to event info fields


website/events/models/registration_information_field.py
```diff
@@ -40,6 +40,15 @@ class RegistrationInformationField(models.Model):
         _("required"),
     )
 
+    delete_after = models.DateField(
```
se-bastiaan, line 43: Why don't we delete this after a predetermined time? This way it could technically be saved forever if people enter a value that is years in the future.

website/events/admin/inlines.py
```diff
@@ -6,7 +6,7 @@
 from pizzas.models import FoodEvent
 
 
-class RegistrationInformationFieldInline(admin.TabularInline):
```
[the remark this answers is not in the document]
se-bastiaan: hahahaha

website/events/models/registration_information_field.py
JobDoesburg: Also possible. The reason I opened the issue myself is because we don't have an exact policy on this, and regarding data minimization I figure that some kinds of data might have different retention times than others and this forces organizers to think about it. We could add a validator of `max 3 years` or something. But I'm open for discussion on this

se-bastiaan: I would personally say that 1/2-ish months after the activity ended would also be fine. So many options. As long as it is not 'unlimited'.

JobDoesburg: As a constant? Or as max value?

se-bastiaan: Constant

JobDoesburg: @pingiun what do you think?

pingiun: Then I think 2 months after the event is over we can delete the registration info

se-bastiaan: Isn't it a bit our job to have thought about this for them?

JobDoesburg: No, because we don't know what kind of data this is, so we are not able to make a judgement on what the retention period should be.

T8902:
Too long inactive

PR #2299, opened 2022-04-15
Add global uniqueness constraint on push notification tokens

### LINKED ISSUE #2177: Add global uniqueness constraint on push notification tokens
### Describe the bug
Currently, in a `Device`, the combination of `user` and `registration_id` (a Firebase Messaging token) is unique. This means that it's possible for one token to be used by multiple users. This makes it possible for one device to receive notifications intended for multiple users. For example.

Alice has registered a device with token `a`. Bob also registers a device with token `a` (although he doesn't have a device that receives from token `a`). Alice now receives push notifications that are sent only to Bob. 

https://github.com/svthalia/concrexit/blob/3dceb39fe0e0d251c0e8a81393d40d1ad6b564e2/website/pushnotifications/models.py#L68-L72

JobDoesburg:
Closes #2177

### Summary
Add global uniqueness constraint on push notification tokens


se-bastiaan (dismissed):
Just needs codestyle fixes

se-bastiaan (approved):

PR #2300, opened 2022-04-15
Django webauth

JobDoesburg:
Closes #ISSUE

### Summary

### How to test

### DIFF
--- a/website/thaliawebsite/settings.py
+++ b/website/thaliawebsite/settings.py
@@ -389,6 +389,7 @@ INSTALLED_APPS = [
     "django.contrib.staticfiles",
     "django.contrib.sitemaps",
     # Dependencies
+    "webauth",
     "oauth2_provider",
     "corsheaders",
     "django_bootstrap5",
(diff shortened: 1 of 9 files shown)


website/thaliawebsite/settings.py
```diff
@@ -604,6 +605,23 @@ def from_env(
...
+    development="localhost:8000",
+    production="thalia.nu",
+    staging="staging.thalia.nu",
+)
+WEBAUTH_ORIGIN = setting(
+    development="http://localhost:8000",
+    production="https://thalia.nu",
+    staging="https://staging.thalia.nu",
+)
+WEBAUTH_VERIFY_URL = "/user/webauth/verify/"
```
se-bastiaan, line 623: This could reuse the SITE_DOMAIN setting

se-bastiaan:
This doesn't support TOTP right? And how would this integrate with the API?

JobDoesburg:
> This doesn't support TOTP right? 
Yes, this works solely via FIDO. I did just find https://github.com/jazzband/django-two-factor-auth which looks promising too (it's from jazzband) so I might investigate this too

> And how would this integrate with the API?
For the API, one could say that 2FA is also required to authorize OAuth applications with certain requirements, and then the OAuth API key becomes a second factor in itself (as it is stored on the device).

se-bastiaan:
Yes I already mentioned that lib in the issue :)

If the 2fa can be done at login then yes it works perfect with oauth, but it needs to do that obviously

JobDoesburg:
https://www.calazan.com/how-to-enable-2fa-in-the-django-admin/

DeD1rk:
django-webauth calls itself not-production-ready and is not being maintained.
https://github.com/jazzband/django-two-factor-auth has become the standard package to use.

PR #2355, opened 2022-05-26
Properly use Django Storage API backends

se-bastiaan:
### Summary
This PR introduces changes to the codebase that makes sure all file operations use the Django storage backends. One for private files and one for public files. It also adds functionality to be able to use AWS S3 as storage backend instead of a local file system.

### How to test
1. Check that all images, documents and files still work


website/thaliawebsite/management/commands/migratemedia.py
```diff
@@ -0,0 +1,45 @@
...
+from django.core.management.base import BaseCommand
+from django.conf import settings
+
+import logging
+import os
+import boto3
+from botocore.exceptions import ClientError
+
+
+class Command(BaseCommand):
```
JobDoesburg, line 11: Maybe it is nicer to make this a pre or post migration signal?

website/thaliawebsite/management/commands/migratemedia.py
```diff
@@ -0,0 +1,45 @@
...
+    def handle(self, *args, **options):
+        # create session to s3
+        session = boto3.session.Session()
+        s3_client = session.client(
+            service_name="s3",
+            aws_access_key_id=settings.AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=settings.AWS_SECRET_ACCESS_KEY,
+            endpoint_url=f"https://{settings.AWS_STORAGE_BUCKET_NAME}.s3.eu-west-1.amazonaws.com/",
```
JobDoesburg, line 19: this endpoint should also be in settings I think

website/thaliawebsite/management/commands/migratemedia.py
```diff
@@ -0,0 +1,45 @@
...
+from django.conf import settings
+
+import logging
+import os
+import boto3
+from botocore.exceptions import ClientError
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
```
JobDoesburg, line 12: Maybe at a return here if AWS_BUCKET_NAME == None so the script can't crash if testing locally

JobDoesburg (changes requested):

JobDoesburg:
Looks good!

website/thaliawebsite/management/commands/migratemedia.py
se-bastiaan: I am not entirely sure what you mean by such a signal. But technically this script is more a one time thing to transfer everything. Not even sure we need it, since the AWS CLI has the `sync` command that could do the same thing.

website/thaliawebsite/management/commands/migratemedia.py
se-bastiaan: It is in settings, but as PUBLIC_MEDIA_URL. And after further looking into it I don't think we really need it since boto3 can construct this based on the bucket name. And I just copied this from an example :)

JobDoesburg: So do we want to change this?

JobDoesburg (approved):

PR #2374, opened 2022-06-06
Update fancybox to v4

JobDoesburg:
### Summary
Upgrade fancy box to v4 and moved its javascript to the photos app

### How to test
Test that albums still work


se-bastiaan:
> Do we still need jQuery actually, or can we remove it now?

I think mostly older JS code is still using jQuery, especially for requests to the API.

JobDoesburg:
> The thabloid and partners apps also use fancybox.

sad...

pingiun (approved):

PR #2384, opened 2022-06-08
Use load eventlistener for thabloids

JobDoesburg:


website/thabloid/static/thabloid/js/main.js
```diff
@@ -19,11 +19,12 @@ function openThabloidFancybox(e, btn) {
     });
 }
 
-window.onload = (event) => {
+window.addEventListener('load', (event) => {
```
pingiun, line 22:
```suggestion
window.addEventListener('DOMContentLoaded', (event) => {
```

pingiun (changes requested):

website/thabloid/static/thabloid/js/main.js
se-bastiaan: do we actually need this? the script should be at the bottom of the html body

JobDoesburg: jelle notified me that the window.onload can only be used once, so it would be bad practice having it in our codebase

se-bastiaan: no I meant, do we even need the onload at all.

PR #2399, opened 2022-07-03
Add API endpoint to claim sales orders

### LINKED ISSUE #1785: Add sales order API

DeD1rk:
Closes #1785.

### Summary
This adds an endpoint that does pretty much the same as the OrderPaymentView linked to by `order.payment_url` (set `payer` if possible, and warn if under-age for age-restricted orders).

### How to test
1. Create all kinds of orders.
2. Claim the orders from all kinds of users.

### DIFF
--- a/website/sales/api/v2/urls.py
+++ b/website/sales/api/v2/urls.py
@@ -1,7 +1,7 @@
 from django.urls import path
 
 from sales.api.v2.views import (
-    OrderPaymentView,
+    OrderClaimView,
     UserShiftListView,
     UserShiftDetailView,
     UserOrderListView,
@@ -11,7 +11,7 @@ from sales.api.v2.views import (
 app_name = "sales"
 
 urlpatterns = [
-    path("sales/order/<uuid:pk>/pay/", OrderPaymentView.as_view(), name="order-pay"),
+    path("sales/order/<uuid:pk>/claim/", OrderClaimView.as_view(), name="order-claim"),
     path("sales/shifts/", UserShiftListView.as_view(), name="user-shift-list"),
     path(
         "sales/shifts/<int:pk>/",
--- a/website/sales/api/v2/urls.py
+++ b/website/sales/api/v2/urls.py
@@ -1,6 +1,7 @@
 from django.urls import path
 
 from sales.api.v2.views import (
+    OrderPaymentView,
     UserShiftListView,
     UserShiftDetailView,
     UserOrderListView,
@@ -10,6 +11,7 @@ from sales.api.v2.views import (
 app_name = "sales"
 
 urlpatterns = [
+    path("sales/order/<uuid:pk>/pay/", OrderPaymentView.as_view(), name="order-pay"),
     path("sales/shifts/", UserShiftListView.as_view(), name="user-shift-list"),
     path(
         "sales/shifts/<int:pk>/",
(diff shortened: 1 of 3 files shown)


website/sales/api/v2/urls.py
```diff
@@ -10,6 +11,7 @@
 app_name = "sales"
 
 urlpatterns = [
+    path("sales/order/<uuid:pk>/pay/", OrderPaymentView.as_view(), name="order-pay"),
```
JobDoesburg, line 14: I wouldn't call it `pay`, but really call it `claim`. Paying should happen via the payments app.

JobDoesburg (changes requested):

website/sales/api/v2/urls.py
DeD1rk: Fair enough, also kind of a misnomer for the non-api version though, but I'll change it here

JobDoesburg: The non-API view redirects to the payment view so it works a bit differently

JobDoesburg: Let's not forget we should register the non api URL in the apple app site association file thingy so the app will open up

DeD1rk: True, but I'll do that only once the app also actually supports it

website/sales/tests/test_api.py
```diff
@@ -630,6 +631,76 @@ def test_user_self_order(self):
...
+        self.o3.save()
+
+        with self.subTest("Claim an age-restricted order as a minor"):
+            with mock.patch("sales.services.is_adult") as is_adult:
+                is_adult.return_value = False
+                response = self.client.get(
+                    reverse("api:v2:sales:order-pay", kwargs={"pk": self.o3.pk})
+                )
+                self.assertEqual(403, response.status_code)
+                print(response.json)
```
JobDoesburg, line 700:
```suggestion
```
Please remove print statements

JobDoesburg (changes requested):

se-bastiaan (commented):
Even tho I like consistency all the changes in the files that are not in the sales app should technically be in a different PR.

website/sales/api/v2/views.py
```diff
@@ -111,4 +108,35 @@ def destroy(self, request, *args, **kwargs):
...
+class OrderClaimView(RetrieveAPIView):
+    """Claims an order to be paid by the current user."""
+
+    queryset = Order.objects.all()
+    serializer_class = UserOrderSerializer
+    schema = AutoSchema(operation_id_base="claimOrder")
+    permission_classes = [IsAuthenticatedOrTokenHasScope]
+    required_scopes = ["sales:order"]
+
+    def retrieve(self, request, *args, **kwargs):
```
JobDoesburg, line 122: I don't think we should do this on a `GET`, as I wouldn't say it good practice to do any writing on a `GET` (especially not something that is irreversible).

JobDoesburg (changes requested):

JobDoesburg (approved):

PR #2458, opened 2022-09-07
Centralize event status logic and allow message overriding

### LINKED ISSUE #1381: Event status messages per event
### Describe the change
Events can show quite some status messages about being able or not being able to cancel, about fines, about possible 'other' ways of registering, etc. This has become quite messy in our codebase and still it does not always work ideally. I propose to move these fields to the event model and only provide some default texts that can be changed per event. We can then also add the event status message to the API and immediately have fixed it for the app as well

### Motivation
Some problems:

- we almost always forget that when we fix something about an event status message (about the fines or so), that we need to do the same update in the app
- for certain events, the standard status messages are a bit misleading because they just work differently, which results in additional text in the event description
- sometimes things just require a change, for example during this coronavirus times it turned out that a change to the default texts is desirable

### Current implementation
A lot of hardcoded messages

### Suggested implementation
Model fields for all the states an event registration can be in.

- unregistered & registration not open: "You cannot register yet"
- unregistered & registration possible: "Register now"
- unregistered & waiting list: "Put on waiting list"
- registered & before cancellation deadline & re-registration possible: "Cancel registration"
- registered & before cancellation deadline & re-registration not possible: "Cancel registration. Note that you cannot re-register afterwards anymore"
- registered & after cancellation deadline & re-registration possible: "You cannot cancel your registration without having to pay a fine of some amount."
- registered & after cancellation deadline & re-registration not possible: "You cannot cancel your registration without having to pay a fine of some amount. Note that you cannot re-register afterwards anymore."

etc. etc. etc. 

It does add a lot more information to our database, but I do think that in the end it is a nicer way of implementing these fields as the behavior of the system is more transparant and flexible. We move a lot of logic out of the views and into the model, which is something I like.

### LINKED ISSUE #2600: Calendar wrongly shows registered for an event.
### Describe the bug
I was registered for an event, and then deregistered, however the calendar still shows "you are registered for this event". This includes the event being magenta. When going to the event page it does not show as registered.

### How to reproduce
1. Register for an event
2. Deregister
3. Go to calendar
4. See magenta event that you are not registered for

### Expected behaviour
The event should not show as registered

### Screenshots
screenshot-2022-10-16-122038
screenshot-2022-10-16-122028

### Additional context
I am sorry for breaking the 2^6 issues. Clearly the only solution to getting back to such a good number is to fix this issue ASAP

ColonelPhantom:
Closes #1381
Closes #2600 

### Summary
Create a centralized place to determine registration status (e.g. cancelled, cancelled-late, open, closed)

### How to test
1. Create an event
2. Register/deregister/etc (all possible combinations)
3. Check that the text makes sense

I should still check that the addition to the API etc make sense. We should also look into removing the relevant logic from the app, and let it use the status (string etc) returned from the server.


ColonelPhantom:
I think this is ready to be reviewed; there's some more places where we could re-use the logic (such as the front-pages) but I think that should be a separate issue and PR.

ColonelPhantom:
This seems done to me, but feedback on the structure is very welcome! More places where this logic can be applied too, of course.

I also managed to actually fix the n queries with a nice `prefetch_related` into a custom attribute :partying_face:

JobDoesburg (approved):
Quite a monstrosity but I like that it's now all rendered at one place!

PR #2459, opened 2022-09-07
Add possibility to have multiple event organisers

### LINKED ISSUE #786: Multiple organisers per event
In GitLab by @se-bastiaan on Nov 29, 2018, 18:19

### One-sentence description

Multiple organisers per event

### Motivation

I noticed that since the introduction of the Partnercie (once again) multiple committees (and the board) involved in events increased (Lunches are board + partnercie, laser tag was evc + partnercie). It would be quite useful if an event can have multiple organisers. With the new societies this might happen even more often, think 'EVC + board game society'.

### Desired functionality

Select more than one organiser per event.

### Food for thought

- Currently you can only select the groups you belong to as organiser unless you're a board member. That's not useful if you want to add a group as organiser that you're not in
- The organiser determines the mailing list that is used for the unregister notifications. Should all organisers have a mailing list to enable that option? Do we send all organisers that email?

rowan-vr:
Closes #786 

### Summary
Events can now have multiple organisers

### How to test
1. Go to the events admin
2. Create or edit any event
3. Select multiple organisers for the event (it will only show your own groups if you do not have the organiser_override permission)
4. Save.


website/events/admin/event.py
```diff
@@ -252,3 +223,14 @@ def get_urls(self):
             ),
         ]
         return custom_urls + urls
+
+    def get_field_queryset(self, db, db_field, request):
+        """Members without the can view as organiser permission can only assign their own groups as organiser"""
+        pk = resolve(request.path_info).kwargs["object_id"]
+        if db_field.name == "organisers" and not request.user.has_perm(
+            "override_organiser"
```
se-bastiaan, line 231:
```suggestion
            "events.override_organiser"
```
based on other code this is currently incorrect

website/events/admin/event.py
```diff
@@ -252,3 +223,14 @@ def get_urls(self):
             ),
         ]
         return custom_urls + urls
+
+    def get_field_queryset(self, db, db_field, request):
```
se-bastiaan, line 227: This looks cool, maybe we can create an issue to do this in the sales app too? I don't think we have something like this there.

_Or_ we decide it is not really a requirement and don't do it because we made that choice there :)

website/events/admin/event.py
```diff
@@ -100,7 +101,7 @@ class EventAdmin(DoNextModelAdmin):
 
     def get_form(self, request, obj=None, change=False, **kwargs):
         form = super().get_form(request, obj, change, **kwargs)
-        form.clean = lambda form: form.instance.clean_changes(form.changed_data)
+        form.request = request
```
se-bastiaan, line 104: This seems not necessary I think

website/events/admin/forms.py
```diff
@@ -49,3 +50,31 @@ class Meta:
...
         }
+
+
+class EventAdminForm(forms.ModelForm):
+    def clean(self):
+        super().clean()
+        self.instance.clean_changes(self.changed_data)
+
+    def clean_organisers(self):
+        if self.request.user.is_superuser or self.request.user.has_perm(
```
se-bastiaan, line 61: There should be a 'is_organiser' helper function that you can use so that the task of deciding that is in that helper instead of here.

website/events/admin/event.py
se-bastiaan: Same is true for the check in the form btw. I like it tho. Good thinking.

website/events/emails.py
```diff
@@ -46,15 +46,20 @@ def notify_organiser(event, registration):
...
 
-    EmailMessage(
-        subject,
-        text_message,
-        to=[event.organiser.contact_mailinglist.name + "@thalia.nu"],
-    ).send()
+        EmailMessage(
+            subject,
+            text_message,
+            to=[organiser.contact_mailinglist.name + "@thalia.nu"],
```
se-bastiaan, line 64: Instead of repeatedly sending an email you could also add them all to the 'to' line of the one email. Then they also see that they all got the notification.

website/events/admin/forms.py
```diff
@@ -49,3 +50,31 @@ class Meta:
...
+            return self.cleaned_data.get("organisers")
+
+        if (
+            self.request.member
+            and self.cleaned_data.get("organisers")
+            and not (
+                self.request.member.get_member_groups()
+                .filter(pk__in=self.cleaned_data.get("organisers").values_list("pk"))
+                .exists()
+            )
```
se-bastiaan, line 73: It would be neater to split the checks for 'only one' and 'cannot remove yourself', that way this 'if' is easier to read and there can be two error messages shown instead of one that says two things and then leaves you to figure our what exactly (of the two things) you did wrong.

se-bastiaan (commented):
On first glance this looks really good already!
I have added a few small comments.

Maybe also look into the admin overview? It showed the organiser as well I think

website/events/emails.py
se-bastiaan: Or batch the messages together so that it requires only one smtp connection :)

JobDoesburg: Lets do the first

website/events/admin/forms.py
rowan-vr: A different action is needed when the user has event_override, hence that function isn't used

website/events/emails.py
```diff
@@ -56,5 +56,8 @@ def notify_organiser(event, registration):
     EmailMessage(
         subject,
         text_message,
-        to=[event.organiser.contact_mailinglist.name + "@thalia.nu"],
+        to=[
+            organiser.contact_mailinglist.name + "@thalia.nu"
```
JobDoesburg, line 60: I think we shouldn't hardcode thalia.nu, I think we can use the default site from the sites framework

JobDoesburg (approved):
Looks good!

website/events/emails.py
se-bastiaan: Or the base url from the settings. I think we use that in most places.

rowan-vr: Added it! :D

PR #2460, opened 2022-09-07
Fix local partners in newsletter

wdoeland:
### Summary
With this fix we can show an infinite number of local partners in the newsletter!

### How to test
1. Create a newsletter
2. Have _i_ local partners
3. See _i_ local partners in sent and previewed email
4. Use _i_ from 0 to like 5

### DIFF
--- a/website/newsletters/views.py
+++ b/website/newsletters/views.py
@@ -1,4 +1,5 @@
 """Views provided by the newsletters package."""
+import math
 import os
 
 from django.contrib.admin.views.decorators import staff_member_required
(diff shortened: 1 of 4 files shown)


website/newsletters/views.py
```diff
@@ -40,6 +41,14 @@ def preview(request, pk, lang=None):
...
     newsletter = get_object_or_404(Newsletter, pk=pk)
     events = services.get_agenda(newsletter.date) if newsletter.date else None
+    all_local_partners = Partner.objects.filter(is_local_partner=True).order_by("?")
+    local_partner_count = len(all_local_partners)
+    local_partners = []
+    for i in range(math.floor(local_partner_count/2)):
+        local_partners.append([all_local_partners[i*2], all_local_partners[i*2 + 1]])
+
+    if local_partner_count % 2 != 0:
+        local_partners.append([all_local_partners[local_partner_count-1]])
```
se-bastiaan, line 51: You can probably write this piece of code only once.

wdoeland: That is true.

JobDoesburg (approved):

PR #2465, opened 2022-09-08
Mark present url

### LINKED ISSUE #2319: Time bases QR code to register presence for events
Feature request (I'm making this via the Github app so the template doesn't work) 

Registering presence of people is a time consuming thing, sometimes resulting in long queues. Now thalia pay exists, there is no reason to actually have these queues.

I propose a system with a QR code that people can scan to indicate their presence. The QR code should be time based and contain a secret string, so the organiseer can publish it and there is some level of certainty that people are actually there.

DeD1rk:
Closes #2319.

### Summary
This adds a url on which people can mark their presence (simply by GETting it when logged in). The url contains a uuid that is generated per event, but does not expire. So e.g. educacie can show a QR code of that url at the door of a lunch lecture. 

In the admin, buttons are added that show a QR code of the url. 

### How to test
1. Register for an event.
2. Show QR from the admin (changeform object-tools or on the bottom of an event's overview).
3. Scan the QR.
4. See that your registration is now present.


website/events/views.py
```diff
@@ -179,3 +181,30 @@ def dispatch(self, request, *args, **kwargs):
...
+
+        Checks if the url is correct, the event has not ended yet, and the user is registered.
+        """
+        event = get_object_or_404(Event, pk=kwargs["pk"])
+        if kwargs["token"] != event.mark_present_url_token:
+            messages.error(request, _("Invalid url."))
+        elif not request.member or not is_user_registered(request.member, event):
+            messages.error(request, _("You are not registered for this event."))
+        elif event.end < timezone.now():
+            messages.error(request, _("This event has already ended."))
```
JobDoesburg, line 201: What if the event has not started yet?

website/events/views.py
```diff
@@ -179,3 +181,30 @@ def dispatch(self, request, *args, **kwargs):
...
+        else:
```
JobDoesburg, line 202: I think we should add a warning or info message if the user was already marked present

website/events/admin/views.py
```diff
@@ -291,3 +293,16 @@ def get(self, request, pk):
             "Content-Disposition"
         ] = f'attachment; filename="{slugify(event.title)}.csv"'
         return response
+
+
+@method_decorator(staff_member_required, name="dispatch")
+@method_decorator(organiser_only, name="dispatch")
+class EventMarkPresentQR(View):
```
JobDoesburg, line 300: Do we really want this to be rendered server-side and not client-side? (For real a question, I'm not sure myself either)

JobDoesburg (commented):
Nice work!

website/events/admin/views.py
DeD1rk: I like the simplicity of server-side here, it's more maintainable than adding yet another javascript front-end part. It also works practically straight away on any platform, making it easy for users to show/print/save/share the QR code, without needing to write maintain js for all of that.

website/events/views.py
DeD1rk: I'm not sure what to do there either. I can see it being used some time before events start (e.g. starting at 12:15 for lunch lectures), but there's really no sensible amount of time to place a limit at. 

Setting lower bound on the time doesn't matter at all anyway: the url is `secret` until the organizer decides to share it, and if someone is very early using the url, then forbidding it until the event starts still doesn't prevent them from using it as soon as the event starts. So I don't think there's really any reason to add a lower bound.

JobDoesburg (commented):
There's one more thing I would like to think about. Maybe we should make the QR code time-based. So change it every 10 seconds so people can't forward it. What do you think about that?

DeD1rk:
I've considered options like that. The thing with changing the url is that you need to do it very quickly (I think at most 5s) to actually prevent forwarding, and at that point it becomes difficult:
- People will have only seconds to scan a code and get it opened in a browser. In many cases it will have expired between the time it's presented and someone actually sends a request.
- Implementing and maintaining it would be much more complex (js to display it, with polling).
- People would no longer be able to print the QR code.

Last wednesday we figured that making it time-based enough to prevent forwarding would be unmaintainable/feature-creep

JobDoesburg (approved):

JobDoesburg:
We will just merge this and see whether there’s gonna be abuse. If so, we can still change it

PR #2476, opened 2022-09-14
Make psycopg2 an optional dependency

JobDoesburg:
### Summary
Prevents errors on certain systems where psycopg2 can't be installed. psycopg2 is only required for production

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ python-magic = "0.4.27"
 Pillow = "9.1.1"
 Django = "4.0.7"
 django-compressor = "3.1"
-psycopg2 = "2.9.3"
+psycopg2 = {version = "2.9.3", optional = true}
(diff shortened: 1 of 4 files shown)


JobDoesburg:
@pingiun can you fix the make files / CI / release stuff

pyproject.toml
```diff
@@ -45,6 +45,7 @@ qrcode = {version = "^7.3.1", extras = ["pil"]}
 
 [tool.poetry.extras]
 docs = ["recommonmark", "sphinx"]
+production = ["psycopg2"]
```
se-bastiaan, line 48: This isn't only required for production. Our CI tests also use postgres as database. So I really feel like this is not the right name. "Postgres" might be more suitable?

se-bastiaan:
@pingiun you realise tests were failing?

PR #2479, opened 2022-09-16
Add workflow that builds the docker container

pingiun:
I want to use this to automate deploys


pingiun:
Because I just wanna test this, imma force merge

PR #2480, opened 2022-09-16
Switch to docker based deployments

pingiun:
### Summary

Adds docker and nixos config to deploy using docker images

### How to test
Try out the deployment on staging


se-bastiaan:
Do I understand correctly that this defines the following?
 - Our infrastructure is managed through Terraform
 - The server defined in the infrastructure is managed through NixOS config
 - The applications that the NixOS config defines are Docker containers and nginx
 - The concrexit Docker container runs the website

pingiun:
> Do I understand correctly that this defines the following?
> 
> * Our infrastructure is managed through Terraform
> * The server defined in the infrastructure is managed through NixOS config
> * The applications that the NixOS config defines are Docker containers and nginx
> * The concrexit Docker container runs the website

yes

JobDoesburg:
I just thought about it again and I think we shouldn't use nix at all. It might be a nice tool for the job, and this is already way more readable than what we had, but just isn't maintainable, as there are so little people that know anything about nix.

pingiun:
I don't want to go back to ansible

JobDoesburg:
> I don't want to go back to ansible

That's true, and I agree that this concrexit-server.nix is pretty well readable. But isn't it easier to just do it manually?

JobDoesburg:
Nvm maybe lets just keep it this way... The comments in nix help a lot

se-bastiaan:
I like it, quite understandable and seems easy to still do quick deployments like this and change the server config. Highly prefer this over ansible as well.

se-bastiaan (approved):

PR #2502, opened 2022-09-22
Optimize sales app

### LINKED ISSUE #2469: Sales shifts admin changelist_view is very very very slow
### Describe the bug
Sales shifts admin changelist_view is very very very slow

### How to reproduce
1. Go to https://thalia.nu/admin/sales/shift/
2. It is slow

### Expected behaviour
Don't be so slow

JobDoesburg:
Might help with #2469 

### Summary
Rewrites some ugly code that also does quite some speedup. Also does some minor admin UI improvements. Hopefully this is enough of a speedup.

### How to test
1. Go to the shift admin page 
2. Check that it isn't that slow


website/sales/models/order.py
```diff
@@ -129,18 +129,33 @@ class Meta:
         Coalesce(Sum("order_items__amount"), Value(0), output_field=IntegerField())
     )
 
+    _is_free = models.BooleanField(
+        verbose_name=_("is free"),
+        default=False,
+    )
+    _total_amount = PaymentAmountField(
```
DeD1rk, line 136: To be sure that the fields stay up to date, I guess the stored field should also be updated in a post_save signal from OrderItems.
With that done, it should also be fine to remove the `total_amount` annotation property, so there is a single source of truth, efficiency everywhere, and no double code to maintain.

DeD1rk: Oh actually, I see you already do that from `OrderITem.save`, that works. Still then I think it would be preferable to have a single source of truth.

JobDoesburg: We already do a order.save() in the order item save method. 

Sadly, the post_save signal does not give guarantees. For example, deletions don't always call post_save signals. The choice to make `_total_amount` be a cached, last-computed value that _might_ be outdated was deliberate. Querying total_amount however will always return the guaranteed correct value

DeD1rk: Well it's true that there's no strict guarantees this way, but as long as we don't do bulk updates or bulk deletes there are guarantees. And we shouldn't be doing bulk operations anyway (that would also break 'immutable' payables for example).

JobDoesburg: Then still, the current aggregate property remains the most efficient way to actually calculate _total_amount so we can't remove it (or I would remove it and replace it by less efficient Python code)

JobDoesburg: @DeD1rk what do you think?

DeD1rk: Fair enough, there's one strange thing though: the save code uses an in-python calculation instead of just getting the current value of the total amount property

JobDoesburg: I see, probably relates to the comment I removed `# We cannot use self.total_amount as it is a requires a database query and hence will not use any updated values`. I will test it

pingiun (approved):

PR #2522, opened 2022-09-28
Remove obsolete pylint checks

pingiun:
### Summary
These checks have been removed in pylint 2.6


.pylintrc
```diff
@@ -129,7 +73,6 @@ disable=print-statement,
         duplicate-code,
         arguments-differ,
         protected-access,
-        no-self-use,
```
se-bastiaan, line 132: We still want this particular line

se-bastiaan (approved):

PR #2547, opened 2022-10-05
Event admin hotfix v2

ColonelPhantom:
Closes (not an issue created)

### Summary
Creating events still did not work :| this should fix that by moving some tests and re-adding some code that was deemed unnecessary.

### How to test
1. Try to create an event
2. Observe that the page still opens
3. Note that trying to actually create it also works.


DeD1rk (approved):
Nice fix!

Any chance we could do some kind of integration test for this kind of thing?

Update: actually, since cleaning on organisers is not done in the model.clean, the admin api would make it possible to break organizers. That's not really related to this but I guess we should make an issue for it?

DeD1rk:
Also there's a typo in the name of the failing test

pingiun (approved):

PR #2548, opened 2022-10-05
Squash newsletter migrations

### LINKED ISSUE #2462: Running tests is very slow due to newsletter migration
### Describe the change
Running tests takes very long due to the migration `newsletters.0015_embed_images_html`. I'm not sure how this migration works and why it is so slow, but we should probably work around the slowness of this in some way, e.g. enabling a developer to re-use the same database when testing.

### Motivation
When debugging some tests, needing to wait multiple minutes to run each test is very demotivating and makes it impossible to iterate quickly to see if my changes are correct.

### Suggested implementation
Either speed up this migration, find a way to 'skip' it, or use enable using a pre-populated database for testing.

JobDoesburg:
Closes #2462 

### Summary
Squashes some migrations that take long

### How to test
1. Run them


DeD1rk (commented):
Looks about right to me, but did you manually edit this? I think marking all the runpython migrations as elidable and then using the completely generated squashed migration is probably a bit safer (or if that's how you did this, why not just check-in those changes as well)

JobDoesburg:
> Looks about right to me, but did you manually edit this? I think marking all the runpython migrations as elidable and then using the completely generated squashed migration is probably a bit safer (or if that's how you did this, why not just check-in those changes as well)

That's what I did. This just makes sure the history remains valid

se-bastiaan (approved):

PR #2549, opened 2022-10-05
Do not throw away bank accounts that may not be deleted during member data minimisation

### LINKED ISSUE #2539: It is possible to delete bank account for user with open thalia pay payments via data minimisation
### Describe the bug
It is possible to delete bank account for user with open thalia pay payments via data minimisation

### How to reproduce
have a user pay with Thalia Pay
Minimise account data
Try to process batch
No bank account, crash 

### Expected behaviour
Minimizing should not be possible

JobDoesburg:
Closes #2539 

### Summary
Moves bank account minimization to the payments app. This will keep bank accounts 13 months after last usage for collecting a Thalia Pay batch, as is required.

### How to test
1. Do data minimisation
2. Bank accounts will only be removed 13 months after last usage

### DIFF
--- a/website/members/services.py
+++ b/website/members/services.py
@@ -236,7 +236,6 @@ def execute_data_minimisation(dry_run=False, members=None) -> List[Member]:
             profile.birthday = datetime(1900, 1, 1)
             profile.emergency_contact_phone_number = None
             profile.emergency_contact = None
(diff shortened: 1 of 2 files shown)


se-bastiaan (approved):
Better place to minimise this model as well, imo this would be the proper way to do it even if it weren't a problem.

PR #2550, opened 2022-10-05
Fix loading animation properly

### LINKED ISSUE #2543: Add event listener to disable loading animation for images after load
### Describe the change
Add event listener to disable loading animation for images after load. 

### Motivation
It is ugly and also problematic for transparent images

### Current implementation
Currently the animation is **behind** the image, but it should actually be deleted

### Suggested implementation
one global javascript event listener

JobDoesburg:
Closes #2543

### Summary
Adds event listener to delete the pulsing-background class when images load

### How to test
1. Have a transparent image, like a society logo
2. The background will pulse during loading
3. The background will not pulse after loading anymore


pingiun (approved):
Seems good, let's try it

PR #2552, opened 2022-10-06
Git hook confirmation

pingiun:
### Summary
Makes sure admins are certain when they are pushing to master

### How to test
1. Be admin
2. Install the hook
3. Try to push to master
4. See prompt

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-Thalia Website 
+Thalia Website
 ==============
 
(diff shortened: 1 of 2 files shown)


pre-push-hook
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
```
se-bastiaan, line 2: This could go in the .github or scripts folders probably

pingiun: That's fair

se-bastiaan (approved):

PR #2553, opened 2022-10-06
Dont call distinct on order queryset

### LINKED ISSUE #2478: Some deletes in the sales app fail
We are using distinct in some places and those general querysets are probably also used for deletion.

Sentry Issue: [CONCREXIT-G3](https://sentry.io/organizations/thalia/issues/3592435888/?referrer=github_integration)

```
TypeError: Cannot call delete() after .distinct().
(9 additional frame(s) were not displayed)
...
  File "django/contrib/admin/options.py", line 1983, in changelist_view
    response = self.response_action(
  File "django/contrib/admin/options.py", line 1588, in response_action
    response = func(self, request, queryset)
  File "django/contrib/admin/actions.py", line 51, in delete_selected
    modeladmin.delete_queryset(request, queryset)
  File "django/contrib/admin/options.py", line 1230, in delete_queryset
    queryset.delete()
  File "django/db/models/query.py", line 809, in delete
    raise TypeError("Cannot call delete() after .distinct().")
```

JobDoesburg:
Closes #2478

### Summary
Removes  the query set.distinct call for the admin's get_queryset method because we don't need it there, but re-adds the call on the shift field query set 

### How to test
1. Have a shift with 2 managers that you're both part of
2. Create an order from the admin (this won't crash)
3. Try to delete the orders via the changelist admin action

### DIFF
--- a/website/sales/admin/order_admin.py
+++ b/website/sales/admin/order_admin.py
@@ -253,7 +253,7 @@ class OrderAdmin(admin.ModelAdmin):
         elif not request.member.has_perm("sales.override_manager"):
             queryset = queryset.filter(
                 shift__managers__in=request.member.get_member_groups()
-            ).distinct()
+            )
 
         queryset = queryset.select_properties(
             "total_amount", "subtotal", "num_items", "age_restricted"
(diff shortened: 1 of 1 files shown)


se-bastiaan (commented):
This gets an approve when there is a description :p

se-bastiaan (approved):

PR #2579, opened 2022-10-13
Use versions for Docker actions

se-bastiaan:
### Summary
I noticed that Renovate is keeping this file up to date with the latest head of the respective repo's. We probably just want to use the releases instead.


pingiun (approved):
Good idea, the initial version was just copied from an example

PR #2581, opened 2022-10-13
Cache docker builds

pingiun:
Use info from https://github.com/docker/build-push-action/blob/master/docs/advanced/cache.md#registry-cache to cache builds


JobDoesburg (approved):
LGTM

PR #2603, opened 2022-10-16
Face recognition

JobDoesburg:
### Summary
Implements face recognition based on https://github.com/svthalia/face-detect-app/blob/master/face_detection/services/detector.py


JobDoesburg:
still todo: 

- [ ] Authenticate API endpoints, think about scopes etc
- [ ] Set up detector infra and documentation
- [ ] Updating / deleting reference faces and files
- [ ] Manually adding photos you're appearing on / deleting matches
- [ ] Update privacy policy

website/photos/models.py
```diff
@@ -98,6 +115,267 @@ class Meta:
...
+#     )
+#
+#     def __str__(self):
+#         return str(self.member) + " " + _("is on") + " " + str(self.photo)
+#
+#     class Meta:
+#         unique_together = ["photo", "member"]
+
+
+class FaceEncoding(models.Model):
```
se-bastiaan, line 145: Maybe we should put the different models in subfiles here now too

JobDoesburg: Im actually thinking about moving the whole thing to a separate app

se-bastiaan:
Am going to mention this here as well: https://github.com/se-bastiaan/face-detect-api

There are a couple of more things to think about. The approach you are taking now requires an external system to manage the face encodings, while concrexit is the main system here and should choose how these encodings are created. So that means not exposing the unprocessed images, but instead actively sending a request to create these encodings. Concrexit knows when a new album is uploaded, or can periodically do a query to check for missing info. And if the info comes back async you probably want a regular webhook and not expose an API endpoint for it, since it is specifically meant for this task.

The actual detector infra is the part that uses a lot of resources here. It was the part that could kill the face-detect-app as well. So when you upload photos that part should probably not happen on an EC2 instance. This is why I created the poc with api gateway, step functions and lambdas. This highly parallelises the computing, but only uses resources when new encodings are required.

JobDoesburg:
So we discussed the ethical side of the story, since we're implementing something AI like here (lets call this a Data Protection Impact Assessment ;p)


There's one main problem that could arise from this feature. People could use this functionality to search for faces of people other than themselves. On the one site, the current photos are already public, but offering face recognition form our side makes them way more accessible. Having the functionality to get an overview of all student life photos of a person, could potentially be problematic. 

We've discussed several measures: 

- putting restrictions on the amount of reference faces or frequency you can upload / delete: this would be complex and not really fix the problem
- approving reference faces: before they are used infeasible
- cross-checking your reference faces with other's peoples reference faces: in case of false positives this would be nasty. And it forces you to use face recognition if you dont want other people to search for your face
- not implementing this feature at all: we think the added benefit of the feature outweighs the marginal (residual) risk

We will implement the following countermeasures:
- We will log the reference faces that you upload (even if you delete them, we will keep them for a while so we see you have searched for a different person). We will notify users of the fact we're keeping this info for a while for this purpose.
- You can only use this while you're a member. If you're not a member anymore, face recognition is not available to you
- We only search for matches during the time you've been a member (and a little while before to cover the orientation week as well). So if you became a member on September 1 2020, you will not be able to do face recognition for photos before August 1 2020.
- You can have a maximum of 5(?) reference faces active for your account.

JobDoesburg:
We also decided to move this to a separate app completely

JobDoesburg:
> Am going to mention this here as well: https://github.com/se-bastiaan/face-detect-api
> 
> There are a couple of more things to think about. The approach you are taking now requires an external system to manage the face encodings, while concrexit is the main system here and should choose how these encodings are created. So that means not exposing the unprocessed images, but instead actively sending a request to create these encodings. Concrexit knows when a new album is uploaded, or can periodically do a query to check for missing info. And if the info comes back async you probably want a regular webhook and not expose an API endpoint for it, since it is specifically meant for this task.
> 
> The actual detector infra is the part that uses a lot of resources here. It was the part that could kill the face-detect-app as well. So when you upload photos that part should probably not happen on an EC2 instance. This is why I created the poc with api gateway, step functions and lambdas. This highly parallelises the computing, but only uses resources when new encodings are required.

Yes we will do this Sebas ;o I was already planning on this

PR #2608, opened 2022-10-19
Fix bank account deletion

JobDoesburg:
Fixes #2549

### Summary
The revokeoldmandates command was wrong

### How to test
1. Have a lot of bank accounts
2. Verify the correct ones remain

### DIFF
--- a/website/payments/services.py
+++ b/website/payments/services.py
@@ -212,7 +212,7 @@ def execute_data_minimisation(dry_run=False):
     # Delete bank accounts that are not valid anymore, and have not been used in the last 13 months
     # (13 months is the required time we need to keep the mandates for)
     queryset_bankaccounts = BankAccount.objects.all()
-    queryset_bankaccounts = queryset_bankaccounts.exclude(
+    queryset_bankaccounts = queryset_bankaccounts.filter(
         valid_until__lt=timezone.now()
     )  # Keep valid bank accounts
     queryset_bankaccounts = queryset_bankaccounts.exclude(  # Also keep bank accounts that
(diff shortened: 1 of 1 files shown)


website/payments/services.py
```diff
@@ -209,12 +209,22 @@ def execute_data_minimisation(dry_run=False):
...
-        Q(owner__paid_payment_set__batch__isnull=True)
-        | Q(owner__paid_payment_set__batch__processed=False)
+    # Delete bank accounts that are not valid anymore, and have not been used in the last 13 months
+    # (13 months is the required time we need to keep the mandates for)
+    queryset_bankaccounts = BankAccount.objects.all()
+    queryset_bankaccounts = queryset_bankaccounts.exclude(
+        valid_until__lt=timezone.now()
+    )  # Keep valid bank accounts
+    queryset_bankaccounts = queryset_bankaccounts.exclude(  # Also keep bank accounts that
+        Q(
```
DeD1rk, line 219: This doesn't have to be in a Q object

website/payments/services.py
```diff
@@ -209,12 +209,22 @@ def execute_data_minimisation(dry_run=False):
...
+            owner__paid_payment_set__type=Payment.TPAY
+        ),  # are used for Thalia Pay payments, AND
+        Q(owner__paid_payment_set__batch__isnull=True)  # have no batch, OR
```
DeD1rk, line 222:
```suggestion
        Q(owner__paid_payment_set__batch__isnull=True)  # have a payment that is in no batch, OR
```

DeD1rk: Clarification that the foreignkey reverse lookups work with "any" might help

pingiun (approved):

PR #2613, opened 2022-10-24
Set new banner for photos page

pingiun:
### Summary
Set new banner for photos page

### How to test
1. Look at photos page
2. See new banner


pingiun:
Yea this is not a good solution

PR #2618, opened 2022-10-26
Add more info to readme to help beginners

NienkeWessel:
### Summary
The README now contains extra information that should be helpful for beginners. This makes the README more self-contained. It also moved some information that is less relevant to beginners to the end. 

### How to test
Read the README.md and see if it is clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Thalia Website
-==============
+# Thalia Website
 
 [![Linting and Testing](https://github.com/svthalia/concrexit/workflows/Linting%20and%20Testing/badge.svg)](https://github.com/svthalia/concrexit/actions)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,8 +84,20 @@ To update these files run `make apidocs`
 
 To render the docs to HTML run `make docs`
 
-For admins
-----------
+
+
+## Things that might be relevant for more advanced users/developers
```
se-bastiaan, line 89: We run a check for black and run all tests when anyone opens a PR. Doesn't that make this relevant for everyone? People will skip this and when the checks fail I am afraid they will have no clue on what to do.

pingiun:
I changed some more things because some parts of the README were outdated too. But this is helpful, thanks!

pingiun (approved):

PR #2619, opened 2022-10-26
Opt-in photos permissions.

### LINKED ISSUE #2592: Opt-out permission for posting event pictures to Instagram
**Problem** 
We (Paparazcie) want the Thalia Instagram to reflect the vibe in Thalia more, and this is really hard without pictures of events. It is very frustrating to contact everyone in event pictures to get permission to post event pictures that they are on. In our experience, people don't object to posting the pictures, but are slow to respond. 

**Possible solution** 
It would be nice to have a checkbox in your Thalia profile to opt-out of consent for posting pictures with you in it, and to have this choice be visible in the attendance list of an event. This way, Paparazcie can independently check if people consent to their picture being posted without having to approach everyone in the picture per picture. 

**Alternative solution** 
We considered having this checkbox when registering for an event, but this seems like more hassle.

LisanneWeidmann:
Closes #2592 
Opt-out permission for posting event pictures to Instagram.

### Summary
Should make it easier for Paparazcie to see if people consent to their pictures appearing on social media. I made a checkbox.

### How to test
1. Check if you can click a checkbox for opt-in social media photos.


website/members/models/profile.py
```diff
@@ -245,6 +245,17 @@ class Profile(models.Model):
...
 
+    # Begin Lisanne
+    photo_permissions = models.BooleanField(
+        verbose_name=_("Opt-in social media pictures"),
+        help_text=_(
+            "Consent to all your photos being shared across social media platforms."
+        ),
+        blank=False,
+        null=True,
+    )
```
se-bastiaan, line 256: Have you considered that this would allow access to the full user profile for the Paparazcie unless we make this a part of some special page? I already mentioned this in the issue because that would be difficult to do.

It's probably easier to create a allowlist model like the blacklist model we have for payments: https://github.com/svthalia/concrexit/blob/master/website/payments/models.py#L111. Then the paparazcie can have view permissions and always see this info.

The only 'hard' part there is making it possible for users to change their preferences thus requiring a larger profile edit page with a part that is in an extra form field.

DeD1rk:
Closing as now that we have ThabloidUser with proper settings, this really should be a similar approach. And rebasing would be a pain

PR #2638, opened 2022-11-02
Make profile fields blankable

### LINKED ISSUE #2629: Make profile fields blankable
### Describe the change
Make profile fields blankable

### Motivation
- When a profile is minimised, it's perfectly fine that the fields are null
- It should be possible to create empty profiles

### Current implementation
Non-nullable

pingiun:
Closes #2629 

### Summary
Make profile fields blankable

### How to test
Multiple flows should be tested:

# Admin flow
1. Have user with profile.
2. Admin removes address fields
3. This is allowed

# User flow
1. Be user with profile
2. Try to remove address field
3. This is not allowed

# Dataminimisation
1. Have old user with profile
2. User is minimised by dataminimisation management command
3. Old profile now has blank fields everywhere and is still editable by the admin


website/members/services.py
```diff
@@ -229,12 +229,12 @@ def execute_data_minimisation(dry_run=False, members=None) -> list[Member]:
...
+            profile.address_street = None
             profile.address_street2 = None
-            profile.address_postal_code = "<removed>"
-            profile.address_city = "<removed>"
-            profile.address_country = "NL"
-            profile.birthday = datetime(1900, 1, 1)
+            profile.address_postal_code = None
+            profile.address_city = None
+            profile.address_country = None
+            profile.birthday = None
```
JobDoesburg, line 237: In website/members/models/profile.py line 276-277 the `is_minimised` property should also be rewritten now

JobDoesburg (changes requested):

website/members/services.py
JobDoesburg: > In website/members/models/profile.py line 276-277 the `is_minimised` property should also be rewritten now

Or maybe we should introduce a `is_minimised` field, and clean the model to only allow null fields if `is_minised` is `True`

website/members/models/profile.py
```diff
@@ -312,6 +316,15 @@ def clean(self):
         super().clean()
         errors = {}
 
+        if not self.is_minimized and not (
```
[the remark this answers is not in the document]
pingiun: No

JobDoesburg (approved):

PR #2639, opened 2022-11-02
Add createmember management command

pingiun:
This builds on #2638 

### Summary
Add createmember management command

This makes it so new programmers don't have to manually create a member again (if they remember the right `make` command)

### How to test
1. Run `make member`
2. Check if the new member can access the profile page

### DIFF
--- a/website/members/management/commands/createmember.py
+++ b/website/members/management/commands/createmember.py
@@ -1,0 +1,190 @@
+import getpass
+import os
+import sys
+
+from django.contrib.auth.management import get_default_username
+from django.contrib.auth.management.commands.createsuperuser import PASSWORD_FIELD
+from django.contrib.auth.management.commands.createsuperuser import (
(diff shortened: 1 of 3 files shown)


website/members/management/commands/createmember.py
```diff
@@ -0,0 +1,190 @@
...
+                database
+            ).create_superuser(**user_data)
+            # Changed from django/contrib/auth/management/commands/createsuperuser.py
+            Profile.objects.create(user=user)
+            Membership.objects.create(
+                user=user, type=Membership.MEMBER, since=timezone.now()
+            )
+
+            if options["verbosity"] >= 1:
+                self.stdout.write("Superuser created successfully.")
```
DeD1rk, line 179: If both createsuperuser and createmember are available, I think the wording should be consistent about this being about members, not just superusers

JobDoesburg (approved):

PR #2644, opened 2022-11-04
Change production envvars to match release branch

ColonelPhantom:
I found that on the release branch these were modified, but not on master. It seems that these should be updated on master as well, so that the next version will not have incorrect values for these.


pingiun (approved):
Thanks

PR #2669, opened 2022-11-16
Implement notifications for more or less registration spots

### LINKED ISSUE #2265: Notification when moving from queue to participants list
### Describe the bug
When number of spots was increased and participants move from queue to participants list, they do not get notification. The system works fine in case of cancellation.

### How to reproduce
1. Create event 
2. Participants are added to the participants list and queue 
3. Number of places is increased
4. People in queue move to participants list
5. Notification about it not sent

### Expected behaviour
Thalia member should receive a notification email about their participation in event.

ColonelPhantom:
Closes #2265 

### Summary
This PR adds email notifications that are sent when the number of possible registrations for an event changes and people are added or kicked out because of this.

### How to test
I'll be honest, I have no clue since it involves the mail subsystem. This also means I haven't been able to test this :cry: 

Furthermore, this currently creates a circular dependency because emails.py refers to EventRegistration which refers to Event which refers to emails.py etc etc etc, and I'm not sure what the best solution is (split emails.py? move some code around?)


se-bastiaan:
You should be able to test this locally because django prints the emails to the terminal.

Have you considered implementing a warning instead?

ColonelPhantom:
> You should be able to test this locally because django prints the emails to the terminal.

Thanks!

> Have you considered implementing a warning instead?

Can't say I have; I thought automatically sending emails would be nicer. Although the problem you mention in the issue is definitely real, but I think by sending notifications both ways this PR already kind of avoids that. 

Perhaps a warning combined with a link to send those notifications could also be a solution. I would prefer to not make organizers send all the emails by hand.

ColonelPhantom:
We have decided to automatically send an email when there are more spots.

When the limit is decreased, we will display a warning with a `mailto:` link so that the organizer can inform the affected members and explain the reason.

ColonelPhantom:
Okay, I haven't figured out how to unit test this precisely; creating a request that passes all validation is difficult, creating a form by itself is not possible. However it would be good to create a unit test for this functionality.

ColonelPhantom:
I'm still not sure how to create the proper request in a test; if someone wants to contribute a test that would be great but otherwise I think not automatically testing this isn't a huge deal. I also fixed it to use BCC's to prevent data leaks :)

pingiun (approved):

PR #2674, opened 2022-11-17
Non nix env files

pingiun:
### Summary
Automatically place env files from the repo on the server on staging deploy

### How to test
1. Merge this
2. Check the github actions output
3. The rsync task should have run succesfully


pingiun:
Secrets will still be managed by nix btw, this is because the decryption step is done by nix-age and that is a useful tool

PR #2717, opened 2022-12-07
Fix summary and exam download

### LINKED ISSUE #2625: Summary download not working
### Describe the bug
Some summaries (for example Data Mining 2019-2020 Practice Exam Midterms) are not downloaded when clicked. They also do not work when trying to view them through the Site Administration. Pressing these buttons only leads to a new empty tab. 

Some summaries still work however.

### How to reproduce
1. Go to 'education'
2. Scroll down to 'data mining'
3. Click on 'data mining'
4. Click on '2019-2020'
5. Click on 'Practice Exam Midterm'

### Expected behaviour
Clicking this button should download the associated file.

JobDoesburg:
Closes #2625 

### Summary
We used to use `send_file`, but refactored and broke it in #2355

### DIFF
--- a/website/education/views.py
+++ b/website/education/views.py
@@ -129,8 +129,7 @@ class ExamDetailView(DetailView):
         exam.save()
 
         ext = os.path.splitext(exam.file.name)[1]
-        filename = f"{exam.course.name}-exam{exam.year}{ext}"
-        return redirect(get_media_url(exam.file, filename))
+        return redirect(get_media_url(exam.file, True))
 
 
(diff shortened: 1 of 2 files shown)


JobDoesburg:
What used to be here is wrong definitely

JobDoesburg:
@pingiun do you know what's wrong. The problem is that we get an **nginx** 403 if we try to download a summary or exam...

pingiun:
which url does it redirect to with this fix?

PR #2726, opened 2022-12-14
Add liked photos API

### LINKED ISSUE #2716: Liked photos API endpoint
To add an overview of liked photos to ThaliApp, we need a new endpoint for liked photos.

I think it would be best to have `api/v2/photos/photos/` with `liked` boolean GET filter. It will need to do some filtering to prevent photos that are not published in an album from being returned.

LisanneWeidmann:
Closes #2716

### Summary
Adds an endpoint to view the current user's liked photos.

### How to test
1. Go to localhost:8000/api/v2/photos/photos/liked/
  1.1 If you did not like any photos yet, the returned list should be empty. 
  1.2 Otherwise you should see a list of liked photos.
2. Go to http://localhost:8000/members/photos/ and like a photo.
3. Go to localhost:8000/api/v2/photos/photos/liked/
4. The photo that you just liked should now be in the list.

### DIFF
--- a/website/photos/api/v2/urls.py
+++ b/website/photos/api/v2/urls.py
@@ -25,7 +25,7 @@ urlpatterns = [
                     "photos/<int:pk>/like/", PhotoLikeView.as_view(), name="photo-like"
                 ),
                 path(
-                    "liked-photos/", LikedPhotosListView.as_view(), name="liked-photos"
+                    "photos/liked/", LikedPhotosListView.as_view(), name="liked-photos"
                 ),
             ]
         ),
--- a/website/photos/api/v2/urls.py
+++ b/website/photos/api/v2/urls.py
@@ -1,7 +1,12 @@
 """Photos app API v2 urls."""
 from django.urls import include, path
 
-from photos.api.v2.views import AlbumDetailView, AlbumListView, PhotoLikeView
+from photos.api.v2.views import (
+    AlbumDetailView,
+    AlbumListView,
+    LikedPhotosListView,
+    PhotoLikeView,
+)
 
 app_name = "photos"
 
@@ -18,6 +23,9 @@ urlpatterns = [
                 ),
                 path(
                     "photos/<int:pk>/like/", PhotoLikeView.as_view(), name="photo-like"
+                ),
+                path(
+                    "liked-photos/", LikedPhotosListView.as_view(), name="liked-photos"
                 ),
             ]
         ),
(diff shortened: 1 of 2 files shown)


DeD1rk (commented):
I like it. One thing I am unsure about is whether it would be worth the effort, considering e.g. usage by face-detect or other new functionality, such as ranking photos by likes, to make this to a general `/photos/photos/` endpoint with `?liked=yes` filter.

That would make it easy to later add other options like ordering by num_likes

website/photos/api/v2/urls.py
```diff
@@ -19,6 +24,9 @@
                 path(
                     "photos/<int:pk>/like/", PhotoLikeView.as_view(), name="photo-like"
                 ),
+                path(
+                    "liked-photos/", LikedPhotosListView.as_view(), name="liked-photos"
```
JobDoesburg, line 28:
```suggestion
                    "photos/liked/", LikedPhotosListView.as_view(), name="liked-photos"
```

website/photos/api/v2/views.py
```diff
@@ -51,6 +55,36 @@ def get_queryset(self):
...
+    """Returns the details the liked album."""
+
+    serializer_class = PhotoListSerializer
+    permission_classes = [
+        IsAuthenticatedOrTokenHasScope,
+    ]
+    required_scopes = ["photos:read"]
+
+    def get_queryset(self):
+        if self.request.member:
```
JobDoesburg, line 68: I would not put this check here, but in a different method, such as `dispatch`, `get` or maybe use a permission class for this. Also, non-members should not get a 400, because it's not a bad request in itself (it's not malformed or something like that). I think a 403 would be better

JobDoesburg (changes requested):

website/photos/api/v2/views.py
```diff
@@ -51,6 +55,35 @@ def get_queryset(self):
...
+    """Returns the details the liked album."""
+
+    serializer_class = PhotoListSerializer
+    permission_classes = [
+        IsAuthenticatedOrTokenHasScope,
+    ]
+    required_scopes = ["photos:read"]
+
+    def get(self, request, *args, **kwargs):
+        if not self.request.member:
```
JobDoesburg, line 68: Don't we have a permission class for this?

JobDoesburg (approved):

website/photos/api/v2/views.py
DeD1rk: Looks like we actually don't. Currently all API endpoint use `IsAuthenticatedOrTokenHasScope`. I think we should probably change that though.
Still, in this specific case a permission class (only) doesn't really make sense, as the check is a technical requirement, not really about permissions.

PR #2728, opened 2022-12-14
Move menu out to individual apps

pingiun:
### Summary
Now menu items can be added by putting extra stuff in the appconfig


website/activemembers/apps.py
```diff
@@ -13,3 +13,28 @@ def ready(self):
...
         from . import signals
+
+    def menu_items(self):
+        return {
+            "categories": [{"name": "association", "title": "Association", "key": 1}],
+            "items": [
+                {
+                    "category": "association",
+                    "title": "Board",
+                    "viewname": "activemembers:boards",
```
JobDoesburg, line 24: For one, its not view name but url name. 
Also, I think it's better to already include the URL itself in here with a reverse or reverse_lazy (if that's possible, at least)

JobDoesburg (changes requested):
Very nice

website/activemembers/apps.py
pingiun: Django calls this viewname themselves: https://docs.djangoproject.com/en/4.1/ref/urlresolvers/#django.urls.reverse

I think reversing it beforehand technically removes some flexibility in the menu creating. I think it would be fine though

JobDoesburg: > Django calls this viewname themselves: https://docs.djangoproject.com/en/4.1/ref/urlresolvers/#django.urls.reverse


Weird haha, but sure 


> I think reversing it beforehand technically removes some flexibility in the menu creating. I think it would be fine though

If I would want to add a URL to the menu to an external page, that's currently not possible without creating a redirect view for it.

pingiun: > If I would want to add a URL to the menu to an external page, that's currently not possible without creating a redirect view for it.

https://github.com/svthalia/concrexit/blob/e98a4809442e92762e3315a6191576f434343a33/website/members/apps.py#L26-L32

website/thaliawebsite/templatetags/menu.py
```diff
@@ -1,12 +1,50 @@
 """Provides a template handler that renders the menu."""
+from collections import defaultdict
+from functools import cache
+
 from django import template
+from django.apps import apps
 from django.urls import reverse
 
-from ..menus import MAIN_MENU
```
DeD1rk, line 5: This file hasn't been removed yet

website/activemembers/apps.py
ColonelPhantom: Another solution would just be to add `"noreferer": True` to the item. I personally think that that would be the cleanest solution.

ColonelPhantom (approved):

JobDoesburg (approved):

PR #2734, opened 2022-12-21
Do announcements in the proper app

### LINKED ISSUE #2635: Look at KiOui/TOSTI filters for adding info banners to pages
### Describe the change
Right now, the base template contains some hardcoded banners, like "you're not a member anymore". That logic shouldn't be in the `thaliawebsite` app but in the members app.

TOSTI solves this in a nice way with `Filters` (stolen from, yes, Wordpress): See https://github.com/KiOui/TOSTI/blob/master/website/tosti/filter.py 
https://github.com/KiOui/TOSTI/blob/7008c332758f6e89716cdf8f9a7220aea36b877a/website/users/views.py#L37-L44 and https://github.com/KiOui/TOSTI/blob/7008c332758f6e89716cdf8f9a7220aea36b877a/website/thaliedje/apps.py#L14-L25

### Motivation
Separation of concerns

pingiun:
Closes #2635 

### Summary
Banners still work but now in the app where the check can be done

### DIFF
--- a/website/announcements/context_processors.py
+++ b/website/announcements/context_processors.py
@@ -17,4 +17,8 @@ def announcements(request):
         for a in Announcement.objects.all()
         if a.is_visible and a.pk not in closed_announcements
     ]
-    return {"announcements": announcements_list}
+    persistent_announcements = getattr(request, "_announcements", [])
+    return {
+        "announcements": announcements_list,
+        "persistent_announcements": persistent_announcements,
+    }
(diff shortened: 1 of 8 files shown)


ColonelPhantom (approved):

DeD1rk:
@pingiun This broke some APIs, `MembersConfig.announcements` accesses `request.member` (a lazy property) which, for API calls, will only resolve correctly later, as the `request.user` is only set correctly (not to AnonymousUser) within the API views, not during middleware evaluation.
The result is that request.member is None for all API calls, breaking the payments API and some filters.

DeD1rk:
Of course the real problem is that authentication is done too late but there seems to be no good fix for that

PR #2781, opened 2023-01-25
Registration without membership

### LINKED ISSUE #2774: Allow people without active membership to register for an event
### Problem
Currently, when creating an event, there is no option to allow people without an active membership to sign up for that event. There is an option to manually add non-members to add the event via the backend, but for alumni events (where mostly people without active membership sign up) this is a pretty tedious process, especially since there are currently two ways of signing up (via mail and via the event).

### Solution
A checkbox in the event backend that, when checked, allows website visitors to register themselves while not having an active membership. A requisite is that you do own a Thalia account, AKA have been a member in the past.

### Motivation
This would make registering for alumni events easier and more centralized. Other than that, being able to see who else is registered might work motivating for alumni to register themselves

### Alternatives
The alternative is the current situation: Mailing to the committee. In our opinion not desired.

### Additional context
See my other feature request regarding the alumni committee.

ColonelPhantom:
Closes #2774 

### Summary
I have added a toggle to events to enable users without memberships to register for them.

### How to test
1. Create an ex-member (or non-member) user
2. Create an event with the toggle checked
3. Observe that the ex/non-member user can register for this event, but not for any others.
4. Check that this works correctly even if data minimization has been applied to the user (@pingiun)


website/events/models/event_registration.py
```diff
@@ -33,7 +33,6 @@ class EventRegistration(models.Model):
         models.CASCADE,
         blank=True,
         null=True,
-        limit_choices_to=registration_member_choices_limit,
```
ColonelPhantom, line 36: I did not remove the relevant filter, because it is still used in makefixtures. I'd prefer fixtures to not do weird things here (such as register for events without a membership while this is not allowed for the event in question).

website/events/services.py
```diff
@@ -214,7 +214,7 @@ def event_permissions(member, event, name=None, registration_prefetch=False):
             event.registration_allowed
             or (event.optional_registration_allowed and not event.registration_required)
         )
-        and (name or member.can_attend_events)
+        and (name or member.can_attend_events or event.registration_without_membership)
```
JobDoesburg, line 217: We only need to consider the case that someone is blocked from attending events

website/members/models/member.py
```diff
@@ -168,6 +168,13 @@ def can_attend_events(self):
             and self.current_membership is not None
         )
 
+    @property
+    def can_attend_alumni_events(self):
```
JobDoesburg, line 172: Looks good, I only wouldn't call this `alumni events` as that sounds quite restricting. I would call it `can_attend_events_without_membership`

DeD1rk:
Could we maybe get a test for this? 👉👈🥺

JobDoesburg (approved):

PR #2782, opened 2023-01-25
Add "oldmembers" mailinglist

### LINKED ISSUE #2773: 'Thalia alumni' automated mailing list
### Problem
There currently is no automated mailing list for 'thalia alumni' AKA people that were once Thalia member but are not anymore. 

### Solution
An automated **opt-out** mailing list linked to everyone that is a _former Thalia member_. This list should _exclude_ current members, current benefactors, and former benefactors. It should _include_ honorary members.

### Motivation
Such a mailing list would be very useful for the alumni committee to reach out to people to promote events.

### Alternatives
The current alternative is a WhatsApp 'mailing list' that's an absolute tumor in my phone.

### Additional context
See my other feature request (#2774) regarding the alumni committee.

ColonelPhantom:
Closes #2773 

### Summary
This adds a mailing list targeting alumni as requested. For now it is a draft because the PR is not fully complete, not tested at all, and the exact requirements are not 100% clear. @ricotewechel please let me know if the following logic matches what you want or not:
```python
m.current_membership.type != Membership.MEMBER and m.has_been_member()
```
where this condition is evaluated on all members `m`. This covers everyone who has once been a member, but is not anymore, for example users who go member->benefactor or member->honorary, or just member->nothing.
### How to test
NOT DONE YET


website/mailinglists/services.py
```diff
@@ -86,6 +93,14 @@ def get_automatic_lists():
             ),
             "moderated": True,
         },
+        {
+            "name": "alumni",
```
JobDoesburg, line 97: the alias `alumni@thalia.nu` already exists and is used by the alumni committee. I would call the list `oldmembers`

JobDoesburg (changes requested):

JobDoesburg:
> ```python
> if m.current_membership_type == Membership.HONORARY
>     or (m.has_been_member() and not m.has_active_membership())
> ```
> 
> where this condition is evaluated on all members `m`. I believe that `has_been_member` does include people who go like member->benefactor>ex-member.
> 


I've discussed it with rico and it should indeed include only people that have ever been a `member` (so not just benefactor, but member → benefactor is fine). The current members should be filtered out, but the current benefactors/honorarymembers that have been a member previously, should be included. 

Finally, we should also make a opt-out for this, but that's for a later issue I think

website/mailinglists/services.py
```diff
@@ -55,6 +55,12 @@ def get_automatic_lists():
     )
     mentors = _get_members_email_addresses([x.member for x in active_mentorships])
 
+    oldmembers = [
+        m
+        for m in Member.objects.all()
+        if m.current_membership.type != Membership.MEMBER and m.has_been_member()
```
ColonelPhantom, line 61: As Job mentioned, this will likely throw an error when there are Members without a current_membership.

ColonelPhantom:
I have updated with a fix for the non-member issue, and added an opt-out flag to the profile.

We may want to hide the flag for current members (?) and/or enable people to set the flag without logging in (if they lost their credentials; although the Forgot Password button should be linked to the same address receiving the mails, so everyone who receives the emails should be able to access their account.)

JobDoesburg (approved):
This does still need a view for easy deregistration if you don't know your password anymore

ColonelPhantom:
> This does still need a view for easy deregistration if you don't know your password anymore

I'm still not sure if this is actually necessary; the forgot password functionality is inherently linked to the same email where you receive the emails. So if you can read the emails, you can reset your password and then unsubscribe. It's a bit more work but should be fairly discoverable, especially if the `edit-profile` link is included (click -> see login page -> forgot password -> enter email) in the emails (although I presume we can't really automate that, nor rely on it being included manually).

That said, less hurdles for unsubscribing is of course better. But I'm not entirely convinced the current barrier is too high, either.

JobDoesburg:
Sure let's keep it like this for now

Issue #988, opened 2020-03-18
Deliver sliders or other large images as webp


thaliawww-service:
In GitLab by @pingiun on Jan 15, 2020, 20:56

### Why?
According to [Google PageSpeed](https://developers.google.com/speed/pagespeed/insights/?hl=nl&url=thalia.nu&tab=mobile), we can speed up our site by a lot by converting slider images to webp.

### Current implementation
Large pictures are all served as png, this is a good image format for things that need to be compressed lossless.

### Suggested implementation
Supply images as `<picture>` [with multiple sources](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Use_modern_image_formats_boldly)

thaliawww-service:
In GitLab by @se-bastiaan on Jan 15, 2020, 22:35

So what do you want to do exactly? Do you want to convert them on the fly on create new images for every single slide on disk? If you want to save the files to disk that could take some space if you want to serve both. And I don't think this makes the image loading stuff ~"easy and fun" with it being quite difficult already. At least, the drawback of complexity vs (supposed) loading time

I do think that increasing the JPG image quality that we save may help, because that's mostly the reason that these images are PNG: the JPEGs are horrible quality. I have changed the images to JPEG now and the warning in PageSpeed is goner. The one for image caching though? Maybe we should add that to the nginx config.

thaliawww-service:
In GitLab by @se-bastiaan on Jan 15, 2020, 22:38

If Safari would support webp I would be in favour of converting every single image to webp, but unfortunately that is not the case: https://caniuse.com/#feat=webp. Even the last non-Chromium version of Edge has support.

2016 -> https://www.cnet.com/news/apple-ios-macos-tests-googles-webp-graphics-to-speed-up-web/

se-bastiaan:
Safari 14 adds webp support, but only on macOS Big Sur and later :|

se-bastiaan:
Note that React Native also has no WebP support without extra deps: https://reactnative.dev/docs/image#gif-and-webp-support-on-android

pingiun:
This is now feasible: https://caniuse.com/?search=webp

PR #2794, opened 2023-02-01
Django thumbnails

### LINKED ISSUE #2659: Find a faster way to request signatures from S3 storage objects
Staging is very slow. It takes seconds to get signatures for S3 objects. And our SSD's are getting too expensive

This should be faster, so we can move to S3.

### Motivation
Money and server load

JobDoesburg:
Closes #2659 

### Summary
Rewrite our complete media and thumbnailing system to Django thumbnails.

We used to have a _lazy_ thumbnailing system. That is: whenever a thumbnail was inserted, we inserted an image source with a route to a thumbnailer view that takes the path of the media to be thumbnailed as an argument. Then, only when that view was loaded, a thumbnail was retrieved (if it already exists) or created. This was a nice approach, as it keeps rendering our pages relatively fast: when loading a photo album with 50 images, this approach makes the initial page load fast, and make the thumbnails appear later.

However, when moving to an AWS S3 storage backend, this didn't seem to work anymore, as retrieving the files from S3 for each thumbnail request turns out to be pretty slow (unverified, but probably because botos3 can't parallelize the requests or something).

This PR completely switches our thumbnailing system. Thumbnails are now generated at the time they are requested. Note: this will make initial page load of an album page slower! But, hopefully, it is better compatible with S3 storage backends, so ultimately it will be faster...

Note that it is possible to pregenerate thumbnails on saving the album, as to prevent this problem!

### How to test
Thumbnails are used in:
- activemembers (MemberGroup)
- members (Profile)
- merchandise (MerchandiseItem)
- partners (multiple models)
- photos (Photo)

Furthermore, announcements (Slides) also contain images, but we never thumbnail those so I didn't implement it there


JobDoesburg:
Django storages botos3 will give us signed urls for AWS, so we can (and should) get rid of private media hosting

JobDoesburg:
Can also close #988

JobDoesburg:
This only needs migrations to populate the thumbnails.source table, as this doesn't happen automatically by the app

JobDoesburg:
Decided that it writing migrations for this sucks, so I implemented a fix for it in the thumbnailing app. Currently waiting for https://github.com/ui/django-thumbnails/pull/123 
Until then, we will need to write our own migrations. Let's give it some time, we can always still do it.

We also need to check our actual thumbnail sizes, and maybe rephrase small to small_300x300 to make sure we can always resize.

JobDoesburg:
TODO:
- check the actual thumbnail sizes being used everywhere
- check the announcements app

website/utils/media/services.py
```diff
@@ -64,24 +53,16 @@ def get_thumbnail_url(file, size, fit=True):
...
-        "name": name,
-        "thumb_path": f"thumbnails/{size_fit}/{name}",
-        "serve_path": f"thumbnails/{size_fit}/{name}",
-        "storage": f"{storage.__class__.__module__}.{storage.__class__.__name__}",
-    }
+    if isinstance(file, ThumbnailedImageFile):
+        if size == "small":
+            return get_media_url(file.thumbnails.small)
+        if size == "medium":
+            return get_media_url(file.thumbnails.medium)
```
DeD1rk, line 64: There are other sizes defined in settings, why aren't those in here too?

DeD1rk: Update: the others are used (in ThumbnailSerializers in the API), so I guess they should be used.

ColonelPhantom:
One thing I did notice is that pages referring to invalid images now fail to load, instead of failing to load just the images.

But I guess that that's not too big of an issue.

Issue #2753, opened 2023-01-11
Don't register event admin inlines in the events app, move to promo and pizzas app


JobDoesburg:
### Describe the change
The events app shouldn't depend on the pizzas and promo app, but the other way around. 

### Motivation
Better modularity of apps

### Suggested implementation
Unregister the events admin in the pizzas app, and reregister with the inline

PR #2796, opened 2023-02-02
Move Event.slide to Slide.event

### LINKED ISSUE #2759: Make announcements depend on events instead of the other way around
### Describe the change
Make announcements depend on events instead of the other way around

### Motivation
Better modularity of apps

### Current implementation
Events can have a slide

### Suggested implementation
Slides can be about an event

### Additional context
Maybe make a class `EventAnnouncements` in a separate sub app for even better modularity

DeD1rk:
Closes #2759.

### Summary
This moves a ForeignKey to a OneToOne in the opposite direction.
After this, events only depends on announcements by including `SlideSerializer` in the event serializers.

### How to test
1. Check that the event screen still shows the right slide header if that was set before migrating.
2. Check that the events API still works.


JobDoesburg:
> After this, events only depends on announcements by including SlideSerializer in the event serializers.

This is still problematic if you ask me...

DeD1rk:
I mean I can simply remove it as it's not really used anyway in ThaliApp, but doing that is probably not possible for all of the inter-app-dependencies changes

website/announcements/templates/events/event.html
```diff
@@ -0,0 +1,18 @@
+{% extends "events/event.html" %}
```
pingiun, line 1: This makes the order of the apps important right? I think you should add a comment about that in the settings.py

pingiun (changes requested):

website/announcements/templates/events/event.html
DeD1rk: That's right. I'll add a note. Once all of the unnecessary dependencies between apps are resolved, there should be a topological ordering or apps.

pingiun:
The event admin should also still include the slide option, as event organisers probably expect to find it there

JobDoesburg:
> The event admin should also still include the slide option, as event organisers probably expect to find it there

That could be implemented indeed (see also #2753), but I actually don't think event organizers think about this as the sliders are not really created by them, but by the promocie

DeD1rk:
I added the inline. There will be a bit of uggliness though when also re-registering the EventAdmin from other apps (at least pushnotifications, but probably also food, sales, promotion): 

To subclass `EventAdmin`, we'd like to import it from `event.admin.event`. However, that version of the class stays the same even when re-registering one in announcements and/or pushnotifications. 

So if you try to override the original EventAdmin in two places, only one of the changes will stay. To prevent that, it looks like I'll have to e.g. check from `pushnotifications/admin` whether `announcements` is installed, and if so, import and extend `announcements.admin.EventAdmin`. Otherwise, import `events.admin.event.EventAdmin`.

That works but becomes quite confusing. Sadly the registry (`AdminSite._registry`) is not really exposed. I guess it should be possible to access it anyway (and take the currently registered class from there, rather than importing it) but that's also hacky.

website/events/api/v2/admin/serializers/event.py
```diff
@@ -30,17 +28,13 @@ def to_internal_value(self, data):
         self.fields["organisers"] = serializers.PrimaryKeyRelatedField(
             queryset=MemberGroup.active_objects.all()
         )
-        self.fields["slide"] = serializers.PrimaryKeyRelatedField(
-            queryset=Slide.objects.all()
```
ColonelPhantom, line 34: Do we want to remove slides from the API? Or do we still want them, by fetching them from the onetoonekey?

JobDoesburg: I would propose removing them from the API if it doesn't break stuff

pingiun (approved):

PR #2808, opened 2023-02-08
fix missing organiser email

### LINKED ISSUE #2802: Invited from queue email missing some stuff
### Describe the bug
image

### How to reproduce
1. Be in queue
2. Get invited (by people deregistering)

Pokemaster69:
Closes #2802

### Summary
Remove the no longer existing organiser_name and organiser_mail from the email template
create list of organiser emails and email lists filtering out organisers without an email-address
Add variable with comma seperate list for email template

### How to test
Create an event (with 1 open registration)
register 1 user and put another user in the queue
deregister original user
observer the email in terminal and see it no longer says "you can reach   at  .

### DIFF
--- a/website/events/emails.py
+++ b/website/events/emails.py
@@ -25,13 +25,19 @@ def notify_first_waiting(event):
         subject = _("[THALIA] Notification about your registration for '{}'").format(
             event.title
         )
+
+        organiser_emails = [
+            organiser.contact_address
+            for organiser in event.organisers.all()
+            if organiser.contact_address is not None
+        ]
         text_message = text_template.render(
             {
                 "event": event,
                 "registration": first_waiting,
                 "name": first_waiting.name or first_waiting.member.first_name,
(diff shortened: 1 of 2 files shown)


website/events/emails.py
```diff
@@ -25,12 +25,19 @@ def notify_first_waiting(event):
         subject = _("[THALIA] Notification about your registration for '{}'").format(
             event.title
         )
+
```
JobDoesburg, line 28: You probably want to `", ".join()` this list, otherwise it will actually be formatted as a python list

JobDoesburg (changes requested):
see comment

ColonelPhantom:
There's also an automatic email for when the limit is increased in `more_places_email.txt` (`def notify_waiting():`).

I think that email has the same wrong/outdated code, so you should add the same changes there.

JobDoesburg (approved):

PR #2810, opened 2023-02-08
Fix tests on python 3.11

### LINKED ISSUE #2689: Tests fail on Python 3.11
### Describe the change
Currently, when running `make test` with Python 3.11 we get failures in two tests. Both errors are

```
AttributeError: <module 'payments.payables' from 'concrexit/website/payments/payables.py'> does not have the attribute 'get_payable'
```

which seem to originate from 
```py
    @mock.patch("payments.payables.get_payable")
```

I haven't tested if any payable-related code also breaks when running the website normally.
### Motivation
Even if we stick to Python 3.9 for the time being, it is always good to be compatible with the latest version so that we can easily upgrade in the future if desired. Supporting the current version of Python is also convenient for users of operating systems with newer Python versions.

### Current implementation
It is possible to work around this issue by installing Python 3.10 (or 3.9) and then instructing Poetry to use it with e.g.
```sh
$ poetry env use python3.10
```

### Suggested implementation
I believe the error originates from how imports are performed? It looks like there is a change in how imports are handled in Python 3.11, as `get_payable` belongs to a class, but we seem to use it as part of a module.

### Additional context
Full backtrace for one of the failing tests:
```
======================================================================
ERROR: test_creates_payment (payments.tests.api.v1.test_viewsets_payments.PaymentProcessViewTest.test_creates_payment)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib64/python3.11/unittest/mock.py", line 1356, in patched
    with self.decoration_helper(patched,
  File "/usr/lib64/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/unittest/mock.py", line 1338, in decoration_helper
    arg = exit_stack.enter_context(patching)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/contextlib.py", line 502, in enter_context
    result = _enter(cm)
             ^^^^^^^^^^
  File "/usr/lib64/python3.11/unittest/mock.py", line 1427, in __enter__
    original, local = self.get_original()
                      ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/unittest/mock.py", line 1400, in get_original
    raise AttributeError(
AttributeError: <module 'payments.payables' from '/home/quinten/code/thalia/concrexit/website/payments/payables.py'> does not have the attribute 'get_payable'
```

ColonelPhantom:
Closes #2689

### Summary
Tests failed on Python 3.11. It seemed the website itself still worked fine on 3.11. This seems related to subtly changed rules re: aliasing imports (`payments` contains a submodule `payables`, but there is also a line that imports `payables.payables`, which is an instance of `Payables`). I changed the test to work with Python 3.11.

I verified that `payables` still worked, by testing the presence of the "Pay with Thalia Pay" button.

### How to test
1. Install Python 3.11
2. Instruct Poetry to use Python 3.11
3. `make test` -> OK
4. Find code that uses `payments.payables.get_payable` and verify that it still works correctly. (eg `payment_button`)

### DIFF
--- a/website/payments/tests/api/v1/test_viewsets_payments.py
+++ b/website/payments/tests/api/v1/test_viewsets_payments.py
@@ -106,7 +106,7 @@ class PaymentProcessViewTest(TestCase):
         )
 
     @mock.patch("payments.services.create_payment")
-    @mock.patch("payments.payables.get_payable")
+    @mock.patch("payments.payables.payables.get_payable")
     def test_creates_payment(self, get_payable, create_payment):
         def set_payments_side_effect(*args, **kwargs):
             self.payable.model.payment = Payment.objects.create(amount=8)
--- a/website/payments/tests/test_admin_views.py
+++ b/website/payments/tests/test_admin_views.py
@@ -69,7 +69,7 @@ class PaymentAdminViewTest(TestCase):
     @mock.patch("django.contrib.messages.success")
     @mock.patch("django.shortcuts.resolve_url")
     @mock.patch("payments.services.create_payment")
-    @mock.patch("payments.payables.get_payable")
+    @mock.patch("payments.payables.payables.get_payable")
     def test_post(
         self, get_payable, create_payment, resolve_url, messages_success, messages_error
     ):


ColonelPhantom:
now they break on 3.9 :facepalm:

PR #2839, opened 2023-03-01
Create event URL slugs 

### LINKED ISSUE #2308: Use event slug instead of pk in urls
Our event urls (that we share quite often) are pretty ugly (`https://thalia.nu/events/123`). I think it would be nice to have a different url format, like `https://thalia.nu/events/orientationbbq-2022` or  `https://thalia.nu/events/2022/orientationbbq` something.

Add a slug field to events, add url patterns for this.

### Motivation
Nicer urls (that don't necessarily leak when they were created).

Not doing it

rowan-vr:
Closes #2308 

### Summary
Create event URL slugs 

### How to test
1. Run the migrations
2. Observe the generated slugs
3. Create a new event
4. Observe the slug field and see the cool pre-population
5. Submit the form
6. Go to the website/events/slug
7. Also check if the backup (website/events/pk) still works
8. Everything works


website/events/migrations/0064_event_slug.py
```diff
@@ -0,0 +1,42 @@
+# Generated by Django 4.1.7 on 2023-02-22 20:07
+
+from django.db import migrations, models
+
+def slugText(str):
```
JobDoesburg, line 5: camelCase is not python_style

JobDoesburg (changes requested):
My personal opinion would be to not give old events slugs and *only* introduce a new view for the main event page. So not change any existing URL patterns, don't generate URLS in the migration
(but I am open to arguments against this opinion)

website/events/migrations/0064_event_slug.py
DeD1rk: Also, https://docs.djangoproject.com/en/4.1/ref/utils/#django.utils.text.slugify actually guarantees a valid slug as output. This function would not output a valid slug if special characters/accents are in the input

DeD1rk (commented):
I'm a bit scared something will break somewhere with this. So either we need to very thoroughly test this (also from thadmin and thaliapp), or perhaps moving to a converter (https://docs.djangoproject.com/en/4.1/topics/http/urls/#registering-custom-path-converters) would be a cleaner and more testable solution.

JobDoesburg:
> I'm a bit scared something will break somewhere with this. So either we need to very thoroughly test this (also from thadmin and thaliapp), or perhaps moving to a converter (https://docs.djangoproject.com/en/4.1/topics/http/urls/#registering-custom-path-converters) would be a cleaner and more testable solution.

I agree with Dirk

PR #2846, opened 2023-03-04
Fix request.member evaluated too early in API calls

DeD1rk:
### Summary
Fixes a problem (introduced in #2734) where the API calls cannot access `request.member` when authenticated with oauth only. See e.g. https://thalia.sentry.io/issues/3952946075/?project=1463433&query=is%3Aunresolved&referrer=issue-stream where this leads to a 500. In other cases, APIs currently would simply return wrong results.

### How to test
1. Make an API call that would use request.member, but without session authentication (i.e. not from the swagger client, unless you remove the sessionid after authenticating it with oauth).
For example `http 127.0.0.1:8000/api/v2/payments/ authorization:"Bearer <insert a token here>" -h`

### DIFF
--- a/website/announcements/context_processors.py
+++ b/website/announcements/context_processors.py
@@ -17,6 +17,8 @@ def announcements(request):
         for a in Announcement.objects.all()
         if a.is_visible and a.pk not in closed_announcements
     ]
+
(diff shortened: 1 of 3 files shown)


pingiun (approved):
Nice fix

PR #2853, opened 2023-03-08
Include query string in partnerevents API redirect

### LINKED ISSUE #2847: Make API redirects respect query params
### Describe the change
As seen in #2828 and #2821, some old API endpoints are redirected to their new replacements. This redirection does not propagate any query parameters. It would be nice to pass query parameters along when redirecting to a replacement API endpoint to prevent breaking consumers.

### Motivation
#2828 #2821

ColonelPhantom:
Closes #2847 as I think this was the only API endpoint with this issue

### Summary
The redirected endpoint now passes through the query string, so that it filters correctly for example

### How to test
1. Go to an endpoint like localhost:8000/api/v2/partners/events/?search=beest&ordering=start&start=2023-02-01T00%3A00%3A00.000
2. Observe that the redirected URL now still includes the search term and start date

### DIFF
--- a/website/partners/api/v2/urls.py
+++ b/website/partners/api/v2/urls.py
@@ -16,14 +16,18 @@ urlpatterns = [
     path(
         "partners/events/",
         RedirectView.as_view(
-            pattern_name="api:v2:events:external-events-list", permanent=False
+            pattern_name="api:v2:events:external-events-list",
+            permanent=False,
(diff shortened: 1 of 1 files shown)


DeD1rk (approved):
Yeah seems to be the only one. Nice to see that RedirectView provides that option.

PR #2859, opened 2023-03-08
Change order on well-being page 

### LINKED ISSUE #2823: Switch student well-being page contents
We want to have the services from university and the trust contact persons switched on the Student well-being page.

Switch the the two page-sections in the HTML for the different services around, making it so the university services are shown first.

### Motivation
We want to show the university-provided services more clearly and explicitly as suggested by the GM earlier.

Show both side-by-side, but an earlier issue has shown that in current circumstances this is probably not possible.

ColonelPhantom:
Closes #2823 

### Summary
I changed the order of the elements on the page, and also the text for the university services. Specifically:
- I added "the" at the beginning of the sentence
- Remove the word "also"
- Change "some" to "various" as I think "some" is a weird Dutch-ism (from "een aantal"?). (Let me know if you don't like this change/think it's wrong/whatever! For example I can imagine "a number of" could also be preferable.)

### How to test
1. Go to student well-being page
2. Observe that the order is now as desired
3. Check if the text is accurate and flows well.

### DIFF
--- a/website/singlepages/templates/singlepages/student_well-being.html
+++ b/website/singlepages/templates/singlepages/student_well-being.html
@@ -7,33 +7,9 @@
 {% block body %}
     <section class="page-section">
         <div class="container">
-            <h1 class="section-title">{% trans "Trust contact persons" %}</h1>
-            <div class="row mt-4">
-                <div class="col-12 col-md-6">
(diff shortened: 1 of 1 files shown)


DeD1rk (approved):
Nice!

PR #2877, opened 2023-03-15
Add pagination for liked photos

### LINKED ISSUE #2668: Add pagination to liked photos view
There can be a lot of liked photos.

We might want to paginate

### Motivation
S.O.S. (Save Our Servers)

Not doing it

ColonelPhantom:
Closes #2668 

### Summary
Added pagination for the liked photos page.

### How to test
1. Go to liked photos
2. Observe that there is pagination
3. See that the pagination works as intended (e.g. with ordering of pictures).

Should the picture order be changed, e.g. to be in order of 'when liked'? Currently it seems to return in the order of the pictures inside each album. Should I add this to the `order_by` in some way, either to keep the current behavior or get something else?


JobDoesburg (commented):
Maybe this is the time to move to class-based views with built-in pagination for the photos app?

ColonelPhantom:
You mean doing something like `MembersIndex(ListView)` as in `website/members/views.py`? Because it seems that there is still a fair bit of boilerplate for pagination in there, although I'm not opposed to the idea!

JobDoesburg:
Yes, see https://docs.djangoproject.com/en/4.1/topics/pagination/

JobDoesburg:
yes

> Seems good, thanks!
> 
> I'm thinking we should also add some central logic for determining page_range and outputting the whole `nav` tag. I guess this would be a new template that we put in `thaliawebsite`, as well as a view or function to determine the page range?

ColonelPhantom:
I added a `PagedView` class and a `paginated_page.html`. Everything should mostly behave as before, although you now cannot overshoot pages anymore on albums. I don't think this is a big loss, if at all. Especially because `last` now exists.

Speaking of first and last, maybe we could introduce first and last buttons in the `page_range`s.

Also, the non-paged views in the photos app are still function-based views. I think this is fine, as we probably do not want to introduce pagination there anyway.

website/thaliawebsite/templates/paginated_page.html
```diff
@@ -0,0 +1,33 @@
+{% extends 'simple_page.html' %}
+{% load django_bootstrap5 %}
+
+{% block pages %}
+<nav>
+    <ul class="pagination mt-4">
+        {% if page_obj.has_previous %}
+            <li class="page-item">
+                <a class="page-link"
```
DeD1rk, line 9: Broken indentation

website/thaliawebsite/templates/paginated_page.html
```diff
@@ -0,0 +1,33 @@
...
+        {% for page in page_range %}
+            <li class="page-item{% if page == page_obj.number %} active{% endif %}">
+                <a class="page-link" href="{{ request.path }}?page={{ page }}">{{ page }}
+                </a>
+            </li>
+        {% endfor %}
+        {% if page_obj.has_next %}
+            <li class="page-item">
+                <a class="page-link"
+                href="{{ request.path }}?page={{ photos.next_page_number }}">
```
DeD1rk, line 25: The `photos.next_page_number` looks not so generic. (same above)

I think it would be nice to have PagedView.get_context_data just provide e.g. `next_page`,  and `previous_page` (`_number` and/or `_url`).

DeD1rk (changes requested):

DeD1rk (approved):

PR #2884, opened 2023-03-16
Check if file already exists when migrating media

### LINKED ISSUE #2660: Moving files to S3 should be idempotent
Moving files to S3 will always upload the file, even if it already exists.

Check if it already exists

JobDoesburg:
Closes #2660 

### Summary
Checks if a file exists before uploading

### How to test
1. Haven't tested this yet

### DIFF
--- a/website/thaliawebsite/management/commands/migratemedia.py
+++ b/website/thaliawebsite/management/commands/migratemedia.py
@@ -26,13 +26,22 @@ class Command(BaseCommand):
         for full_path in self._get_all_media_file():
             upload_path = self._split_path_to_upload(full_path)
             try:
-                if s3_client.head_object(
-                    Bucket=settings.AWS_STORAGE_BUCKET_NAME, Key=upload_path
-                ):
+                try:
+                    remote_file = s3_client.head_object(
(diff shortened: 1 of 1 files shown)


website/thaliawebsite/management/commands/migratemedia.py
```diff
@@ -24,13 +24,21 @@ def handle(self, *args, **options):
...
         for full_path in self._get_all_media_file():
+            upload_path = self._split_path_to_upload(full_path)
             try:
-                upload_path = self._split_path_to_upload(full_path)
-                s3_client.upload_file(
-                    full_path, settings.AWS_STORAGE_BUCKET_NAME, upload_path
-                )
-                print(f"success upload {upload_path}")
-                logging.info(f"success upload {upload_path}")
+                if s3_client.head_object(
```
DeD1rk, line 29: From https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/head_object.html I gather head_object will raise if the key doesn't exist. So this return value probably shouldn't be used as boolean

DeD1rk (approved):

PR #2865, opened 2023-03-10
Generate model relation diagram in CI

mark-boute:
Closes #2840

### Summary
Requires [GraphViz](https://graphviz.org/download/)
Create workflow that generates a graph png file (could be dot) of the selected application models (all for now).
For configuration see: [django_extensions graph_model](https://django-extensions.readthedocs.io/en/latest/graph_models.html#default-settings)

### How to test
1. Create a PR
2. See a correct model in models_graph.png

For the commit that fixed the poetry.lock conflict:
[Actions run](https://github.com/svthalia/concrexit/actions/runs/4417743204)

PR #2900, opened 2023-03-21
Fix faulty Generate model relation diagram in CI (#2865) dependencies and revert the revert

mark-boute:
Moved django_extensions (needed for graph_models command in manage.py) to only be included when DJANGO_ENV is equal to development or testing


DeD1rk (approved):
Looks about right now :)

PR #2904, opened 2023-03-22
Use docker compose

### LINKED ISSUE #2890: Move from Nix to docker only
None of use can work with Nix, and the current state of stuff doesn't really seem to allow us to even release a new version (i.e. we could probably manage to replace the docker image but not the nginx configs).

We should change our server stuff such that anyone can understand and maintain it: docker compose on a simple linux instance. I'm not entirely sure about the details but I'm thinking:

- Terraform config staying mostly the same but: not NixOS but just e.g. debian/ubuntu server. We can probably also switch to a t4g instance.
- On that instance just install docker and run a docker compose thing. deploying = updating the docker compose file (and optionally a .env) to pull new images. CI can build and push (concrexit-web and concrexit-nginx) images, perhaps to githubs registry?

So the steps to do so I think are:

- [ ] Add a Dockerfile for nginx
- [ ] Create a working docker compose file for production
- [ ] Run cronjobs within the concrexit web container
- [ ] Update terraform config
- [ ] Figure out how exactly to install stuff on a fresh ec2 instance. Ideally this would be pretty much automatic. I guess we can make an install script 
- [ ] Add/update workflows to build and deploy
- [ ] Document all of this so that anyone can use it

DeD1rk:
Closes #2890.

### Summary
This removes NixOS entirely, moving to Debian with Docker Compose for everything.

See `infra/README.md`.

### Postgres 15

Since we'll need to manually move the database from staging/production into the new servers anyway, now seems like the perfect time to also do the upgrade from postgres 11 to postgres 15: the process of upgrading is the same as that of the move we'll have to do:

- Stop old server (at least we should probably prevent it from writign to the db, so we can just stop the concrexit container).
- Dump the existing database: `sudo docker exec -i <postgres container> pg_dump -U concrexit > dump.sql`
- Copy the dump to the new server (`scp`?)
- Create postgres container (don't start concrexit yet, the postgres data volume should be empty).
- Copy dump into postgres container: `sudo docker compose cp /home/admin/dump.sql postgres:/dump.sql`
- Import the dump: `sudo docker compose exec postgres sh -c "psql -U concrexit < /dump.sql"`
- Clean up any files that are left behind.

The above can be done a bit more efficiently using a compressed dump format (see `pg_dump` docs) and `pg_restore` instead of plain `psql`. 

Using a SQL dump I've tested these steps copying the db from staging into docker-staging. This works well, except of course the S3 bucket has not been copied so right now the media are unavailable.

#### To do now

- [x] Figure out why cronjobs are executed, but management commands don't seem to be getting called
- [ ] ~Figure out how to get cron logging output into the docker logs.~ (No need if we're switching to django-cron soon).
- [x] Change workflow to deploy to `staging` environment (on push master).
- [x] Figure out how if we can simply import the existing S3 bucket into the new terraform config.
- [x] Get rid of DisallowedHost (sentry) logs: scanner bots are spamming sentry and the docker logs with thousands of DisallowedHost errors. Can we filter those requests out, or prevent logging them?

#### To do later

- [ ] Can we make concrexit container on alpine?
- [ ] Add a local docker-compose file (with build definitions and perhaps even mounted source code
- [ ] django-cron


JAicewizard (commented):
In general it might be nice to have the hardcoded paths be a variable declared at the start

DeD1rk:
> In general it might be nice to have the hardcoded paths be a variable declared at the start

Tempted against this, as deploy workflows will have to rely on the paths, and I don't really see any good reason to deviate from a default

PR #2907, opened 2023-03-22
Add information about nixos deployments

pingiun:
### Summary
Add extra information about the two nix commands needed


ColonelPhantom (approved):
Nice and clear!

PR #2920, opened 2023-03-28
Moneybird synchronisation

nvoers:
Creates a synchronisation between payments in the website and the bookkeeping in Moneybird.

Each payment is posted to Moneybird upon creation.
Ideally, each day all of the card and cash payments are linked to financial statements.
Upon processing of a ThaliaPay batch, those payments are linked to a financial statement.

NOTE: there are still some hardcoded things which need to be changed, but this shouldn't affect the review (I think)


website/moneybirdsynchronization/migrations/0005_contact_moneybird_version_alter_contact_moneybird_id.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 4.1.7 on 2023-02-23 10:59
```
JobDoesburg, line 1: Please merge those migrations (that is: remove them and regenerate them)

website/moneybirdsynchronization/models.py
```diff
@@ -0,0 +1,47 @@
...
+                "lastname": self.member.last_name,
+                "address1": self.member.profile.address_street,
+                "address2": self.member.profile.address_street2,
+                "zipcode": self.member.profile.address_postal_code,
+                "city": self.member.profile.address_city,
+                "country": self.member.profile.address_country,
+                "send_invoices_to_email": self.member.email,
+                "custom_fields_attributes": {
+                    "0":{
+                        "id": 380500576947930894,
```
JobDoesburg, line 38: Let's make this a setting, yeah

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,67 @@
+from payments.models import Payment
+import datetime
+
+def get_financial_account_id(api, name):
```
JobDoesburg, line 4: I propose to rewrite this to a class based MoneybirdAPIService that takes `api` at construction

JobDoesburg, line 4: what is `name` exactly?

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,67 @@
...
+                }
+            )
+
+            for x in range(len(new_cash_payments)):
+                instance = new_cash_payments[x]
+                instance.moneybird_financial_statement_id = statement_response["id"]
+                instance.moneybird_financial_mutation_id = statement_response["financial_mutations"][x]["id"]
+                instance.save()
+
+                mutation_response = api.patch("financial_mutations/{}/link_booking".format(instance.moneybird_financial_mutation_id),{
```
JobDoesburg, line 62:
```suggestion
                mutation_response = api.patch(f"financial_mutations/{instance.moneybird_financial_mutation_id}/link_booking",{
```

website/moneybirdsynchronization/services.py
JobDoesburg: Or even better, calls `HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)` at construction itself

website/moneybirdsynchronization/signals.py
```diff
@@ -0,0 +1,359 @@
...
+    api = HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)
+    member = Member.objects.get(profile=instance)
+    contact = Contact.objects.get_or_create(member=member)[0]
+    if contact.moneybird_version is None:
+        response = api.post("contacts", contact.to_moneybird())
+        contact.moneybird_id = response["id"]
+        contact.moneybird_version = response["version"]
+        contact.save()
+    else:
+        response = api.patch("contacts/{}".format(contact.moneybird_id), contact.to_moneybird())
```
JobDoesburg, line 32:
```suggestion
        response = api.patch(f"contacts/{contact.moneybird_id}", contact.to_moneybird())
```

website/moneybirdsynchronization/signals.py
```diff
@@ -0,0 +1,359 @@
...
+@suspendingreceiver(
+    pre_delete,
+    sender="members.Profile",
+)
+def post_profile_delete(sender, instance, **kwargs):
+    """Delete contact on profile deletion."""
+    api = HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)
+    member = Member.objects.get(profile=instance)
+    contact = Contact.objects.get(member=member)
+    api.delete("contacts/{}".format(contact.moneybird_id))
```
JobDoesburg, line 46:
```suggestion
    api.delete("contacts/{contact.moneybird_id}")
```

JobDoesburg, line 46: What will happen if we're deleting a person that has invoices attached to it on Moneybird? Will it delete the contact? Will it throw an error and crash this code?

website/moneybirdsynchronization/signals.py
```diff
@@ -0,0 +1,359 @@
...
+@suspendingreceiver(
+    post_save,
+    sender="auth.User",
+)
+def post_user_save(sender, instance, **kwargs):
+    """Update contact on user creation."""
+    api = HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)
+    contact = Contact.objects.get_or_create(member=instance)[0]
+    if contact.moneybird_version is not None:
+        response = api.patch("contacts/{}".format(contact.moneybird_id), contact.to_moneybird())
```
JobDoesburg, line 59: Maybe don't copy this code, but call the same code as `post_profile_save`

website/moneybirdsynchronization/signals.py
```diff
@@ -0,0 +1,359 @@
...
+)
+def post_event_registration_payment(sender, instance, **kwargs):
+    """ Create external invoice on payment creation."""
+    if instance.payment is None:
+        return
+    if instance.payment.moneybird_invoice_id is not None:
+        return
+    
+    api = HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)
+    contact_id = api.get("contacts/customer_id/34")["id"]
```
JobDoesburg, line 76: Is contact with customer id 34 the anonymous contact? That should be documented (and again, don't hardcode)

website/moneybirdsynchronization/signals.py
```diff
@@ -0,0 +1,359 @@
...
+            ],
+        }
+    }
+
+    try:
+        response = api.post("external_sales_invoices", invoice_info)
+        instance.payment.moneybird_invoice_id = response["id"]
+        instance.payment.save()
+    except Exception as e:
+        pass
```
JobDoesburg, line 112: Do you really want to do this?

JobDoesburg: Same for all other exceptions in this file

website/moneybirdsynchronization/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
+
+# Create your views here.
```
JobDoesburg, line 3: can be removed

JobDoesburg (changes requested):
Looks good in general, I think some small refactor could improve readability a lot. Some constants should be moved to settings, or queries/created on MB.

And finally, something seems to have gone wrong with git, as it shows 198 file changes. Maybe try to merge with master one more time.

website/moneybirdsynchronization/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
+
+# Create your tests here.
```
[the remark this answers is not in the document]
nvoers: Ideally yes, but something to make later(?)

website/moneybirdsynchronization/services.py
nvoers: identifier of the financial account. Changed the name accordingly

website/moneybirdsynchronization/services.py
```diff
@@ -1,67 +1,75 @@
...
+            if ledger["name"] == "Contribution":
+                return ledger["id"]
+        return None
+
+
+    def link_transaction_to_financial_account(self, account_id, new_cash_payments):
+        financial_mutations_attributes = []
+        if account_id is not None:
+            for instance in new_cash_payments:
+                payment_response = self.api.post("external_sales_invoices/{instance.moneybird_invoice_id}/payments",
```
JobDoesburg, line 37: this should probably be an f-string

website/moneybirdsynchronization/services.py
```diff
@@ -1,67 +1,75 @@
...
+    def get_contribution_ledger_id(self):
+        for ledger in self.api.get("ledger_accounts"):
+            if ledger["name"] == "Contribution":
+                return ledger["id"]
+        return None
+
+
+    def link_transaction_to_financial_account(self, account_id, new_cash_payments):
+        financial_mutations_attributes = []
+        if account_id is not None:
```
JobDoesburg, line 35: I would swap the condition and early-return to improve readability

website/moneybirdsynchronization/services.py
```diff
@@ -1,67 +1,75 @@
...
+                    instance.save()
 
-                mutation_response = api.patch("financial_mutations/{instance.moneybird_financial_mutation_id}/link_booking",{
-                        "booking_type": "ExternalSalesInvoice",
-                        "booking_id": instance.moneybird_invoice_id,
-                        "price": str(instance.amount),
-                        "description": instance.topic,
-                    })
\ No newline at end of file
+                    mutation_response = self.api.patch("financial_mutations/{instance.moneybird_financial_mutation_id}/link_booking",{
```
JobDoesburg, line 70: more f-strings

website/moneybirdsynchronization/signals.py
```diff
@@ -53,10 +53,10 @@ def post_profile_delete(sender, instance, **kwargs):
 )
 def post_user_save(sender, instance, **kwargs):
     """Update contact on user creation."""
-    api = HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)
+    apiservice = MoneybirdAPIService()
     contact = Contact.objects.get_or_create(member=instance)[0]
     if contact.moneybird_version is not None:
-        response = api.patch("contacts/{contact.moneybird_id}", contact.to_moneybird())
+        response = apiservice.api.patch("contacts/{contact.moneybird_id}", contact.to_moneybird())
```
JobDoesburg, line 59: even momber f-strings

JobDoesburg (changes requested):

website/moneybirdsynchronization/signals.py
nvoers: probably the latter. We could choose to not remove them if they have an invoice, but that will clutter Moneybird over time

nvoers: Moneybird archives the contact when it cannot be deleted

JobDoesburg:
@nvoers could you try to merge with master or something? The diff currently shows 200 changed files which isn't correct. If you want, I can also add you as outside collaborator to the repo

website/moneybirdsynchronization/emails.py
```diff
@@ -0,0 +1,23 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def send_sync_error(error, payment):
+
+    send_email(
+        to=[settings.TREASURER_NOTIFICATION_ADDRESS],
+        subject="[MONEYBIRD] Error while syncing payment",
+        txt_template="apifails/weekly_overview.txt",
```
JobDoesburg, line 15: weekly overview?

website/moneybirdsynchronization/management/commands/contactsync.py
```diff
@@ -0,0 +1,45 @@
...
+from thaliawebsite import settings
+from members.models import Member
+from django.db.models import Q, Subquery, OuterRef
+
+class Command(BaseCommand):
+    """This command can be executed periodically to sync the contacts with MoneyBird."""
+
+    def handle(self, *args, **options):
+        members_without_contact = Member.objects.filter(~Q(id__in=Subquery(Contact.objects.filter(member=OuterRef('pk')).values('member'))))
+
```
JobDoesburg, line 22: It would be better to put the actual syncing in `services.py` and call it from this command. This way, it is easier to implement tests and it becomes possible to call the sync in a different way than just via the management command

website/moneybirdsynchronization/management/commands/statementsync.py
```diff
@@ -0,0 +1,31 @@
...
+from payments.models import Payment
+from moneybirdsynchronization.administration import HttpsAdministration
+from thaliawebsite import settings
+
+class Command(BaseCommand):
+    """This command needs to be executed every day to sync the bookkeeping with MoneyBird."""
+
+    def handle(self, *args, **options):
+
+        payment_identifiers = {
```
JobDoesburg, line 14: what are payment identifiers? it are names for financial accounts I think, right (so I think it would be better to name the variable something like `payment_type_to_financial_account_mapping`)

website/moneybirdsynchronization/management/commands/statementsync.py
```diff
@@ -0,0 +1,31 @@
...
+            Payment.TPAY: "ThaliaPay",
+            Payment.CASH: "cashtanje",
```
JobDoesburg, line 16: Let's not hardcode the values here, but put this mapping in settings.py

website/moneybirdsynchronization/management/commands/statementsync.py
```diff
@@ -0,0 +1,31 @@
...
+        new_card_payments = Payment.objects.filter(type=Payment.CARD, created_at__year=date.year, created_at__month=date.month, created_at__day=date.day, moneybird_financial_statement_id=None)
+        new_cash_payments = Payment.objects.filter(type=Payment.CASH, created_at__year=date.year, created_at__month=date.month, created_at__day=date.day, moneybird_financial_statement_id=None)
+        
+        apiservices = MoneybirdAPIService()
+
+        card_account_id = apiservices.get_financial_account_id(payment_identifiers[Payment.CARD])
+        apiservices.link_transaction_to_financial_account(card_account_id, new_card_payments)
+
+        cash_account_id = apiservices.get_financial_account_id(payment_identifiers[Payment.CASH])
+        apiservices.link_transaction_to_financial_account(cash_account_id, new_cash_payments)
```
JobDoesburg: Same as for the contact sync, let's put this code in services.py

website/moneybirdsynchronization/models.py
```diff
@@ -0,0 +1,48 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from thaliawebsite import settings
+
+from members.models import Member
+
+class Contact(models.Model):
```
JobDoesburg, line 8: I think our CI requires you to override __str__ and define a meta class with verbose names

JobDoesburg, line 8: maybe renaming to MoneybirdContact would be more descriptive?

website/moneybirdsynchronization/models.py
```diff
@@ -0,0 +1,48 @@
...
+    member = models.OneToOneField(
+        Member,
+        on_delete=models.CASCADE,
+        verbose_name=_("member"),
+        null=True,
+        blank=True,
```
JobDoesburg, line 14: I think adding a related name `moneybird_contact` would be nice, so you can access it via `member.moneybird_contact`

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+from moneybirdsynchronization import emails
+from thaliawebsite import settings
+from django.template.defaultfilters import date
+
+import datetime
+
+class MoneybirdAPIService():
+
+    def __init__(self):
+        api = HttpsAdministration(settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID)
```
JobDoesburg: Maybe even better, don't hardcode `settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID` in the constructor but make it arguments. This uncouples your moneybird code from concrexit

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+from members.models import Member
+from moneybirdsynchronization.models import Contact
+from moneybirdsynchronization.administration import HttpsAdministration, Administration
+from moneybirdsynchronization import emails
+from thaliawebsite import settings
+from django.template.defaultfilters import date
+
+import datetime
+
+class MoneybirdAPIService():
```
JobDoesburg: Looking at it right now, I propose to move this to a separate file

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+def update_contact(member):
+    apiservice = MoneybirdAPIService()
+    apiservice.update_contact(Contact.objects.get_or_create(member=member)[0])
+
+
+def delete_contact(instance):
+    apiservice = MoneybirdAPIService()
+    member = Member.objects.get(profile=instance)
+    contact = Contact.objects.get(member=member)
+    apiservice.api.delete(f"contacts/{contact.moneybird_id}")
```
JobDoesburg, line 127: It would be better to implement `MoneybirdAPIService.delete_contact` instead of referencing `MoneybirdAPIService.api.delete` yourself

website/moneybirdsynchronization/services.py
JobDoesburg: I would make `api` private, so refactor to `_api` or even `__api`, as you don't want allow others to make calls to it

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+    contact.delete()
+
+
+def register_event_registration_payment(instance):
+    apiservice = MoneybirdAPIService()
+    contact_id = apiservice.api.get(f"contacts/customer_id/{settings.MONEYBIRD_UNKOWN_PAYER_ID}")["id"]
```
JobDoesburg: same as above, implement `MoneybirdAPIService.get_contact_id_from_customer_id`

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+                    "description": instance.payment.topic,
+                    "price": str(instance.payment.amount),
+                    "project_id": project_id,
+                },
+            ],
+        }
+    }
+
+    try:
+        response = apiservice.api.post("external_sales_invoices", invoice_info)
```
JobDoesburg, line 165: Same as above, implement `MoneybirdAPIService.create_external_sales_invoice` (possibly with a lot of (optional) parameters, rather than taking a random dict for invoice info)

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+        project_name = f"{event.title} [{start_date}]"
+    except:
+        start_date = date(instance.start, "Y-m-d")
+        project_name = f"{instance.__str__()} [{start_date}]"
+    
+    apiservice = MoneybirdAPIService()
+    project_id = apiservice.get_project_id(project_name)
+
+    for order in orders:
+        contact_id = apiservice.api.get(f"contacts/customer_id/{settings.MONEYBIRD_UNKOWN_PAYER_ID}")["id"]
```
JobDoesburg: same as above

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+                        "description": order.payment.topic,
+                        "price": str(order.payment.amount),
+                        "project_id": project_id,
+                    },
+                ],
+            }
+        }
+
+        try:
+            response = apiservice.api.post("external_sales_invoices", invoice_info)
```
JobDoesburg, line 213: same as above

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+        response = apiservice.api.post("external_sales_invoices", invoice_info)
+        instance.payment.moneybird_invoice_id = response["id"]
+        instance.payment.save()
+    except Exception as e:
+        emails.send_sync_error(e, instance.payment)
+
+
+def register_contribution_payment(instance):
+    apiservice = MoneybirdAPIService()
+    contact_id = apiservice.api.get(f"contacts/customer_id/{settings.MONEYBIRD_UNKOWN_PAYER_ID}")["id"]
```
JobDoesburg: same as above

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,297 @@
...
+        response = apiservice.api.post("external_sales_invoices", invoice_info)
+        instance.payment.moneybird_invoice_id = response["id"]
+        instance.payment.save()
+    except Exception as e:
+        emails.send_sync_error(e, instance.payment)
+
+
+def delete_payment(instance):
+    apiservice = MoneybirdAPIService()
+    apiservice.api.delete(f"external_sales_invoices/{instance.moneybird_invoice_id}")
```
JobDoesburg, line 302: same as above, implement `MoneybirdAPIService.delete_external_sales_invoice`

website/moneybirdsynchronization/templates/apifails/api_fail.html
```diff
@@ -0,0 +1,13 @@
...
+  Hi,
+
+  There has been an error when trying to synchronize with Moneybird.
+  Error: {{ error.message }}
+  Payment: {{ payment }}
+  Link to payment: {{ payment.get_absolute_url }}
+  
+  Kisses,
+  The Website
+{% endblock %}
```
JobDoesburg, line 13: Please put these email templates in a folder like `templates/email/moneybird_api_fail.html`

website/payments/admin.py
```diff
@@ -68,6 +71,9 @@ class PaymentAdmin(admin.ModelAdmin):
         "topic",
         "notes",
         "batch",
+        "moneybird_invoice_id",
+        "moneybird_financial_statement_id",
+        "moneybird_financial_mutation_id",
```
JobDoesburg, line 76: Don't have the payments app depend on the moneybird sync app. Instead, create a separate admin (possibly, you can extend the PaymentAdmin and reregister the payments admin in the moneybird app, though!, see https://stackoverflow.com/questions/25964028/how-do-i-register-a-model-that-is-already-registered-in-admin)

website/payments/models.py
```diff
@@ -179,6 +179,27 @@ class Payment(models.Model):
...
+        null=True,
+    )
+
+    moneybird_financial_mutation_id = models.CharField(
+        verbose_name=_("moneybird financial mutation id"),
+        max_length=255,
+        blank=True,
+        null=True,
+    )
+
```
JobDoesburg, line 202: This should, similar to the moneybird contact, be a separate model in the moneybird app with a one-to-one key to payments. We want to separate payments from moneybird sync, so even when moneybird breaks, payments doesn't break.

website/payments/services.py
```diff
@@ -158,7 +158,7 @@ def process_batch(batch):
             bank_account.last_used = batch.withdrawal_date
             bank_account.save()
 
-    batch.save()
+    batch.save(update_fields=["processed"])
```
JobDoesburg, line 162: Why do we change this? I think this also breaks some logic in the batch's `save()` method that tries to set the `processed_at` field. 

If you want to know which field was updated in the signal, please detect it yourself.

website/registrations/models.py
```diff
@@ -45,7 +45,7 @@ class Entry(models.Model):
         verbose_name=_("status"),
         choices=STATUS_TYPE,
         max_length=20,
-        default="confirm",
+        default="review",
```
JobDoesburg, line 48: Why is this changed?

website/thaliawebsite/settings.py
```diff
@@ -135,10 +135,10 @@ def from_env(
 
 # We use this setting to generate the email addresses
 SITE_DOMAIN = from_env(
-    "SITE_DOMAIN", development="thalia.localhost", production="thalia.nu"
+    "SITE_DOMAIN", development="localhost:8000", production="thalia.nu"
```
JobDoesburg, line 138: I agree that this would be better, but this should be a different PR

website/thaliawebsite/settings.py
```diff
@@ -135,10 +135,10 @@ def from_env(
...
 )
 # We use this domain to generate some absolute urls when we don't have access to a request
-BASE_URL = os.environ.get("BASE_URL", f"https://{SITE_DOMAIN}")
+BASE_URL = os.environ.get("BASE_URL", f"http://{SITE_DOMAIN}")
```
JobDoesburg, line 141: These are development settings

website/thaliawebsite/settings.py
```diff
@@ -601,6 +605,10 @@ def from_env(
             "propagate": False,
         },
     },
+    "root": {
+        "handlers": ["console"],
+        "level": "INFO",
+    },
```
JobDoesburg, line 611: This change shouldn't be there

website/thaliawebsite/settings.py
```diff
@@ -944,3 +952,14 @@ def from_env(
         "auth",
     ],
 }
+
+MONEYBIRD_ADMINISTRATION_ID = 380475297324075003
```
JobDoesburg: make this a os.environ.get
(also, what happens if we don't set it? Then `MONEYBIRD_ADMINISTRATION_ID` should be `None` and the whole sync should just be inactive if you ask me. Is this implemented? Or will it crash?)

website/thaliawebsite/settings.py
```diff
@@ -944,3 +952,14 @@ def from_env(
...
+MONEYBIRD_API_KEY = "-Xw8P1CfYFAA8Z0VEEnOgDvXiNNb8mgdBCgEMl05Yus"
```
JobDoesburg: this secret is now in the git reflog 🥳 so you should revoke it

website/thaliawebsite/settings.py
```diff
@@ -944,3 +952,14 @@ def from_env(
...
+
+MONEYBIRD_AUTO_PUSH = True # Push changes to Moneybird automatically (so you don't have to call `instance.push_to_moneybird()` manually)
```
JobDoesburg: I think these 2 settings, `MONEYBIRD_AUTO_PUSH` and `MONEYBIRD_FETCH_BEFORE_PUSH` are not used

JobDoesburg (changes requested):

website/moneybirdsynchronization/emails.py
nvoers: godver

website/moneybirdsynchronization/management/commands/statementsync.py
nvoers: How exactly? apps are not loaded when settings.py is initialised so making this exact mapping there won't work

JobDoesburg: hmmm that's right... in that case, I would not take `Payment.TPAY` but the literal values these constants represent: `"tpay"`, `"cash"`, `"card"`

website/payments/services.py
nvoers: I tried but `update_fields` doesn't show a value if this is not set. I also couldn't find a different way to find out if the batch went to `processed` during the save

JobDoesburg: I just realized the other approach only works with pre_save signals. Maybe create a `PushedThaliaPayBatch` model in the moneybirdsynchronization app and check at each signal if it was pushed already

website/registrations/migrations/0030_alter_entry_status.py
```diff
@@ -0,0 +1,28 @@
...
+                    ("rejected", "Rejected"),
+                    ("accepted", "Accepted"),
+                    ("completed", "Completed"),
+                ],
+                default="confirm",
+                max_length=20,
+                verbose_name="status",
+            ),
+        ),
+    ]
```
JobDoesburg, line 28: These 2 migrations should be removed

website/thaliawebsite/settings.py
```diff
@@ -944,3 +948,19 @@ def from_env(
...
+MONEYBIRD_SYNC_ENABLED = MONEYBIRD_ADMINISTRATION_ID and MONEYBIRD_API_KEY
+
+MONEYBIRD_CUSTOM_FIELD_ID = os.environ.get("MONEYBIRD_CUSTOM_FIELD_ID", None)
+MONEYBIRD_UNKOWN_PAYER_ID = os.environ.get("MONEYBIRD_UNKOWN_PAYER_ID", None)
+MONEYBIRD_CONTRIBUTION_LEDGER_ID = os.environ.get(
+    "MONEYBIRD_CONTRIBUTION_LEDGER_ID", None
+)
+
+PAYMENT_TYPE_TO_FINANCIAL_ACCOUNT_MAPPING = {
+    "tpay": "ThaliaPay",
```
JobDoesburg, line 963: maybe just make this a financial account id?

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,246 @@
...
+from payments.models import Payment
+from thaliawebsite import settings
+
+
+def get_moneybird_api_service():
+    if (
+        settings.MONEYBIRD_ADMINISTRATION_ID is None
+        or settings.MONEYBIRD_API_KEY is None
+    ):
+        return RuntimeError("Moneybird API key or administration ID not set")
```
JobDoesburg, line 22: don't return but raise

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,246 @@
...
+    )
+
+
+def push_thaliapay_batch(instance):
+    if settings.MONEYBIRD_SYNC_ENABLED is False:
+        return
+
+    apiservice = get_moneybird_api_service()
+    payments = Payment.objects.filter(batch=instance)
+    tpay_account_id = apiservice.get_financial_account_id("ThaliaPay")
```
JobDoesburg, line 34: This account id can be taken from settings

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,246 @@
...
+        settings.MONEYBIRD_ADMINISTRATION_ID is None
+        or settings.MONEYBIRD_API_KEY is None
+    ):
+        return RuntimeError("Moneybird API key or administration ID not set")
+    return MoneybirdAPIService(
+        settings.MONEYBIRD_API_KEY, settings.MONEYBIRD_ADMINISTRATION_ID
+    )
+
+
+def push_thaliapay_batch(instance):
```
JobDoesburg, line 28: rename instance to `batch`

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,246 @@
...
+def delete_contact(instance):
+    if settings.MONEYBIRD_SYNC_ENABLED is False:
+        return
+
+    apiservice = get_moneybird_api_service()
+    member = Member.objects.get(profile=instance)
+    apiservice.delete_contact(MoneybirdContact.objects.get(member=member))
+
+
+def register_event_registration_payment(instance):
```
JobDoesburg, line 55: rename `instance`

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,246 @@
...
+
+        try:
+            response = apiservice.api.post("external_sales_invoices", invoice_info)
+            order.payment.moneybird_invoice_id = response["id"]
+            order.payment.save()
+        except Administration.Error as e:
+            emails.send_sync_error(e, instance.payment)
+
+
+def register_food_order_payment(instance):
```
JobDoesburg, line 109: rename instance

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,246 @@
...
+    new_cash_payments = Payment.objects.filter(
+        type=Payment.CASH,
+        created_at__year=date.year,
+        created_at__month=date.month,
+        created_at__day=date.day,
+        moneybird_financial_statement_id=None,
+    )
+
+    card_account_id = apiservice.get_financial_account_id(
+        settings.PAYMENT_TYPE_TO_FINANCIAL_ACCOUNT_MAPPING["card"]
```
JobDoesburg, line 239: here you can just use `Payment.CARD` I think

JobDoesburg (changes requested):
We're almost there!

website/moneybirdsynchronization/services.py
nvoers: no since the value of `Payment.CARD` is different from the current value that is passed

website/thaliawebsite/settings.py
```diff
@@ -944,3 +948,23 @@ def from_env(
         "auth",
     ],
 }
+MONEYBIRD_ADMINISTRATION_ID = os.environ.get("MONEYBIRD_ADMINISTRATION_ID", None)
+MONEYBIRD_API_KEY = os.environ.get("MONEYBIRD_API_KEY", None)
+
+MONEYBIRD_SYNC_ENABLED = MONEYBIRD_ADMINISTRATION_ID and MONEYBIRD_API_KEY
+
+MONEYBIRD_CUSTOM_FIELD_ID = os.environ.get("MONEYBIRD_CUSTOM_FIELD_ID", None)
```
JobDoesburg, line 956: What is the purpose of this custom field (what do we put in the custom field? is it the member id in concrexit? in that case I would consider refactoring to MONEYBIRD_MEMBER_PK_CUSTOM_FIELD_ID or something)

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,257 @@
...
+    invoice_info = apiservice.create_external_sales_info(
+        contact_id, registration.payment, project_id
+    )
+
+    try:
+        response = apiservice.create_external_sales_invoice(invoice_info)
+        registration.payment.moneybird_external_invoice.moneybird_invoice_id = response[
+            "id"
+        ]
+        registration.payment.moneybird_external_invoice.save()
```
JobDoesburg, line 81: I think it is better to attach the external invoice to the event registration and not to the payment

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,257 @@
...
+    for order in orders:
+        contact_id = apiservice.get_contact_id_by_customer(order)
+
+        invoice_info = apiservice.create_external_sales_info(
+            contact_id, order.payment, project_id
+        )
+
+        try:
+            response = apiservice.create_external_sales_invoice(invoice_info)
+            order.payment.moneybird_external_invoice.moneybird_invoice_id = response[
```
JobDoesburg, line 111: Same here, I think it makes more sense to attach the `moneybird_external_invoice` to the order instead of the payment

website/moneybirdsynchronization/models.py
```diff
@@ -0,0 +1,118 @@
...
+    def __str__(self):
+        return f"Moneybird contact for {self.member}"
+
+    class Meta:
+        verbose_name = _("moneybird contact")
+        verbose_name_plural = _("moneybird contacts")
+
+
+class MoneybirdExternalInvoice(models.Model):
+    payment = models.OneToOneField(
```
JobDoesburg, line 62: I think it is nicer to make this a GenericForeignKey to event registrations, orders, food orders and registrations

website/thaliawebsite/settings.py
nvoers: Yes it is the member id/pk

website/moneybirdsynchronization/templates/email/moneybird_api_fail.txt
```diff
@@ -0,0 +1,10 @@
+{% load i18n bleach_tags %}
```
DeD1rk, line 1: I don't think these loads are needed.

website/moneybirdsynchronization/models.py
```diff
@@ -0,0 +1,115 @@
...
+        return f"Moneybird contact for {self.member}"
+
+    class Meta:
+        verbose_name = _("moneybird contact")
+        verbose_name_plural = _("moneybird contacts")
+
+
+class MoneybirdExternalInvoice(models.Model):
+    payable_model = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    object_id = models.CharField(max_length=255)
```
nvoers, line 65: @JobDoesburg any suggestion for a default? Or set `null=true`?

JobDoesburg: Hmm, I think this should not be necessary? Maybe it is because you are _migrating_ and not _introducing_ the field? In that case, simply removing the migrations and recreating them would fix it. Otherwise I would say null=true (but I really don't think its necessary)

website/moneybirdsynchronization/templates/email/moneybird_api_fail.html
```diff
@@ -0,0 +1,12 @@
+{% extends "email/board_email.html" %}
+
+{% block content %}
+  <p>
+  Hi,
+
+  There has been an error when trying to synchronize with Moneybird.
+  Error: {{ error.message }}
+  Payment: {{ payment }}
+  Link to payment: <a>{{ payment.get_absolute_url }}</a>
```
DeD1rk, line 10:
```suggestion
  Link to payment: <a href="{{ payment.get_absolute_url }}">{{ payment.get_absolute_url }}</a>
```

website/moneybirdsynchronization/services.py
```diff
@@ -0,0 +1,249 @@
...
+            )
+
+    ids = [info["id"] for info in contact_info]
+    for moneybird in moneybird_info:
+        if int(moneybird["id"]) not in ids:
+            apiservice.delete_contact(moneybird["id"])
+
+
+def sync_statements():
+    if settings.MONEYBIRD_SYNC_ENABLED is False:
```
DeD1rk, line 223:
```suggestion
    if not settings.MONEYBIRD_SYNC_ENABLED:
```
If `MONEYBIRD_SYNC_ENABLED=None`, moneybird sync should be disabled.

JobDoesburg: @nvoers can you change this everywhere

PR #2969, opened 2023-04-05
Fix event cancellation and other mail improvements

### LINKED ISSUE #2964: Error when canceling certain event on staging
Sentry Issue: [CONCREXIT-Q9](https://thalia.sentry.io/issues/4064695756/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'name'
(5 additional frame(s) were not displayed)
...
  File "events/views.py", line 147, in post
    services.cancel_registration(request.member, event)
  File "events/services.py", line 323, in cancel_registration
    emails.notify_organiser(event, registration)
  File "events/emails.py", line 53, in notify_organiser
    to=[
  File "events/emails.py", line 54, in <listcomp>
    organiser.contact_mailinglist.name + "@" + settings.SITE_DOMAIN
```

ColonelPhantom:
Closes #2964 and does some code cleanup

### Summary
- Event emails (about late event cancellation etc) now uses contact_address instead of contact_mailinglist so that all member groups can now be reached. This also means no None checking is necessary because this field is always true: we verify in `clean()` that there is either a mailinglist or an address.
- Remove check for organisers without contact mailing list because we can now also mail the contact address.
- Remove some duplicate code in the admin, which does change shown email addresses on staging and localhost
- The `committees` and `societies` mailinglists now include membergroups with contact_email rather than contact_mailinglist set.

### How to test
1. Create an event with some organisers that do not have a mailing list associated with them, and cancellation notifications enabled.
2. Observe that you can now cancel from this event, unlike before.
3. Also observe that committees and societies still have correct email addresses in the admin list display. (Note that thalia.nu may be replaced by thalia.localhost, this is because it uses the SITE_DOMAIN.)
4. Somehow check that the mailinglist integration still reports the correct email addresses. I am not sure how to do this (@JobDoesburg do you maybe know this?)

### DIFF
--- a/website/activemembers/admin.py
+++ b/website/activemembers/admin.py
@@ -46,7 +46,7 @@ class MemberGroupAdmin(admin.ModelAdmin):
 
     inlines = (MemberGroupMembershipInline,)
     form = MemberGroupForm
-    list_display = ("name", "since", "until", "active", "email")
+    list_display = ("name", "since", "until", "active", "contact_address")
     list_filter = (
         "until",
         "active",
@@ -66,14 +66,6 @@ class MemberGroupAdmin(admin.ModelAdmin):
         "active",
         "display_members",
     )
-
-    @staticmethod
-    def email(instance):
-        if instance.contact_email:
-            return instance.contact_email
-        if instance.contact_mailinglist:
-            return instance.contact_mailinglist.name + "@thalia.nu"
-        return None
 
 
 @admin.register(models.Committee)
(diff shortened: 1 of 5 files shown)


DeD1rk (approved):
Code looks good to me! I'm also not sure about testing this with gsuite. On staging some gsuite/email stuff isnt quite working anyway.

DeD1rk:
One more update, I now fixed the gsuite sync on staging. So we can try it out there

PR #2995, opened 2023-04-12
Add basic git commands in readme

T8902:
### Summary
Added git commands to read me to make it a bit more accessible to new technici members so they don't have to ask what commands to use continuously


ColonelPhantom:
Nice idea, but it might be clearer to put the essential commands "in order". That way it is also documented that you first need to branch, then add, then commit, then push, etc.

Also, maybe use `git switch -c` instead of `git checkout -b` for consistency.

DeD1rk (approved):

PR #2997, opened 2023-04-12
Implement rate-limiting

### LINKED ISSUE #2933: Implement rate-limiting for password-forgot
### Describe the bug
There is no rate-limiting on the password-forgot view. It is thus possible to spam a user with known emailadress with a lot of password reset emails and perform some kind of DoS.

Also note that for password-forgot, we use uuid's, which are not secure! We should implement secure secret tokens.

### Expected behaviour
Check if a password reset token was sent in the past x minutes. 

### Additional context
We were notified about this via responsible disclosure (the rate-limiting, not the uuid part)

JobDoesburg:
Closes #2933 

### Summary
Add rate-limiting to the most important anonymous user-facing views


website/registrations/tests/test_views.py
```diff
@@ -86,7 +86,8 @@ def _give_user_permissions(self):
     def test_permissions(self):
         url = f"/registration/admin/process/{self.entry1.pk}/"
         response = self.client.post(url)
-        self.assertRedirects(response, f"/admin/login/?next={url}")
+        self.assertEqual(response.status_code, 302)
```
DeD1rk, line 89: Why do this? assertRedirects assumes 302 by default...

website/thaliawebsite/views.py
```diff
@@ -51,3 +54,19 @@ def get_context_data(self, **kwargs) -> dict:
         )
 
         return context
+
+
+class RateLimitedPasswordResetView(PasswordResetView):
+    @method_decorator(ratelimit(key="ip", rate="5/h"))
```
DeD1rk, line 60: Wouldn't
```python
    @method_decorator(ratelimit(key="ip", rate="5/h"), name="post")
    class RateLimitedPasswordResetView(PasswordResetView):
        pass
```
be just a bit cleaner?
But also I don't really care if you'd rather not do the effort of changing it

DeD1rk (approved):

website/registrations/tests/test_views.py
JobDoesburg: I changed the admin login view to redirect to the regular login view, which redirects again. So it's a 302 to a 302. Idk but without it, the tests failed

website/thaliawebsite/views.py
JobDoesburg: I thought it didn't work like this but now I see it does

PR #2998, opened 2023-04-12
Prefetch thumbnail metadata from database

### LINKED ISSUE #2986: N+1 query for thumbnails
There is a bit of an N+1 problem getting thumbnails:

Sentry Issue: [CONCREXIT-NP](https://thalia.sentry.io/issues/4028393801/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Transaction Name** | db - SELECT "thumbnails_thumbnailmeta"."id", "t... |
| **Parent Span** | view.render - api:v2:photos:album-list |
| **Repeating Spans (55)** | event.django - django.db.models.fields.files.Im... |

ColonelPhantom:
Closes #2986 

### Summary
I created a function that prefetches thumbnail metadata from the database for a list of images.

### TODO
This function still needs to be used in more places, and the relevant size limit also needs to be added when calling the function like in members/views.py

### How to test
1. Go to `/members/directory/`
2. Open the query list in the debug toolbar
3. See one big query instead of many small ones for fetching thumbnail info:
image

(Note: it would be better if we could fetch these in-line, but that's probably not going to happen and this single big query is still a massive improvement.)

### DIFF
--- a/website/events/api/v2/serializers/event.py
+++ b/website/events/api/v2/serializers/event.py
@@ -120,3 +120,37 @@ class EventSerializer(CleanedModelSerializer):
 
     def _maps_url(self, instance):
         return create_google_maps_url(instance.map_location, zoom=13, size="450x250")
+
+
+class EventListSerializer(EventSerializer):
+    class Meta:
+        model = Event
+        fields = (
+            "pk",
+            "slug",
+            "url",
+            "title",
+            "description",
+            "caption",
+            "start",
+            "end",
+            "category",
+            "registration_start",
+            "registration_end",
+            "cancel_deadline",
+            "optional_registrations",
+            "location",
+            "price",
+            "fine",
+            "num_participants",
+            "max_participants",
+            "no_registration_message",
+            "registration_status",
+            "cancel_too_late_message",
+            "has_fields",
+            "food_event",
+            "maps_url",
+            "user_permissions",
+            "user_registration",
+            "documents",
+        )
--- a/website/events/api/v2/views.py
+++ b/website/events/api/v2/views.py
@@ -25,6 +25,7 @@ from events.models.external_event import ExternalEvent
 from events.services import is_user_registered
 from thaliawebsite.api.v2.permissions import IsAuthenticatedOrTokenHasScopeForMethod
 from thaliawebsite.api.v2.serializers import EmptySerializer
+from utils.media.services import fetch_thumbnails_db
 
 
 class EventListView(ListAPIView):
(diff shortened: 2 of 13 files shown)


website/events/api/v2/views.py
```diff
@@ -124,6 +125,11 @@ def get_queryset(self):
             ).select_related("member__profile")[: self.event.max_participants]
         return EventRegistration.objects.none()
 
+    def get_serializer(self, registrations=None, *args, **kwargs):
```
ColonelPhantom, line 128: I overrode get_serializer in several places, I'm not sure it's the best function to do this but get_queryset is called twice for some reason.

website/partners/api/v2/views.py
```diff
@@ -18,6 +19,11 @@ class PartnerListView(ListAPIView):
         framework_filters.OrderingFilter,
         framework_filters.SearchFilter,
     )
+
+    def get_serializer(self, partners, *args, **kwargs):
```
ColonelPhantom, line 23: For some reason the vacancy API does not show thumbnails but rather the full company_logo?

ColonelPhantom:
I think I have basically all views that deal with (lots of) images covered!

website/events/api/v2/serializers/event.py
```diff
@@ -120,3 +120,37 @@ def _url(self, instance: Event):
 
     def _maps_url(self, instance):
         return create_google_maps_url(instance.map_location, zoom=13, size="450x250")
+
+
+class EventListSerializer(EventSerializer):
```
DeD1rk, line 125: This should probably be handled in the parent with `many=True`

website/photos/views.py
```diff
@@ -132,9 +137,11 @@ class LikedPhotoView(LoginRequiredMixin, PagedView):
     context_object_name = "photos"
 
     def get_queryset(self):
```
DeD1rk, line 139: In a few places the change is in `get_queryset`. Does that work? I guess as long as the returned queryset is not modified outside of get_queryset it would work on the cached contents, but if it changes at all it would still do N+1 afterwards

ColonelPhantom: I tested using the Django debug toolbar and it seems to work. I agree that it's not a perfect solution though, maybe if there was something similar to `prefetch_related` for custom code?

website/events/api/v2/serializers/event.py
ColonelPhantom: Other places in the code do this too.

website/photos/views.py
DeD1rk: Yeah there isn't... If it had all been nicely made with models instead non-model python objects it would be possible with a Prefetch object, but alas
So I guess this is still the best

website/partners/templatetags/frontpage_vacancies.py
```diff
@@ -10,7 +10,7 @@
 def render_frontpage_vacancies():
     vacancies = []
 
-    for vacancy in Vacancy.objects.order_by("?")[:6]:
+    for vacancy in Vacancy.objects.order_by("?")[:6].select_related("partner"):
```
DeD1rk, line 13: This should be before the slicing right?

DeD1rk (approved):

PR #3015, opened 2023-04-19
Add production docker build and terraform config

DeD1rk:
This should make everything ready to deploy to production. It should cause a tag event to create docker images tagged with `latest`, add automatic updating with ssh, and adds the necessary terraform setup to create the new infrastructure.

Note that for creating the infra, we'll need to import the existing s3 bucket.


.github/workflows/build.yaml
```diff
@@ -119,3 +141,51 @@ jobs:
...
+      - name: Update env files
+        run: |
+          rsync infra/.env deploy@thalia.nu:/var/lib/concrexit/.env
+          rsync infra/env.public deploy@thalia.nu:/var/lib/concrexit/env.public
+          rsync infra/env.secret deploy@thalia.nu:/var/lib/concrexit/env.secret
+          rsync infra/docker-compose.yml deploy@thalia.nu:/var/lib/concrexit/docker-compose.yml
+
+      - name: Restart containers
+        run: |
+          ssh deploy@staging.thalia.nu << EOF
```
JobDoesburg, line 185: This is wrong

JobDoesburg (changes requested):

PR #3016, opened 2023-04-19
Move static files to S3

JobDoesburg:


DeD1rk:
I did a whole lot of trying stuff out on master (yes i changed master history to remove it all afterwards sorry) but I can't seem to get rid of this problem in `manage.py compress`:
`CommandError: An error occurred during rendering photos/liked-photos.html: Missing staticfiles manifest entry for 'CACHE/css/output.a97b4b22fee8.css'` (or some other css output). Those files are being created (saved both on S3 and locally) but don't show up in the manifest.

PR #3022, opened 2023-04-19
Prevent boards with no start or stop date

### LINKED ISSUE #2991: Boards can have `until` field be `None` (and that should be possible)
Sentry Issue: [CONCREXIT-RJ](https://thalia.sentry.io/issues/4079745334/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'year'
(28 additional frame(s) were not displayed)
...
  File "activemembers/models.py", line 206, in get_absolute_url
    "activemembers:board", args=[str(self.since.year), str(self.until.year)]
```

LisanneWeidmann:
Closes #2991

bug

### Summary
Added a clean function inside Board in website/activemembers/models.py which checks for the "since" and "until" date of a newly created board. Message is put on screen when (either one of) these fields are empty.

### How to test
1. Go to 'Active Members' and then 'Boards'
2. Click on 'add'
3. Fill the form in until 'Founded in' and 'Existed until'.
4. Try to leave 'Founded in' and 'Existed until' empty.
5. You should now see a message saying that these fields cannot be empty.

### DIFF
--- a/website/activemembers/models.py
+++ b/website/activemembers/models.py
@@ -201,6 +201,23 @@ class Board(MemberGroup):
         self.active = True
         super().save(**kwargs)
 
+    def clean(self):
+        if self.since is None:
+            raise ValidationError(
+                {
+                    "since": _("Please insert a starting year for the new board."),
+                }
+            )
+
+        if self.until is None:
+            raise ValidationError(
+                {
+                    "until": _(
+                        "Please insert the year until when the board is active."
+                    ),
+                }
+            )
+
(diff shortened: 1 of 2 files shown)


JobDoesburg:
Well, the thing is that it should be possible actually! It should be possible to create a board without end date (and even without a start date).

JobDoesburg (changes requested):
Well, the thing is that it should be possible actually! It should be possible to create a board without end date (and even without a start date if you ask me, like for the candidate board)

PR #3033, opened 2023-04-22
Make public storage not need/use signed urls

DeD1rk:
### Summary
This makes cloudfront allow serving `/public/*` files without a signed url, and modifies the PublicS3Storage to not output signed urls. There is an exception when the ReponseContentDisposition param is used, as that is only supported in signed urls.

This should help make newsletters and other emails not have their iamge urls expire and makes public media url's shareable.

### How to test
1. Merge it and try out whether staging works nicely. 
2. Afaik there's no public storage file we allow you to download as attachment right now, but you could try get_media_url on e.g. a profile photo from the shell. (btw, is there any reason we have profile photos public? the member list isn't public...)


website/newsletters/emails.py
```diff
@@ -38,32 +33,23 @@ def send_newsletter(newsletter):
...
+    context = {
+        "newsletter": newsletter,
+        "agenda_events": events,
+        "main_partner": main_partner,
+        "local_partners": local_partners,
+    }
+
+    try:
+        send_email(
+            to=[f"newsletter@{settings.GSUITE_DOMAIN}"],
```
JobDoesburg, line 45: Maybe import from mailinglist app here

website/newsletters/emails.py
```diff
@@ -38,32 +33,23 @@ def send_newsletter(newsletter):
...
-            logger.exception("Failed to send the %s newsletter", language[1])
+    context = {
+        "newsletter": newsletter,
+        "agenda_events": events,
+        "main_partner": main_partner,
+        "local_partners": local_partners,
+    }
+
+    try:
+        send_email(
```
JobDoesburg, line 44: I think this won't reuse the smtp connection, so it will be very slow. Maybe we should update `send_email` with some caching on the smtp connection

JobDoesburg (commented):
You're doing more than just the S3 urls in this PR...

Also, do we want to remove translations completely?

website/newsletters/emails.py
DeD1rk: We could do some global long term connection caching but if that's useful I reckon django would support that better

website/newsletters/emails.py
DeD1rk: Best we could do is take the list from https://github.com/svthalia/concrexit/blob/71936f46d69b275e6dcdc450e5312daf9f23fadd/website/mailinglists/services.py#L183, but thats a dict by name, so to get the item we'd need the name which is directly the 'user' part of the address. So there's little to do

DeD1rk:
> You're doing more than just the S3 urls in this PR...

Hmm yeah I guess so, I can split it up, but also that's so much effort and lazy me thinks they're somewhat related

> Also, do we want to remove translations completely?

I do, but thats just me 😂

Scarletto (approved):
lgtm

PR #3035, opened 2023-04-23
Remove language from pushnotifications Device

DeD1rk:
### Summary
The code was messy for no reason, and this should solve https://thalia.sentry.io/issues/3988073545/?project=1463433&query=is%3Aunresolved+issue.category%3Aerror&referrer=issue-stream&statsPeriod=14d&stream_index=0.


website/pushnotifications/models.py
```diff
@@ -139,56 +132,49 @@ def send(self, **kwargs):
...
+                notification=messaging.Notification(
+                    title=data["title"],
+                    body=data["body"],
+                ),
+                data=data,
+                android=messaging.AndroidConfig(
+                    ttl=datetime.timedelta(seconds=ttl),
+                    priority="normal",
+                    notification=messaging.AndroidNotification(
+                        color="#E62272",
```
JobDoesburg, line 159: This should not be hardcoded

website/pushnotifications/models.py
```diff
@@ -60,12 +59,6 @@ class Device(models.Model):
     date_created = models.DateTimeField(
         verbose_name=_("registration date"), auto_now_add=True, null=False
     )
-    language = models.CharField(
```
JobDoesburg, line 63: Do we really want to remove this?

DeD1rk: Gladly, hasn't been used (or set properly) for years

PR #3044, opened 2023-04-25
Fix typos in documents page

### LINKED ISSUE #3037: Fix "The statues have last been amended in September 2007"
See https://thalia.nu/association/documents/. This text is outdated. We can just drop the line.

DeD1rk:
Closes #3037.


JobDoesburg:
Do we really want to remove the translation tags? At all other places in the website I did keep them

PR #3070, opened 2023-05-03
Document API serializer returns signed cdn URL 2: Electric Boogaloo

### LINKED ISSUE #3050: Expand/improve document part of event endpoint
### Describe the change
Change the current URL or add a URL to a link to the actual document (that is, one that ends with `/[name].pdf`).

### Motivation
It's very nice for the app to be able to just fetch files directly.
Also, I don't think the current URL is even necessary because you also have the PK with which you can create the URL.

### Current implementation
There is a URL that leads to the file, but only after a redirect to `cdn`.

Scarletto:
Closes #3050

### Summary
Changed the document API serializer to return a signed cdn URL if the file is public or the user is authenticated and authorized.

### How to test
1. Go to the API page of an event with an attached document (you must be logged in for this)
2. See the URL is now the actual cdn URL of the file

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,11 +20,11 @@ repos:
   hooks:
   - id: isort
 - repo: https://github.com/psf/black
-  rev: "23.1.0"
+  rev: "23.3.0"
   hooks:
   - id: black
     exclude: .*/migrations/.*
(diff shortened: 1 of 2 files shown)


DeD1rk (approved):
This time I'm pretty sure this is okay

PR #3077, opened 2023-05-05
Face detection (with lambda)

DeD1rk:
Supersedes #2603

### Summary
Implements face detection as a separate django app in concrexit, making use of [svthalia/concrexit-facedetection-lambda](https://github.com/svthalia/concrexit-facedetection-lambda) to perform extraction of face encodings.

### How to test
You can test this locally, but that's not trivial:
1. Build a slightly modified lambda image, adding `verify=False` in https://github.com/svthalia/concrexit-facedetection-lambda/blob/6a254af7298d5b37ac73052d1fac0151cb53fdde/app.py#L27.
  This will prevent SSLErrors in the lambda when working with a local concrexit.
2. Replace https://github.com/svthalia/concrexit/blob/76eda9504c2e0188a880c817de2810c95fcfc035/website/facedetection/services.py#L74-L88 with:

```python
        response = requests.post(
            settings.FACEDETECTION_LAMBDA_ARN,
            json=payload,
        )

        if response.status_code != 200:
            raise Exception("Couldn't invoke Lambda function by url.")
```
3. Run the modified docker image with `docker run -p 9000:8080 concrexit-face-detection-lambda`
4. Run concrexit with `FACEDETECTION_LAMBDA_ARN="http://localhost:9000/2015-03-31/functions/function/invocations" BASE_URL="http://host.docker.internal:8000" ./manage.py runserver`

5. Now concrexit should be working with your local lambda emulator. Might take some effort to figure out the details though. I don't expect we'll be touching any of this a lot later, so I ended up not including support for local running in the code. For normal local development just never extracting encodings seems fine to me.

Anyway, the only way to really be sure this works would be to test it with a real lambda. So we can pretty much just try it on staging when the code looks good. 

### TODO:
- [x] Triggering analysis of albums/photos (Made a nightly management command, but I would like a way to also trigger it after an album has been uploaded.)
- [x] Add trigger after album has been uploaded (a custom signal sent from the album admin after saving is done?)
- [x] Admins
- [x] What to do when there is no lambda configured?... (It's logged but ignored. Same goes for failures, so in the lambda being down shouldn't ever crash concrexit).
- [x] A readme documenting the facedetection app
- [x] Modify media url expiration time to allow longer queuing (lambda calls may be queud for up to 6 hours by default, but right now after 1 hour the urls will break)
- [ ] Tests


JobDoesburg:
Looks good, sad that my commits are gone :(

JobDoesburg:
> > Looks good, sad that my commits are gone :(
> 
> Will be squashed right, so if you're set as coauthor on the squashed commit it doesnt matter?

Yeah I guess, though I'm always surprised by that these commits stats never work the way I expect them to work. /care

website/facedetection/management/commands/trigger_facedetection_lambda.py
```diff
@@ -0,0 +1,66 @@
...
+                [
+                    FaceDetectionPhoto(photo=photo)
+                    for photo in Photo.objects.filter(facedetectionphoto__isnull=True)[
+                        :400
+                    ]
+                ]
+            )
+
+            self.stdout.write(f"Submitting {len(photos)} photos.")
+            trigger_facedetection_lambda(photos)
```
JobDoesburg, line 66: I would rather move the logic here to services.py and only invoke in this file. So we can easily trigger a sync in a different way too in the future

website/facedetection/migrations/0004_referenceface_created_at.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.AddField(
+            model_name="referenceface",
+            name="created_at",
+            field=models.DateTimeField(
+                auto_now_add=True, default=django.utils.timezone.now
+            ),
+            preserve_default=False,
+        ),
+    ]
```
JobDoesburg, line 21: Let's merge the migrations

website/facedetection/models.py
```diff
@@ -0,0 +1,332 @@
...
+    user = models.ForeignKey(
+        FaceDetectionUser,
+        on_delete=models.CASCADE,
+        related_name="reference_faces",
+    )
+
+    file = ImageField(upload_to=reference_face_uploadto)
+
+    created_at = models.DateTimeField(auto_now_add=True, editable=False)
+    marked_for_deletion_at = models.DateTimeField(null=True, blank=True)
```
JobDoesburg, line 101: Why is this a separate field? We always compute this right?

DeD1rk: marked_for_deletion_at? What would we be able to compute that from?

DeD1rk:
> One more thing: where is the contents of the facedetection-lambda itself?

https://github.com/svthalia/concrextit-facedetection-lambda

website/facedetection/models.py
JobDoesburg: Nvm, I forgot for what we use them.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ beautifulsoup4 = "^4.11.1"
 django-storages = "^1.12.3"
 boto3 = "^1.23.7"
 gunicorn = "^20.1.0"
+requests = "^2.30.0"
```
JobDoesburg, line 40: I think we do not use this actually

website/facedetection/models.py
```diff
@@ -0,0 +1,332 @@
...
+from members.models import Member
+from photos.models import Photo
+
+
+class FaceDetectionUser(Member):
+    class Meta:
+        proxy = True
+
+
+def authentication_token() -> str:
```
JobDoesburg, line 21: Technically, this is a authorization token and not an authentication token

website/photos/admin.py
```diff
@@ -8,6 +9,8 @@
 from .models import Album, Like, Photo
 from .services import extract_archive, save_photo
 
+uploaded_photos = Signal()
```
JobDoesburg, line 12: Maybe we should refactor this to `album_extracted` or `album_uploaded`

pyproject.toml
DeD1rk: Hmm yeah that's true, I only used it for testing with a docker image

website/facedetection/models.py
DeD1rk: Hmm depends on how it's used though. changed it to secure_token

JobDoesburg (approved):

PR #3093, opened 2023-05-11
HTTP redirect from events/next to the next Event

### LINKED ISSUE #2871: Add http endpoint for the next event
We want to use a dns redirect to redirect to the first upcoming Thalia event. 

An endpoint (like /events/next) that redicects to the next Thalia event 
### Motivation
We want to setup a domain (sog.je) that redirects to this 

Static html host that uses javascriot, but this would require using additional resources

Brambleberry4:
Closes #2871

### Summary
The link thalia.nu/events/next will now redirect via HTTP to the next Thalia event. The next Thalia event is defined as the event whose end is closest to the current time. Furthermore, a 404 page is given in case there is no next event. 

### How to test
1. Type in the URL: thalia.nu/events/next
2. This should redirect to the next upcoming event
3. For the case that there is no further event, a 404 page with the Hint "There is no upcoming event." is given.


DeD1rk:
Whoops something went wrong with the PR

DeD1rk:
The PR is the wrong way around, youre trying to merge master into your branch, instead of your branch into master

Brambleberry4:
So delete and write again?

Issue #2630, opened 2022-11-02
Proxy models and admin views for giving certain committees / apps access to certain parts of members database


JobDoesburg:
### Describe the change
Create proxy models for the member model, like we did for PaymentUsers. And then create custom admin views displaying certain parts of this data.

### Motivation
Think of:
- a SalesUsers proxy model with an admin that displays the people that are underage, and their order history
- a EventsUser proxy model with an admin that displays the people that are allowed to attend events and their registration history
    - The events app should then get a model for keeping the events blacklist with a OneToOne relation to the users model
- a ThabloidUser proxy model with an admin that displays the people that indicated they want to receive the Thabloid
    - The thabloid app should then get a model for keeping the opt-in with a OneToOne relation to the users model
- an ActiveMembers proxy model with an admin that displays all memberships of a person (only the currently members and people that were active in the past)

- Possibly also relevant for partners app (with the opt-in partner mailing list), food orders app (with everyone who ordered food), newsletters (with the newsletter opt-in).

This is all similar to the situation we have with blacklisted TPay payment users. 

This has some benefits:
- Centralising code in the proper apps the code belongs to
- More granular permission management, like required for #2619
- Easy way to export more data. Like checking a person's event registration history

### Current implementation
- Only board has access to the members data, it is not possible to give other people easy access to parts of the members data
- Certain code is included in the members profile though it doesn't belong there.

JobDoesburg:
See also #2592 and #2565

PR #3105, opened 2023-05-16
Thabloid proxy user

Noah-marc:
Part of #2630 

### Summary

I added the models proxy model ThabloidUser and the BlacklistedThabloidUser in the thabloid app.  Furthermore, I created an Admin for displaying all the ThabloidUsers and showing whether they want to receive it or not. The user can change the value of the field when editing his profile. 

### How to test
1. Go to the ThabloidUser admin to see the current status of the field
2. Edit the field of the user on the edit_profile page
3. Go back to the ThabloidUser to see whether the field was actually changed.


website/members/forms.py
```diff
@@ -5,12 +5,20 @@
...
+
 from .models import Profile
 
 
 class ProfileForm(forms.ModelForm):
     """Form with all the user editable fields of a Profile model."""
 
+    receive_thabloid = forms.BooleanField(
+        required=False,
+        label="Receive thabloid",
```
JobDoesburg, line 18: We actually don't want to define this field in the `members` app at all. It is already an improvement that it's not in the `members` database anymore, but now we have a dependency from `members` onto `thabloid` (while it can only exist the other way around). 

So this should be dynamically defined. For example, via some registration of `profile fields`

website/thabloid/migrations/0007_thabloid_cover_alter_thabloid_file.py
```diff
@@ -40,21 +40,21 @@ class Migration(migrations.Migration):
...
             ),
         ),
         migrations.RunPython(populate_thabloid_cover, migrations.RunPython.noop),
         migrations.AlterField(
             model_name="thabloid",
             name="cover",
-            field=models.ImageField(upload_to=thabloid.models.thabloid_cover_filename),
+            field=models.ImageField(upload_to=thabloid_cover_filename),
         ),
     ]
```
JobDoesburg, line 60: These migrations shouldn't change

JobDoesburg (changes requested):
Nice work on the proxy model!
If you want me to, I can have a look at the dynamic form fields

JobDoesburg (dismissed):
Nice work on the proxy model!
If you want me to, I can have a look at the dynamic form fields

JobDoesburg:
I've implemented the dynamic fields! @DeD1rk can you review

website/thabloid/services.py
```diff
@@ -0,0 +1,8 @@
+from thabloid.models import ThabloidUser
+
+
+def update_thabloid_blacklist_for_user(form, instance, commit, *args, **kwargs):
```
DeD1rk, line 4: This might as well be a single method or even setter on ThabloidUser

website/thabloid/apps.py
```diff
@@ -21,3 +22,20 @@ def menu_items(self):
                 }
             ],
         }
+
+    def user_profile_form_fields(self, instance=None):
+        from thabloid.models import ThabloidUser
```
DeD1rk, line 27: Do these really need to be deferred imports? I think it shouldn't be required for apps to be initialized jsut to import models

website/thabloid/apps.py
```diff
@@ -21,3 +22,20 @@ def menu_items(self):
...
+        from thabloid.services import update_thabloid_blacklist_for_user
+
+        default_value = False
```
DeD1rk, line 30:
```suggestion
        default_value = True
```

DeD1rk (commented):
That's pretty clean :)
I'll probably go over some leftover/ugly imports and then its perfect

DeD1rk (approved):

JobDoesburg:
Well now we should also display the addresses and implement the CSV download before merging it

PR #3114, opened 2023-05-17
Refactor data minimization methods

JobDoesburg:


DeD1rk:
I like it! CI is mad mad but looks great!

PR #3115, opened 2023-05-17
Former member api

### LINKED ISSUE #3067: Members API doesn't list all members
### Describe the bug
Luko, pk=23 on staging, is listed as member on the website, but is not a member according to the API

### How to reproduce
No idea.
check https://staging.thalia.nu/members/profile/23 vs https://staging.thalia.nu/api/v2/members/23/

### Expected behaviour
These urls should agree on the memberness of Luko

ColonelPhantom:
Closes #3067 

### Summary
- Allow MemberDetailView to display former members
- Add a FormerMemberFilter to the MemberListView

### How to test
1. Go to `/api/v2/members/<id>`  for a former member
2. Observe that the page loads
3. Go to `/api/v2/members/?former={any|true|false}`
4. See that there are {also/only/none} former members in the list

### DIFF
--- a/website/members/api/v2/filters.py
+++ b/website/members/api/v2/filters.py
@@ -1,4 +1,9 @@
-from rest_framework import filters
+from datetime import datetime
+
+from django.db.models import Q
+from django.utils import timezone
+
+from rest_framework import filters, serializers
 
 from members.models import Membership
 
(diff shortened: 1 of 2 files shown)


JobDoesburg:
I don't think former members should be accessible via the API! There are no grounds for making that info public

ColonelPhantom:
But they are accessible via the website. There are also dead references due to this in the API and app on staging, see the issue this PR closes.

Also, I'm pretty sure the website and API member lists both require being signed in?

DeD1rk:
> Also, I'm pretty sure the website and API member lists both require being signed in?

Yes, although some committees (i.e. partnercie, maybe some others) do show their members publicly with name and avatar, but not the full profile.

website/members/api/v2/filters.py
```diff
@@ -28,6 +33,62 @@ def get_schema_operation_parameters(self, view):
...
+
+            # Filter out all current active memberships
+            memberships_query &= Q(type=Membership.MEMBER) | Q(type=Membership.HONORARY)
+            memberships = Membership.objects.filter(memberships_query)
+            members_query &= ~Q(pk__in=memberships.values("user__pk"))
+
+            memberships_query = Q(type=Membership.MEMBER) | Q(type=Membership.HONORARY)
+            memberships = Membership.objects.filter(memberships_query)
+            all_memberships = Membership.objects.all()
+            # Only keep members that were once members, or are legacy users
```
DeD1rk, line 64: This seems a bit complicated and I don't expect it to scale well: We first query memberships to python, and then filter the members by a _huge_ set of memberships. We can at least simplify it to use an Exists subquery instead of a full query on memberships: `qs.filter(Exists(Membership.objects.filter(<types>, user=OuterRef("pk"), <time>)))`

DeD1rk: Of course we do the same on the website memberlist, but that's also inefficient if there are many memberships

ColonelPhantom: We do not query to python, according to my SQL trace in DJDT. It seems the Django query optimizer is smart enough to turn the queryset into a subquery thanks to laziness.

DeD1rk (approved):
okay sick secret optimization

PR #3116, opened 2023-05-17
Made comparisson case insensitive.

### LINKED ISSUE #3099: Make mailinglist sync case-insensitive
### Describe the bug
I was checking some google admin audit logs and saw that the gsuit sync keeps removing and re-adding some people with capitals in their email adress to the mailinglists they're in. 

For example, if I have Dirk@gmail.nl, google would normalize that to dirk@gmail.com. When the sync runs, it determines the deisred group to be `["Dirk@gmail.com"]`. It requests the current members of a group, getting `["dirk@gmail.com"]`, then remove `dirk@gmail.com` and add `Dirk@gmail.com`. This clearly has no effect, but is just inefficient and ugly. And it might cause some issues because the people in question leave their groups for a few seconds every hour.

### Expected behaviour
capitalized email addresses are treated the same as their lowercase version.

### Screenshots
image

LisanneWeidmann:
Closes #3099

Now the Google mailinglist and Thalia mailinglist are compared in lowercase. So no longer weird behavior when comparing uppercase mailing addresses with lowercase ones.

Changes made in gsuite.py in malinglists.

### DIFF
--- a/website/mailinglists/gsuite.py
+++ b/website/mailinglists/gsuite.py
@@ -289,15 +289,15 @@ class GSuiteSyncService:
                 members_list += members_response.get("members", [])
 
             existing_members = [
-                m["email"] for m in members_list if m["role"] == "MEMBER"
+                m["email"].lower() for m in members_list if m["role"] == "MEMBER"
             ]
             existing_managers = [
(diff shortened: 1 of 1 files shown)


DeD1rk:
TOP (test in production)

PR #3151, opened 2023-05-31
Rotate photo's when saving them to the server

### LINKED ISSUE #3133: Actually rotate photos
Currently we store in the db how photos should be rotated (it's part of exif metadata on jpegs or something, and gets lost when e.g. thumbnailing). It would be so much nicer to just have all photos be upright (rotating them) as soon as we save them. 
Then other places where we use photos don't have to consider rotation any more.

rowan-vr:
Closes #3133 

### Summary
Rotate photo's when saving them to the server

#### Uploading new images
1. Go to the admin panel
2. Upload a new album
3. Upload a rotated jpg image
4. Finish uploading the album and observe

#### Migrating old images
1. Have some old rotated images before checking out on this branch
2. Run `python website/manage.py photos_rotate_migrate` in the poetry environment
3. Observe the fixed images

> Note regarding migrations, we do not have the data about if an image is mirrored or not, therefore they will only be rotated and not be mirrored.

*Useful images for testing: https://github.com/recurser/exif-orientation-examples*


website/photos/management/commands/photos_rotate_migrate.py
```diff
@@ -0,0 +1,56 @@
...
+        photo.file = InMemoryUploadedFile(
+            content,
+            None,
+            image_path,
+            "image/jpeg",
+            content.tell,
+            None,
+        )
+
+        photo.save()
```
DeD1rk, line 53: We may need to do something to make sure this overwrites the old photo in S3 (I'm not sure what yet)

website/photos/services.py
```diff
@@ -46,6 +46,29 @@ def photo_determine_rotation(pil_image):
     return 0
 
 
+def photo_determine_mirrored(pil_image):
+    EXIF_ORIENTATION = {
+        1: 0,
```
DeD1rk, line 51: True and False seem more accurate for being mirrored or not

website/photos/services.py
```diff
@@ -173,11 +196,19 @@ def save_photo(photo_obj, file, filename):
     ):
         return False
 
+    rotation = photo_determine_rotation(image)
+    mirrored = photo_determine_mirrored(image)
```
DeD1rk, line 200: Unrelated to this specific piece of code but I think it might be possible to move all of this into Photo.save(). I'll probably look into that myself once I find some time

DeD1rk:
Closing in favour of #3324.

PR #3185, opened 2023-06-14
Don't pregenerate thumbnails for photos

DeD1rk:
Paparazcie requested this as they prefer this over having the save timeout.


ColonelPhantom:
Would it maybe be possible to automatically launch a background task to generate the thumbnails (or just request the page)? That way we do not need to ask users to visit the album page.

DeD1rk:
I would like a nicer way but am going to release. Anyway I don't really like an idea like 'run a management command in a subprocess, and would like to get a cleaner approach at some point

PR #3216, opened 2023-06-28
Implement app-based user menu

### LINKED ISSUE #3158: Use same menu rendering as for main menu for user dropdown menu as well
### Describe the change
Use same menu rendering as for main menu for user dropdown menu as well

### Motivation
Consistency

### Current implementation
Just html tags in a `thaliawebsite` app template

### Suggested implementation
Register menu items in each apps.py

ColonelPhantom:
Closes #3158 

### Summary
This PR separates out the user menu so that entries are stored in the apps, just like the general menu.
Additionally, the currently active page in the user menu is highlighted, again just like the general menu.

### How to test
1. Go to a page
2. See that the menu still contains all entries
3. Click a link in the menu
4. See that it colors magenta


DeD1rk (changes requested):
I like it! However, this shows the same items to every user. The site administration button should only be shown to staff. So I guess the appconfigs should be allowed to return some kind of lambdas? E.g. an optional lambda per item taking a request, returning whether or not to show the item. Or a single lambda returning additional items (that's more flexible maybe but makes caching a bit harder).

DeD1rk (approved):
Very nice!

PR #3218, opened 2023-06-29
Add facedetection api endpoints

### LINKED ISSUE #3101: Add facedetection API
I'm always frustrated when I use ThaliApp but can't use facedetection from there.

A few API endpoints so that ThaliApp can support face detection.
For example:
- `api/v2/facedetection/your-photos/` (GET)
- `api/v2/facedetection/reference-faces/[:id/]` (GET,POST,DELETE)

### Motivation
Then we can have feature parity in the app. And being able to upload references from your phone makes it so much easier to use facedetection.

JeeVee11:
Closes #3101 

### Summary
Adds API endpoints to view the users facedetection matches and see/create/delete their reference faces.

### How to test
1. Try every added endpoint.
2. Test that the matches can't be seen, and reference faces can't be created, when not a member


website/facedetection/api/v2/views.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+        return self.list(request, *args, **kwargs)
+
+    def get_serializer(self, *args, **kwargs):
+        if len(args) > 0:
+            photos = args[0]
+            fetch_thumbnails([photo.file for photo in photos])
+        return super().get_serializer(*args, **kwargs)
+
+    def get_queryset(self):
```
ColonelPhantom, line 38: The original view has the same logic. We should probably deduplicate that so that changing this logic is not painful.

ColonelPhantom (changes requested):
We should probably deduplicate the logic here.

website/facedetection/api/v2/views.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def get_serializer(self, *args, **kwargs):
+        if len(args) > 0:
+            reference_faces = args[0]
+            fetch_thumbnails([reference.file for reference in reference_faces])
+        return super().get_serializer(*args, **kwargs)
+
+    def create(self, request, *args, **kwargs):
+        if request.member.current_membership is None:
+            raise PermissionDenied(detail="You need to be a member to use this feaure.")
```
T8902, line 59:
```suggestion
            raise PermissionDenied(detail="You need to be a member to use this feature.")
```

T8902 (approved):

PR #3229, opened 2023-07-05
Optimize member API views

### LINKED ISSUE #3228: N+1 Query on listing memberships due to membership_latest
Sentry Issue: [CONCREXIT-YM](https://thalia.sentry.io/issues/4256012007/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Offending Spans** | db - SELECT "members_membership"."id", "members...

ColonelPhantom:
Closes #3228 

### Summary
I added support for prefetching the `latest_membership` property. (Perhaps I should make it so that it becomes cached too.)

The code isn't the prettiest, so maybe this can also be refactored in some way (e.g. by putting the `Prefetch` object in the model somewhere?)

I believe that using a subquery will not work because Django does not support putting entire ORM objects in a subquery, only single values :disappointed: 

### How to test
1. Go to /api/v2/members
2. Observe that the results are still correct.
3. Observe that the number of queries has gone down significantly.

### DIFF
--- a/website/members/api/v2/views.py
+++ b/website/members/api/v2/views.py
@@ -1,5 +1,6 @@
 """API views of the activemembers app."""
 
+from django.db.models import Prefetch
 from django.shortcuts import get_object_or_404
 
 from oauth2_provider.contrib.rest_framework import IsAuthenticatedOrTokenHasScope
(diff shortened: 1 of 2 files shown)


website/members/api/v2/views.py
```diff
@@ -81,4 +91,4 @@ class MemberCurrentView(MemberDetailView, UpdateAPIView):
     }
 
     def get_object(self):
-        return get_object_or_404(Member, pk=self.request.user.pk)
+        return get_object_or_404(MemberDetailView.queryset, pk=self.request.user.pk)
```
DeD1rk, line 94: Maybe this can actually be self.get_queryset()?

DeD1rk (approved):
Nice

PR #3324, opened 2023-09-05
Overhaul image processing

### LINKED ISSUE #988: Deliver sliders or other large images as webp

### LINKED ISSUE #1245: Move photos to S3 (fullres)
Currently there are 2 places where we store photos: the Paparazcie photo archive on a shared G Suite drive, and the low res photos on concrexit. This is inconvenient, as we want to keep those 2 in sync but that needs to be done manually, and because people need to ask the Paparazcie for fullres files.

As we are on AWS now, we could decide to move our photos to S3.

### Motivation
- One single place and interface for Thalia's photos
- Everyone can easily access fullres photos
- Less server crashes whenever anyone tries to download an album

What we have now.

### LINKED ISSUE #2936: Thumbnails are of low quality
### Describe the bug
Some thumbnails are of really low quality, even though the resolution is set higher.

### How to reproduce
Check out the member list. Note that it is not for every image.

### Expected behaviour
Don't be blurry. 

### Screenshots
image

### Additional context
Probably some settings about how the thumbnails are generated should be changed.

### LINKED ISSUE #3133: Actually rotate photos

### LINKED ISSUE #3408: Incorrect rotation on profile photos
### Describe the bug
In our board page Widad's photo is rotated 90 degrees, while on their profile it is correct.

DeD1rk:
Closes #3133.
Closes #2936.
Closes #1245.
Closes #988.
Closes #3408.

Fixes https://thalia.sentry.io/issues/4305077187/ (and a similar event from a while back with MPO format).

### Summary
I made custom processors to get past the limitations of django-thumbnails' builtin ones. 
Using `resize_source_to`, everything that was done in `save_photo` etc. can now be done by the source processor, and model `.save()` methods.

I've added a management command that we can use to clear the `Photo.rotation` of existing photos, as any newly saved ones don't need that.

I changed thumbnailing to use the exact same modifications that we made before django-thumbnails. All thumbnails are now saved to WebP (all apple devices released since 2013 support it).

### How to test
1. Try any way you can upload photos (zipped, tarred, single photo directly, facedetection reference, profile pic, etc.) with various pictures. For example, try out pictures with EXIF rotation (https://github.com/recurser/exif-orientation-examples). And photos in peculiar formats, and corrupted photos that should be rejected, and not crash the server.

2. Try out the management command locally (make sure there are rotated pictures before).

3. Check the quality of generated thumbnails, to check that #2936 really is solved by this. if not, perhaps the webp quality can be increased to e.g 95.

4. When everything looks good, we have to test the management command on staging, to make sure the S3 file overwriting works. So again we should make sure there are some rotated photos on there.

### TODO:
- [x] Use a custom processor for thumbnailing too (instead of the builtin 'resize'). Hopefully this can fix the reported low quality.
- [x] Use a different 'source' size for partner images, that can be transparent.
- [x] Fix current test.
- [ ] Add some new tests.


T8902 (commented):
Duplicate photo to the clean maybe?

website/utils/media/processors.py
```diff
@@ -0,0 +1,66 @@
...
+    Warning: for django-thumbnails to save the image with the right filename,
+    the '<size>.FORMAT' key in settings.THUMBNAILS_SIZES must be set to 'webp'.
+
+    """
+    pil_image = image.get_pil_image()
+
+    size = kwargs["size"]
+    if not kwargs["cover"]:
+        ratio = min(a / b for a, b in zip(size, pil_image.size))
+        size = tuple(int(ratio * x) for x in pil_image.size)
```
ColonelPhantom, line 57: I'm not sure, but I think I found why some thumbnails are really low quality, and this has the same problem.

The problem is as follows: some users have very non-square profile pictures. (e.g. 16:9 and I even found one that was more than 2:1 portrait). The thumbnailing code resizes it then to be e.g. 300px along the largest axis. This results in images of e.g. 300*200px or 145*300px.

When the time to display the thumbnail comes, the website uses square card views. This means that images get cropped by the browser. But because the thumbnail gets cropped, we get effective resolutions of 200*200px or even 145*145px.

We should probably fix this by cropping the thumbnails beforehand, either as a separate type of thumbnail (because I can imagine we also have use-cases for non-cropped thumbnails, although I'm not sure), or making the cropping along the shortest axis (which might not be a great idea, as images can then be larger than expected).

ColonelPhantom (commented):
I don't really know enough about the photo pipeline to say much and I haven't tested the PR locally, so I can't say too much, but I don't think this solves #2936. See comment for more detail.

website/utils/media/processors.py
DeD1rk: That's the case for sizes without cover=True. But that has actually also been the case before djangp-thumbnails (I coppied this size calculation from our old custom thumbnailer). I did notice, in the django-thumbnails settings from before this, that the sizes had some weird stretching settings.

So with this PR, the 'small' thumbnails are cropped to cover 300x300 (only less if the original smallest dimension was <300px). Only the `photo_medium`, `photo_large` and `source` sizes are resized to _fit_ in some size (and hence potentially much smaller)

DeD1rk: This was indeed the cause for [#2936](https://github.com/svthalia/concrexit/issues/2936). One of the examples in the screenshot there is indeed 199x300, instead of 300x300. But with this PR small does get cropped.

ColonelPhantom (approved):
Seems good! One request though, blurry thumbnails do not get fixed unless the image is uploaded. Perhaps we should ensure thumbnails are regenerated or whatever? Although I guess we can just do that by hand again using whatever the management command was.

DeD1rk:
Yep we need to regenerate all thumbnails after this. I think using django-thumbnails' management command is fine. With celery and redis, we can also swap django-thumbnails' backend to redis, so that would probably also cause them to all be regenerated (but clearing the S3 thumbnails folder would also be nice)

DeD1rk:
I'll probably merge this in a few weeks, after another release to improve the moneybird and registrations situations.

DeD1rk:
We can re-upload every existing photo with this: 
```python
from django.apps import apps
from django.core.files.uploadedfile import InMemoryUploadedFile


def reupload():
    models = {
        "facedetection.ReferenceFace": ["file"],
        "members.Profile": ["photo"],
        "activemembers.MemberGroup": ["photo"],
        "photos.Photo": ["file"],
        "merchandise.MerchandiseItem": ["image"],
        "thabloid.Thabloid": ["cover"],
        "partners.Partner": ["logo", "alternate_logo", "site_header"],
        "partners.PartnerImage": ["image"],
        "partners.Vacancy": ["company_logo"],
    }
    for model_path, fields in models.items():
        app_label, model_name = model_path.rsplit(".", 1)
        model = apps.get_model(app_label, model_name)
        print(f"\n\nProcessing {model_path}\n")
        for instance in model.objects.all():
            has_changed = False
            try:
                for field in fields:
                    field_file = getattr(instance, field)
                    if field_file:
                        has_changed = True
                        content = field_file.open()
                        setattr(
                            instance,
                            field,
                            InMemoryUploadedFile(
                                content,
                                field,
                                field_file.name,
                                "image/jpeg",
                                content.tell(),
                                None,
                            ),
                        )
                if has_changed:
                    instance.save()
                    print(".", end="")
            except Exception as e:
                print(f"Error processing {instance}: {e}")
                print("investigate: {model_path}, {instance.pk}")
```

That will take quite some time of course.

The S3 storage we use currently prevents overwriting existing files, so all of these files either get a randomised suffix or a completely different name. 

Particularly, reuploading a Photo.file causes the upload_to function to give a new number to it, which is really quite ugly. I'd like to replace that, and possibly most other file uploads, to just use a uuid filename. Then we don't have to worry at all about making filenames unique, safe and not ugly (I hate the random suffix names, they look so messy). I'll probably make a separate PR changing the upload_to's so that renaming happens at the same time.

PR #3350, opened 2023-09-13
Deleted hidden photos

### LINKED ISSUE #3311: Drop Photo.hidden
### What
I suggest we drop the `Photo.hidden` field, in favour of simply removing photos we don't want to be seen.

### Why
If someone doesn't want their photo seen, we have no business storing that photo in any way at all.
Hiding a photo is not sufficient, and hence a hiding feature is worthless.  

### How
Drop the field, and remove any references to it (in filters, etc.). In the migration, delete any photo that is hidden.

There are currently +- 5 hidden photos.

Reaxit currently expects a `hidden` field in the API (although it's never used). That has to be removed _before releasing this_, as part of https://github.com/svthalia/Reaxit/issues/437.

Noah-marc:
Closes #3311 

### Summary

Deleted the option to have hidden photos, along with existing hidden photos.


DeD1rk (approved):
Nice (and tested locally)!

Issue #3298, opened 2023-08-21
Renewal discount triggers prevent_saving on payment creation


DeD1rk:
### Describe the bug
https://thalia.sentry.io/issues/4409801880/?project=1463433&referrer=issue-stream&statsPeriod=14d&stream_index=6

When people create a renewal to study-time membership, and a payment is made for it, the saving prevention is triggered because the `contribution` amount changed. The discount calculation should probably change to be calculated earlier so the amount is accurate from the start.

### How to reproduce
Steps to reproduce the behaviour:
1. Have an expiring membership
2. Renew to study-long membership
3. Pay

DeD1rk:
Duplicate of #3297, github was broken

PR #3355, opened 2023-09-15
Optimize latest memberships everywhere

DeD1rk:
### Summary
This is a continuation of #3229. I'm trying to get rid of membership-related N+1 queries everywhere (they are in lots of places) with a few tricks:

- Making `latest_membership` cached. 
  _This is a logical difference, that may be problematic for the registrations app. There are already bugs (#3298, ???) caused by incorrectly working with memberships. #3298 might actually get worked around by this caching, but a core problem in the registrations app is that it uses properties as if they're normal fields (not expected to change). This is a fine assumption everywhere, except in registrations, where memberships should be expected to change within a request-response cycle. Registrations should IMO never use the `latest/current_membership` properties, and only use normal explicit model methods (called with `()` to be explicitly recalculated) or only access the Membership model directly without any convenience methods._
- Prefetching memberships for more (API) views: event registrations, ...?

I looked through registrations, and think that after some small fixes this shouldn't break anything. Refactoring/fixing registrations further is another thing for another (bugfix) PR.

This improves at least:
- `/api/v2/events/`
- `/api/v2/events/:pk/registrations`
- `/api/calendarjs/birthdays`

And likely several other (non-api) views.

### How to test
1. Open the changed views with and without these changes.
2. Check that the results are exactly the same.
3. Check that the queries are much better.

### DIFF
--- a/website/events/api/v2/views.py
+++ b/website/events/api/v2/views.py
@@ -56,6 +56,7 @@ class EventListView(ListAPIView):
                 "organisers__board",
                 "organisers__committee",
                 "organisers__society",
+                "organisers__contact_mailinglist",
             )
         )
         if self.request.member:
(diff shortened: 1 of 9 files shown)


ColonelPhantom (approved):
I don't know enough about registrations to vouch for it still working there, but the changes seem good otherwise.

PR #3374, opened 2023-09-24
Add VSCode settings

### LINKED ISSUE #3371: Provide VSCode settings
### What?

It would be nice to check in some VSCode configuration to help people have a pleasant experience when working on concrexit, without having to be told to set up black, isort, etc on save.

### Why?

This will make it easier for people to start with concrexit, and help them commit more easily, as proper autoformatting and linting will prevent them having to 'Fix lints' every time they want to commit. 

Also, I think nearly everyone is using VSCode these days.

### How?

Add a .vscode dir with settings, such as:
- Black on save
- isort on save
- recommend ruff extension
- recommend pylance
- Set html indentation
- ...

This file should be committed, but then added to .gitignore to allow people to override it without git picking that up as changes.

DeD1rk:
Closes #3371.

This checks in:

- Recommended extensions for linting and formatting
- Settings for those extensions
- A launch config for manage.py runserver and test with debugger 

### How to test
Try working on concrexit on a laptop with only VSCode, and no python-specific configuration. The see that, if you follow the recommendations, you get format on save and the rigt linting out of the box.


ColonelPhantom:
Would it also be possible for this to deal with Poetry in some way?

DeD1rk:
I'm not sure. The poetry install location depends on user settings and the python version, so we can't hardcode the python interpreter path afaik.

ColonelPhantom:
It's possible to find the venv path with `poetry env info -p`, but I don't know if VSCode allows executing anything to find it. Also I'm guessing one still needs to install poetry (obviously), and then also do `poetry install` et al?

Maybe a bit cursed but one potential workaround could maybe be a 'setup' shellscript or similar that ensures poetry is setup, creates/updates the venv, and sets the correct path in the VSCode config file? Although that's very deep into magic territory.

DeD1rk:
Yep that all still needs to be done manually. We can automate that in codespaces though 😏

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
...
+            "name": "Django: manage.py runserver",
+            "type": "python",
+            "request": "launch",
+            "program": "${workspaceFolder}/website/manage.py",
+            "args": [
+                "runserver",
+            ],
+            "justMyCode": true,
+            "django": true,
+        },
```
ColonelPhantom, line 14: Maybe add a test runner too? That way people can run tests in one click, and it also works with the graphical debugger.

DeD1rk: As of now, integrating with the VSCode testing tab (so people can e.g. run single tests from the UI) is not possible. But some work is going on: https://github.com/microsoft/vscode-python/pull/21468 is under consideration. 

I can add a launch config to simply run the tests with a debugger, but 'breakpoint on uncaught exception' will not open automatically open the debugger when a test fails/errors. But it's useful if you manually set breakpoints of course. Sadly a launch config also doesn't really make it easy to filter the tests you want to run.

T8902 (approved):

PR #3383, opened 2023-09-27
Committee chair permission

### LINKED ISSUE #3303: Committee chair permissions
Some committees need certain permissions on the website, but they aren't really necessary for every member of that committee. For example, with #3105, the Thabloidcie should have access to the admin page introduced there. But it would be sufficient, and hence best, if only the committee chair, or perhaps another small subset of the committee has that permission.

I propose extending the `activemembers.backends.MemberGroupBackend` (and `MemberGroupMembership`) to optionally support increased permissions for certain members of a committee. That is:

- We add `MemberGroup.chair_permissions`.
- We add `MemberGroupMembership.has_chair_permissions` (or not and reuse the `chair` field, if we think only allowing the single chair per committee to have increased permissions)
- People with `has_chair_permissions` in their membership can use `Union(permissions, chair_permissions)`

### Motivation
This way we can drop permissions for many people, and increase permissions for a few who need them.

Leave it as is, give every committee member the same permissions, and rely on the board if someone in a committee needs something done that we find to risky to give everyone permission for.

Noah-marc:
Closes #3303.

### Summary
Added the option to add permissions that only a chair should have to MemberGroups and for  MemberGroupMemberships to have those chair permissions. Also for all existing MemberGroupMemberships that are already marked as chair also automatically have chair permissions 

### How to test
1. Go to a Committee or Society in the admin, select the desired chair-permissions in the respective field and save them
2. Go to the MemberGroupMembership admin and click on a MemberGroupMembership that you want to give chair permissions, mark the checkmark has_chair_permissions and save
3. Make sure that the member connected to that MemberGroupMembership has no other permissions form other MemberGroups
4. Log into that user, go to the admin page and try do the things that you gave permissions to do. It should work then
5. Log into a user that is part of that same MemberGroup and where his Membership does not have those permissions
6. Try to do the same actions as in 4. This time it should fail


DeD1rk (commented):
Looks good as far as I can tell. Is there any chance you could test the migration with a local postgres db (as runpython + field changes has been giving issues on postgres in the past that didn't occur on sqlite3)? 
And maybe it would be nice if someone else also looks at this as I'm biased on the actual permissions backend changes.

DeD1rk:
Oh and yes I see this includes an unrelated commit  (86978fcd087429a3001142cabfcc1576822c2d64) that's causing a conflict. So could you rebase it out? (`git rebase -i origin/master` and `drop` the photo.hidden commit)

website/photos/admin.py
```diff
@@ -101,11 +101,10 @@ class PhotoAdmin(admin.ModelAdmin):
     list_display = (
         "__str__",
         "album",
-        "hidden",
         "num_likes",
     )
     search_fields = ("file",)
-    list_filter = ("album", "hidden")
```
T8902, line 108: I think removing the hidden picture feature would be better in a different pull request.

T8902 (changes requested):
Removing hidden picture feature should be in a different branch/pull request in my opinion

DeD1rk:
I did the rebase

DeD1rk (approved):

PR #3384, opened 2023-09-27
events/admin: disable delete if not organiser

ColonelPhantom:
### Summary
Active members are able to delete events that they do not organize. This PR fixes that.

### How to test
1. Have an active member user that has permissions to edit/view/change/add events
2. Create an event that they are NOT an organiser off
3. As that user, go to the events admin
4. Clicking already doesn't work, but Edit does work
5. On Edit, you can also delete the event without this PR applied, while applying it removes the big red Delete button.

### DIFF
--- a/website/events/admin/event.py
+++ b/website/events/admin/event.py
@@ -129,6 +129,12 @@ class EventAdmin(DoNextModelAdmin):
             title=obj.title,
         )
 
+    def has_delete_permission(self, request, obj=None):
+        """Only allow deleting an event if the user is an organiser."""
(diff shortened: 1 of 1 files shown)


ColonelPhantom:
I'm looking at this a bit more, and is it intended that the detail page is also not visible to non-organizers? Because non-organizers being able to view event information does seem intended. But the detail page also has information on e.g. registrations, which might be considered private?

If the 'general' info is intended to be viewable, maybe we should make the link to the detail page go to the view page for users with insufficient permissions?

Scarletto (approved):
Looks good

PR #3390, opened 2023-09-28
Check that deletion is forbidden in event admin tests

ColonelPhantom:
Closes #ISSUE

### Summary
Adds a test for #3384 to ensure correctness

### How to test
1. Run `make test` and see that it succeeds
2. Revert PR #3384 
3. Run `make test` again and see that it now fails because it fails to give a 403.


DeD1rk (approved):
<3 <3 <3 !!!!

PR #3394, opened 2023-09-29
Use celery

### LINKED ISSUE #3361: Use Celery for background processing
### What?

Celery (and celery beat) are the standard solution for doing background processing. We should use that instead of cronjobs with management commands, and expensive work in the request-response cycles.

Most cases where we send emails from django views should be done in the background for better performance. And all periodic management commands should be changed to use celery beat.

### Why?
Performance and flexibility. 

With celery set up, we can also look into changing stuff with album uploads and e.g. face detection.

### How?

See use-celery branch. Me and @T8902 are already working on it.

T8902:
Closes #3361 

### Summary
Cronjobs and registrations conformation moved to celery workers and celery workers implemented. 

### How to test
Start redis-server and wait about 2 minutes for a cronjob to execute


website/members/tasks.py
```diff
@@ -0,0 +1,13 @@
+from members import emails
+
+
+def membership_announcement():
```
DeD1rk, line 4: These aren't tasks

website/facedetection/tasks.py
```diff
@@ -0,0 +1,23 @@
...
+from facedetection.services import (
+    resubmit_photos,
+    resubmit_reference_faces,
+    submit_new_photos,
+)
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task()
```
DeD1rk, line 14:
```suggestion
@shared_task
```

website/promotion/tasks.py
```diff
@@ -0,0 +1,8 @@
+from celery import shared_task
+
+from promotion import emails
+
+
+@shared_task()
```
DeD1rk, line 6:
```suggestion
@shared_task
```

website/pushnotifications/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+from django.utils import timezone
+
+from celery import shared_task
+
+from pushnotifications.models import ScheduledMessage
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task()
```
DeD1rk, line 13:
```suggestion
@shared_task
```

website/registrations/tasks.py
```diff
@@ -0,0 +1,5 @@
+from registrations import services
+
+
```
DeD1rk, line 3: @shared_task

website/pushnotifications/tasks.py
```diff
@@ -0,0 +1,28 @@
...
+
+from pushnotifications.models import ScheduledMessage
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task
+def send_scheduled_messages():
+    """Send a scheduled push notifications."""
+    interval = int(120)
```
DeD1rk, line 16: int(120) is a bit strange. This can just be a parameter as it was before, and set in the settings (`"args": (120,)`).

PR #3407, opened 2023-10-01
Set registration queue_position=None when cancelled

### LINKED ISSUE #3315: Canceled event registrations are marked as queued
### Describe the bug
After canceling a registration, 

### How to reproduce
1. Go to https://staging.thalia.nu/api/v2/admin/events/183/registrations/?limit=999999999&ordering=-date_cancelled&cancelled=true&queued=true
2. Go to https://staging.thalia.nu/api/v2/admin/events/183/registrations/?limit=999999999&ordering=-date_cancelled&cancelled=true&queued=false
3. Notice that the canceled registrations are only listed under `queued=true`, and that the queue position is 3
4. Go to https://staging.thalia.nu/api/v2/admin/events/183/registrations/?limit=999999999&ordering=-date_cancelled&cancelled=false&queued=true
5. Notice that the queued position 3 already exists.

### Expected behaviour
Canceled registrations should show up as canceled and not in queue, as they are not in the queue

### Additional context
I think the reason they are listed as queued is because the queue position is set. Ideally this field would be nulled or set to a special value like -1.

ColonelPhantom:
Closes #3315 

### Summary
Sets `queue_position=None` when a registration is cancelled.

### How to test
1. Create an event with queued registrations
2. Cancel a queued registration
3. See that before this PR it gets a duplicate queue position on `/api/v2/admin/events/1234/registrations/?ordering=-date_cancelled`, and with this PR applied it gets.

Alternatively:
1. Run tests
2. Revert 1d7a0e1
3. Run tests again and see them failing

### DIFF
--- a/website/events/models/event_registration.py
+++ b/website/events/models/event_registration.py
@@ -1,7 +1,7 @@
 from django.core import validators
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models import Count, F, Q
+from django.db.models import Case, Count, F, Q, When
 from django.db.models.functions import Greatest, NullIf
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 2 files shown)


DeD1rk (approved):

JAicewizard:
no idea if its actually good, but LGTM

JAicewizard (approved):

PR #3423, opened 2023-10-04
Add events.view_unpublished permission

### LINKED ISSUE #3379: I can see unpublished events
### Describe the bug
Unpublished events are visible to me, while I don't think I should be able to see them.

### How to reproduce
1. Go to 'Events', either on the website or in the admin
2. See unpublished events
3. Click on it -> get a 403

### Expected behaviour
Do not see unpublished events unless they are organized by a committee I am in

### Additional context
Should be hidden in the list, keeping them visible in the admin list might be a good idea (especially for committee chairs #3303) for planning reasons.

TRMMax:
Closes #3379 

### Summary
Added a new permission (Can see any unpublished event) that shows any unpublished event in the agenda & events admin tab. In turn, unpublished events which are not organized by the user are not shown to them now by default.

### How to test
1. Go to Calendar or Events admin
2. Verify that as a superuser you can see any unpublished event, and as non-superuser you can only see your own unpublished events.
3. Verify that having the override_organiser or view_unpublished permission allows you to see any unpublished events.


DeD1rk (commented):
Very nice. I have one more request: initially, we don't want the behaviour to change. Taking away the current permissions for certain people will be a policy decision by the board. 
So it would be great if you can add a custom migration (see RunPython) that gives the permission to all groups that currently have view_event.

There should be a pretty close example in #3383 that you can more or less copy.

PR #3446, opened 2023-10-18
Overhaul moneybird

### LINKED ISSUE #3295: Moneybird synchronization error: InvalidData("API error 422: {'sepa_mandate_date': ['cannot be today or in the future']}")
https://thalia.sentry.io/issues/4111947935/events/ce4b0f82c04244b58c9864ce6bceb729/?project=1463433

### LINKED ISSUE #3391: Make moneybirdsynchronization more robust
Moneybirdsynchronization is error-prone and untested. We need to improve that. 
At the very least, it seems to me that, when duplicate Thalia Pay payments were made (and possibly also in other cases), this logic ended up making concurrent different changes to the same `index_nr`, such that multiple payments ended up linked to the same mutation_id:
https://github.com/svthalia/concrexit/blob/c9a5b87c53a771a360905fb34c8204a01aeb5f14/website/moneybirdsynchronization/services.py#L181-L194

We should look into how we can make it more robust. Ideally, we should be able to run moneybirdsynchronization at any time in a batch, so that we can always rerun it correctly if something fails.

With #3361 we can also make sure that synchronization happens quickly, without blocking the request cycle as it currently does.

### LINKED ISSUE #3426: Fix registrations/renewals not being pushed to moneybird.
The signal that creates an 'external sales invoice' on moneybird for registrations and renewals listens for post_save on Registration and Renewal models. However, withe the current registrations code, only Entry.save is called after paying and completing a registration or renewal. So the invoices weren't actually getting created. 

Note that, as registrations and renewals are deleted 31 days after being completed, it's important that we create invoices on moneybird for them before then.

This will probably be solved by #3420.

DeD1rk:
Closes #3391.
Closes #3426.
Closes #3295.

### Summary
I made moneybirdsynchronization happen mostly nightly. Some things other than creating invoices still need to happen, but are less crucial I think. 

Pushing all payments of a day (including older ones if for any reason they hadn't been pushed yet) should solve the issue of changing financial payments being so weird. I'll need to check whether there is any guarantee on the order in which mutations are listed immediately after creating the batch. We could lookup based on payment pk (that needs to match the 'message') to be sure. As I made it, each day a statement is made per payment type, labelled with the sync time (not per se grouped by payment date). That's a bit different from what it was previously.

And I introduced a `MONEYBIRD_START_DATE` setting, to prevent pushing old things from before april or before october.

I still need to:

- [x] Change migration to set needs_synchronization False for old invoices.
- [x] Make invoices and payments be deleted in the background. Let's set a `needs_deletion` flag from post_delete signals. 
- [x] Show `needs_synchronization` and `needs_deletion` in the admin.
- [x] Write some tests. I'm quite certain that even though this PR may solve a bunch of bugs, it also introduces a few. Luckily all of this is quite easy to test.

@JobDoesburg @nvoers @movieminer could you take a look at the changes so far and perhaps throw in your opinions about the stuff that still needs to happen?


website/moneybirdsynchronization/services.py
```diff
@@ -40,6 +57,54 @@ def create_or_update_contact(member):
     return moneybird_contact
 
 
+def synchronize_moneybird():
+    """Perform all synchronization to moneybird."""
+    logger.info("Starting moneybird synchronization.")
+
+    # First make sure all moneybird contacts' SEPA mandates are up to date.
+    sync_contacts_with_outdated_mandates()
```
JobDoesburg, line 65: I would refactor this to simply sync_contacts (but can be done in the future as well)

website/moneybirdsynchronization/services.py
```diff
@@ -151,90 +216,183 @@ def delete_external_invoice(obj):
...
-            )
-            _add_payment_to_statement(moneybird_payment, existing_statement_id)
-        else:
-            _create_statement_and_add_payment(moneybird_payment, financial_account_id)
+        logger.info(
+            "Pushing %d %s payments to Moneybird.", payments.count(), payment_type
+        )
+
+        financial_account_id = financial_account_id_for_payment_type(payment_type)
+        reference = f"{payment_type} payments at {timezone.now():'%Y-%m-%d %H:%M'}"
```
JobDoesburg, line 372: %H and %M can be left out

JobDoesburg (commented):
Looks good! I will add some minor other changes to this branch in a minute to

JobDoesburg:
@DeD1rk this also needs the updated moneybird api client that waits and retries on a timeout

website/moneybirdsynchronization/services.py
DeD1rk: Yup I though about that as well, will probably do that at some point but for now it doesn't really matter

website/moneybirdsynchronization/services.py
DeD1rk: Yeah, the thing is it might be confusing that in the 2023-10-15 statement, there would mainly be payments from 2023-10-16

JobDoesburg: Hmm what about really just selecting the previous day? so 

`statement_day = timezone.now()-timedelta(days=1)`

and filter on

`created_at__date=statement_day`

Sounds more robust

DeD1rk: That's also nice but doesn't automatically fix stuff if the sync doesn't run on some day, or if it runs twice somehow (it'd make the same reference twice)

website/moneybirdsynchronization/signals.py
```diff
@@ -126,66 +126,10 @@ def post_bank_account_delete(sender, instance, **kwargs):
...
-def post_order_save(sender, instance, **kwargs):
-    if instance.payment is None:
-        return  # We only create invoices for orders that are paid, because we can have quite some phantom orders
-
-    try:
-        services.create_or_update_external_invoice(instance)
-    except Administration.Error as e:
-        send_sync_error(e, instance)
-        logging.exception("Moneybird synchronization error: %s", e)
+# TODO: delete eventregistration invoice when it becomes free, or not invited.ss
```
movieminer, line 129: Would it be necessary to change this to a scheduled task? I don't think a lot can go wrong here, since the invoice already exists with an id? Same for updating

DeD1rk: Not much can go wrong indeed, but we shouldn't do anything heavy (api calls) in post_save signals for performance reasons mainly.

website/moneybirdsynchronization/services.py
```diff
@@ -151,97 +163,309 @@ def delete_external_invoice(obj):
...
-            }
-        }
+    if mandates_to_push.exists():
+        logger.info(
+            "Pushing %d contacts with outdated mandates to Moneybird.",
+            mandates_to_push.count(),
+        )
+
+    for mandates in mandates_to_push:
+        create_or_update_contact(mandates)
```
nvoers, line 261:
```suggestion
        create_or_update_contact(mandates.member)
```

website/moneybirdsynchronization/services.py
```diff
@@ -18,18 +35,10 @@ def create_or_update_contact(member):
...
-        else:
-            # Link the existing moneybird contact. Then update the data.
-            moneybird_contact.moneybird_id = response["id"]
-            response = moneybird.update_contact(
-                moneybird_contact.moneybird_id, moneybird_contact.to_moneybird()
-            )
+        # Push the contact to Moneybird.
+        response = moneybird.create_contact(moneybird_contact.to_moneybird())
+        moneybird_contact.moneybird_id = response["id"]
+        moneybird_contact.moneybird_sepa_mandate_id = response["sepa_mandate_id"]
```
nvoers, line 41: If there is no SEPA mandate, the response is empty and Django processes this as an empty string. However there is are null=true and unique=true constraints on the model field.
This means that Django checks if "" == "" which is true, thus the constraint fails ([source](https://stackoverflow.com/questions/17257031/django-unique-null-and-blank-charfield-giving-already-exists-error-on-admin-p)) This can be solved by doing your own uniqueness checks.

DeD1rk: I made it cast to None

JobDoesburg (approved):
Looks great!

PR #3452, opened 2023-10-25
Order MemberGroup by name by default

ColonelPhantom:
Closes #ISSUE.

### Summary
@Scarletto wanted this for the effect on the events API. Ordering like this by default seems reasonable because i18n seems the only reason it wasn't done in the past and that's not a concern anymore.

### How to test

1. Go to an event in the API with many organisers
2. Observe that organisers are sorted by name instead of PK or something else.
3. See that this happens consistently.


DeD1rk (approved):
The migration got outdated, but looks good!

PR #3456, opened 2023-10-26
Set period on registration/renewal invoices

nvoers:
Sets a period and tax rate for external sales invoices created for registrations and renewals. This improves the bookkeeping as registrations are accounted for the current year, whereas renewals are accounted for the following year.


JobDoesburg:
There's no need to transition to actual invoices, periods can also be selected for external invoices 🙂

JobDoesburg (approved):

website/moneybirdsynchronization/models.py
```diff
@@ -269,6 +275,12 @@ def to_moneybird(self):
 
         ledger_id = ledger_id_for_payable_model(self.payable_object)
 
+        period = None
+        tax_rate_id = None
+        if isinstance(self.payable_object, (Registration, Renewal)):
+            period = membership_to_mb_period(self.payable_object.membership)
```
DeD1rk, line 281: This will crash on renewals/registrations that have been paid but not yet completed. That in turn only happens when there's a bug in registrations, but there are currently such cases.

DeD1rk: I'll have to sit with @JeeVee11 to clean those cases up some day soon but to prevent this from crashing the sync maybe we can just check for it. The assumption that the time is based on created_at was also just fine (as registrations app makes the same assumption)

JobDoesburg: I will try-catch these

Issue #3453, opened 2023-10-25
ThaliaPleasePay


mark-boute:
### What?
Payment Request, or ThaliaPleasePay (`TTP'tje`)

### Why?
Invoices are annoying, they require administrative work on both the sending and receiving end. This could be made much easier by using Thalia(Please)Pay an admin would get a nice overview of which request has been payed through a Payment for a Payable (could be any type of payment). And a payer would be able to use Thalia Pay instead of manually transferring the payment.

### How?
A new concrexit app, `paymentrequests` is needed for this, it will depend on `payments`

In admin there will be an entry per `TTP'tje`, which contains a few display fields matching the payment, namely `amount`, `topic`, `notes`. And an inline row per `Member` that has been requested to pay, this is a `Payment Request` and is a `Payable` object. It also has a status, it can be `Pending`, `Paid`, `Cancelled` (essentially a removed payment request, and therefor a separate flag on the class).

The create form will have text fields for `amount`, `topic` and `notes`, and an inline where you can search for and select members to whom the payment request apply (this may also be a single Member). Upon saving this request will just be saved as a concept (_iff_ at least one Member is selected). It will now also be displayed in the Admin page with a `[Concept]` label. Upon sending this will change to `[Pending]`, and once there are no more `Pending` payments linked to the `TTP` the status will change to `[Done]`.

Upon sending all Members linked to the `TTP` through a Payable will be notified of the request, this could be through a Push Notification and/or an e-mail.

DeD1rk:
This diagram is ugly but might be informative (sadly github doesn't suport this diagram type i guess):
[...]

PR #3464, opened 2023-10-30
Merchandise sales moneybird integration

### LINKED ISSUE #1657: Merchandise products in sales productlists
With the new `sales` app, it would be nice to have the possibility (for the board) to also do merchandise sales via this system.

Make a `sales.ProductList` of `merchandise.Item`s 

### Motivation
Merchandise products are also products.

### Additional context
- Perhaps, we want to make `MerchandiseItem`s actually extend `Product`s (so the `sales` app doesn't need to change too much). 
- We should think about how to go with different sizes. For example, maybe we want to keep what we currently have as `MerchandiseProduct`s and then have multiple `MerchandiseVariant`s for different sizes, that are the actual products. This way, we can display a single merchandise product but actually have multiple versions for it.
- We should think about how this would work with product lists and prices.

tmgdejong:
Closes #1657.

Adds a way to sell merchandise via the site, such that the merchandise sale gets registered automatically and will be processed in Moneybird.

It works using Thadmin, where a board member can enter a `Merchandise Sale` shift to sell merchandise. this shift has as product list a unique list, containing all merchandise items. This list is generated and updated automatically.
`Merchandise Sale` shifts are also created automatically, on a daily basis, and on creating a new shift, all old `Merchandise Sale` shifts get locked, or deleted if there are 0 items sold.
Just like a normal shift sale, a payment and an external invoice get created, but next to this, a general journal document gets created as well. This allows us in Moneybird to add a general journal document, which creates a credit transaction on the `merchandise stock` ledger, linked with a debit transaction on the `merchandise costs` ledger.

~~2 'problems' that are still in the code:~~
~~- In the migration for a merchandise item, even though Python code is ran to retain our current merchandise items, this does not seem to work yet.~~
~~- When a general journal document is 'updated', we now remove it and simply create a new one. This should be implemented the same way as with external invoices, where an attribute id is saved.~~

@JobDoesburg @DeD1rk @nvoers let me know what you think!


DeD1rk:
I don't really see an advantage in using sales here, as opposed to making e.g. a MerchandiseSale payable model. If we had a way to make miscellaneous payment requests, would any of this still be necessary?

movieminer:
> I don't really see an advantage in using sales here, as opposed to making e.g. a MerchandiseSale payable model. If we had a way to make miscellaneous payment requests (#3453) would any of this still be necessary?

The first implementation was with a MerchandiseSale payable model, see c708f322806190f126cf559cb3c7114ea36de2d8 and before.  However, as @JobDoesburg mentioned, this way the board can go to thadmin.thalia.nu and use the continuous merchandise sale shift to add a sale, and the rest (payments, external invoices) will be handled by the already existing code.

Miscellaneous payment requests will not solve the problem of needing a general journal document for a merchandise sale to change the merchandise stock in moneybird.

website/sales/signals.py
```diff
@@ -0,0 +1,21 @@
...
+from django.dispatch import receiver
+
+from merchandise.models import MerchandiseItem
+
+from .models.product import ProductList
+
+
+@receiver(post_save, sender=MerchandiseItem)
+def update_product_list(sender, instance, **kwargs):
+    product_list = ProductList.objects.get_or_create(name="Merchandise Product List")[0]
```
JobDoesburg, line 11:
```suggestion
    product_list = ProductList.objects.get_or_create(name="Merchandise")[0]
```

website/sales/signals.py
```diff
@@ -0,0 +1,21 @@
...
+    product_list_products = product_list.products.all()
+    merchandise_items = MerchandiseItem.objects.all()
+
+    for merchandise_item in merchandise_items:
+        if merchandise_item not in product_list_products:
+            product_list.product_items.create(
+                product=merchandise_item, price=merchandise_item.price
+            )
+
+    product_list.save()
```
JobDoesburg, line 21: This line is not required I think

website/sales/services.py
```diff
@@ -33,3 +38,35 @@ def execute_data_minimisation(dry_run=False):
     if not dry_run:
         queryset.update(payer=None)
     return queryset.all()
+
+
+def create_daily_merchandise_sale_shift():
+    today = timezone.now().date()
+    merchandise_product_list = ProductList.objects.get_or_create(
+        name="Merchandise Product List"
```
JobDoesburg, line 46:
```suggestion
        name="Merchandise"
```

website/sales/admin/product_admin.py
```diff
@@ -19,3 +24,7 @@ class ProductListAdmin(admin.ModelAdmin):
     inlines = [
         ProductListItemInline,
     ]
+
+    def get_queryset(self, request: HttpRequest) -> QuerySet[Any]:
+        queryset = super().get_queryset(request)
+        return queryset.filter(~Q(name="Merchandise Product List"))
```
JobDoesburg, line 30: IMO we don't need to exclude the list here

website/moneybirdsynchronization/services.py
```diff
@@ -322,6 +418,36 @@ def _sync_sales_orders():
     _try_create_or_update_external_invoices(sales_orders)
 
 
+def _sync_merchandise_sales():
+    """Create journals for sales orders that are merchandise sales."""
+    merchandise_sales = Order.objects.filter(
```
JobDoesburg, line 423: I would say don't filter on the shift title, but on the product being a merchandise product

website/moneybirdsynchronization/models.py
```diff
@@ -389,3 +397,80 @@ def to_moneybird(self):
...
+
+    needs_deletion = models.BooleanField(
+        default=False,
+        help_text="Indicates that the journal has to be deleted from moneybird.",
+    )
+
+    def __str__(self):
+        return f"Moneybird journal for {self.order}"
+
+    def to_moneybird(self):
```
JobDoesburg, line 438: This is very specific to merchandise item bookings, while the model itself is a "general" model for all sorts of General Journal Documents. Right now we only use them for merchandise, but it might be nice to either rename it to explicitly be called MerchandiseJournalDocuments, or subclass it or give it some field that specifies it is about merchandise. So that later, in case we might want to sync different kinds of general journal documents as well, we don't have problems

JobDoesburg: Im in favor of subclassing it btw

website/merchandise/migrations/0012_remove_merchandiseitem_id_and_more.py
```diff
@@ -0,0 +1,63 @@
...
+        ("sales", "0007_alter_productlistitem_options_and_more"),
+        ("merchandise", "0011_alter_merchandiseitem_image"),
+    ]
+
+    def store_and_delete_merchandiseitems(apps, schema_editor):
+        MerchandiseItem = apps.get_model("merchandise", "MerchandiseItem")
+        itemlist = list(MerchandiseItem.objects.all())
+        MerchandiseItem.objects.all().delete()
+
+    # This does not seem to work yet
```
JobDoesburg, line 21: this shouldn't be a method on the Migration, but a separate function

JobDoesburg (changes requested):

website/moneybirdsynchronization/models.py
```diff
@@ -389,3 +397,106 @@ def to_moneybird(self):
...
+        null=True,
+    )
+
+    moneybird_details_credit_attribute_id = models.CharField(
+        verbose_name=_("moneybird details attribute id (credit)"),
+        max_length=255,
+        blank=True,
+        null=True,
+    )
+
```
JobDoesburg, line 433: Let's add .to_moneybird() here and raise an NotImplementedError (or return [])

JobDoesburg (changes requested):

website/moneybirdsynchronization/services.py
JobDoesburg: Doest filtering on 

```
        items__in=MerchandiseItem.objects.all(),

```

work

movieminer: Since `items

JobDoesburg:
Looks good! Some more remarks that might be nice to account for, but are not important:


- It is possible to change the price of the product list items to a different value than the price of the merchandise item. You can't really prevent that (except with setter and getter behavior specifically overriding this). But you might want to disable editing in the admin, simply by overriding the has_edit_permission and has_add_permission on the inline
- Might be nice to add a help text specifying which values are incl and excl VAT

JobDoesburg:
Also, I just realized, we might actually not want to make merchandise items actual sales products (extending the model like is implemented in this PR now). But only add a foreign key between them. Namely, we probably want to keep track of what sizes are sold, while we don't want to list every individual size on the website. So there should be a "t-shirt" in the merchandise app, and "T-shirt size L 22-23" in the sales app (with a foreign key to the merchandise app). That also simplifies the migrations

DeD1rk:
I really don't like making sales depend on merchandise. While I haven't really looked into it yet, it does look like the added stuff from services can simply move into merchandise/services. 

Regarding different sizes, a FK from product to merchandise item would be a nasty dependency, but how it is now seems fine (although it could just be a FK _to_ product instead of inheritance) but if we want separate sizes IMO that should not use a FK _from_ product (either a through model to emulate that without changing product, or e.g. make separate linked merchandise items and don't display some of them)

JobDoesburg:
@movieminer @DeD1rk I have pushed one commit that changes the data model a little bit, to fix the things mentioned above. This does need some finishing touches, but I currently don't have the time. Feel free to work on it yourself, otherwise I might do it in the upcoming days

movieminer:
I don't think we need to integrate different sizes in the site, since all the different sizes still cost the same, and we don't want to keep track of the sales in concrexit (since we also don't keep track of stock). The board still needs to note the sale in their cash book, and adding the sale to the site is more of a tool to register the sale in moneybird.

DeD1rk:
I just moved the changes from sales to merchandise so the dependence is clean. I'll probably take a look at the rest in a few days

JobDoesburg:
> we don't want to keep track of the sales in concrexit (since we also don't keep track of stock). The board still needs to note the sale in their cash book

That's not the idea right? The website becomes the cash book as far as I understand it

JobDoesburg:
@movieminer could you test if it still works for you? I think it should be ready now

JobDoesburg (approved):

JobDoesburg:
I just realized we actually had to sync things even differently

ColonelPhantom:
This seems pretty stale. Do we still want to do anything with this?

T8902:
@tmgdejong Im going to close this due to no activity

PR #3467, opened 2023-11-02
Promo Request status updates for organisers

### LINKED ISSUE #3005: Promo requests: status information for organisers
The way promo requests work currently is that there is no communication to the organisers of an event when e.g. the status of the promo request changes or the request has been delivered. This means that the organisers need to remember to check it manually every once in a while. I don't think this is a very good solution, especially if there's time constraints.

One possible solution could be to email organisers on status changes of promo requests for their events.
However, I think it would be better to first discuss whether this needs to be implemented in the website or not: it may be sufficient for the promocie to actively communicate themselves when they deliver something.

### Motivation
It's not that great to have to check over and over to see if promo is done.

- Have the promocie send emails to organisers when things are done.

### LINKED ISSUE #3483: promo weekly overview starts with Hi Paparazcie in the html file.
### Describe the bug

promo weekly overview email starts with Hi Paparazcie in the html file which is not very nice

NotAtec:
Closes #3005 .
Closes #3483.

### Summary

An update email will be sent out to the organising committees when the status of a promotion request is updated. There will also be a daily summary email sent out to the promocie with all updates.

### How to test

1. Create promotion request for an event on the admin page.

   i. Confirm no email sent at this point. (in terminal)

2. Change the status of the just created promotion on the promocie admin page.

   i. Confirm email sent to all organising committees (in terminal)

3. open a `manage.py shell` instance, and run the following:
    ```python
    from promotion.tasks import promo_update_overview_daily
    
    promo_update_overview_daily()
    ```

   i. Confirm that the daily update overview was created for the promocie.

4. In the `manage.py shell` run the following:
    ```python
    from promotion.models import PromotionRequest
    
    PromotionRequest.objects.filter(status_updated=True)
    ```

   i. Confirm this returns an empty QuerySet


website/promotion/models.py
```diff
@@ -106,7 +113,15 @@ def __str__(self):
     def save(self, **kwargs):
         if not self.publish_date and self.event:
             self.publish_date = self.event.start.date()
-        return super().save(kwargs)
+        oldstatus = None
+        if self.pk:
+            oldstatus = PromotionRequest.objects.get(pk=self.pk).status
+            if not self.status_updated:
+                self.status_updated = oldstatus != self.status
+        ret = super().save(kwargs)
```
DeD1rk, line 121:
```suggestion
        ret = super().save(**kwargs)
```

website/promotion/management/commands/sendpromodailyupdateoverview.py
DeD1rk, line 1: This one can be deleted

website/promotion/management/commands/sendpromoupdate.py
DeD1rk, line 1: This one can be deleted

website/promotion/emails.py
```diff
@@ -47,3 +47,42 @@ def send_daily_overview():
...
+            send_email(
+                to=[email],
+                subject="[PROMO] Status update",
+                txt_template="promotion/email/status_update.txt",
+                context={
+                    "updated_request": updated_request,
+                },
+            )
+        else:
+            print("No email for organiser")
```
DeD1rk, line 72:
```suggestion
            logger.info("No email for organiser")
```

DeD1rk: Or actually, I guess this case is unreachable?

NotAtec: > Or actually, I guess this case is unreachable?

It should indeed be unreachable, but for during testing I put it in anyways.
I could delete it now ig

DeD1rk (approved):

PR #3420, opened 2023-10-04
Overhaul registrations

DeD1rk:
Closes #3297.
Closes #2834.
Closes #1930.

Also closes #3426.

### Summary
This makes the registrations code a lot more straightforward. While it adds some duplication, it makes it so that it's easy to follow what happens when, as registrations and renewals aren't mixed up anymore.

### TODO

- [x] Add bulk approve/reject action back into the admin
- [x] Test views
- [x] Test admin
- [x] Test services
- [x] Test forms
- [x] Test models
- [x] admin fields that should be readonly
- [x] Do all of the TODOs in code

### How to test
Try every way of making registrations and renewals you can think of:

- [x] making a registration, paying in cash.
- [x] making a registration, paying with thalia pay.
- [x] making a registration while a user with the username already exists.
- [x] making a registration, then create a user with the username just before accepting the registration, and just before completing (paying) the registration.
- [x] and do this with different membership types and durations.
- [x] making a renewal in august, paying in cash
- [x] making a renewal in august, paying with thalia pay
- [x] making a membership upgrade renewal (in august), pay with tpay in august.
- [x] making a membership upgrade renewal (in august), pay with cash in august.
- [x] making a membership upgrade renewal (in august), pay with tpay in september.
- [x] making a membership upgrade renewal (in august), pay with tpay in september.
- [ ] making a renewal to study time in september (i.e. paying full price)
- [ ] Anything else?..

PR #3468, opened 2023-11-02
Fix deleting invoices upon dataminimization

DeD1rk:
### Summary
Sets a flag when deleting objects for data minimisation, and checks for that flag in the moneybird post_delete receiver to ignore deletions of registrations and renewals.

### TODO

- [x] I should really probably really really write a test. (yes no I actually will do that)
- [x] Fix the tests for registrations dataminimisation

I just removed the test in registrations, as I will soon refactor that a lot anyway (see #3420).

### DIFF
--- a/website/moneybirdsynchronization/signals.py
+++ b/website/moneybirdsynchronization/signals.py
@@ -158,20 +158,23 @@ def mark_invoice_outdated(sender, instance, **kwargs):
 
 
 @suspendingreceiver(post_delete, sender="registrations.Renewal")
-@suspendingreceiver(
-    post_delete,
-    sender="registrations.Registration",
-)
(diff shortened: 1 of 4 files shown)


JobDoesburg:
Some remarks:

1. Shouldn't we indeed implement this globally? Not just for registrations / renewals?
2. I'm not sure if this is the nicest thing of doing this (annotating the queryset). Can't you inspect the kwargs that you receive with the signal? Also, I do like your earlier idea to check if a payable is paid, and never delete in such instances. I think it is correct.

DeD1rk:
Checking for it being paid is neat, but I'm not sure if it's more general. Deleting an unpaid event registration for example shouldn't keep its invoice.

PR #3474, opened 2023-11-08
Member profile permission

### LINKED ISSUE #3444: Can use member list API without a membership
### Describe the bug

When somebody used to have access to the member list but not any more the list is still visible in the app -> API. On the website member list you get a 403 error, but apparently not in the API.

### How to reproduce

1. Be a no-longer-member.
2. Use the members API.

### Expected behaviour

403

ColonelPhantom:
Closes #3444.

### Summary

This adds an HasActiveMembership permission, so that non-member users cannot view the member list.

Additionally, I restricted the ability of non-member users to view profiles except their own in the final commit. Not sure if this is actually desirable.

### How to test

1. Have a non-member user
2. go to /api/v2/members, /api/v2/members/ID, /members/profile/ID, and view the 403 errors
3. go to /members/profile and /api/v2/members/me and check that theyre still accessible

### Additional context
@Scarletto wanted to do this via `members:read` scope, but that doesn't seem to be set or used anywhere. This implementation mirrors what is done for the normal view.

### DIFF
--- a/website/members/api/v2/permissions.py
+++ b/website/members/api/v2/permissions.py
@@ -3,5 +3,6 @@ from rest_framework.permissions import BasePermission
 
 class HasActiveMembership(BasePermission):
     """DRF permission equivalent of the `@membership_required` view decorator."""
+
     def has_permission(self, request, view):
         return request.member and request.member.has_active_membership()
--- a/website/members/api/v2/permissions.py
+++ b/website/members/api/v2/permissions.py
@@ -1,0 +1,6 @@
+from rest_framework.permissions import BasePermission
+
+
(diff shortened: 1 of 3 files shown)


website/members/api/v2/permissions.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework.permissions import BasePermission
+
+
+class HasActiveMembership(BasePermission):
```
DeD1rk, line 4:
```suggestion
class HasActiveMembership(BasePermission):
    """DRF permission equivalent of the `@membership_required` view decorator."""
```

DeD1rk (approved):

PR #3475, opened 2023-11-08
Switch to chached_db session backend

### LINKED ISSUE #3363: Switch to cached_db session backend
### What?

Once we have Redis set up (mainly for celery, #3357, #3361) we can use it to cache sessions.
See https://docs.djangoproject.com/en/4.2/topics/http/sessions/#using-cached-sessions

### Why?

A little performance boost for virtually no effort.

tvanonna:
Closes #3363 .

### Summary
This does exactly what is described in #3363, in which redis is used to cache sessions.

### How to test
Dirk said this was fine.

### DIFF
--- a/website/thaliawebsite/settings.py
+++ b/website/thaliawebsite/settings.py
@@ -744,6 +744,8 @@ CACHES = {
     },
 }
 
+SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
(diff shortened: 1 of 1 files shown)


DeD1rk (approved):
The change looks good. Let's see on staging if it really works and performs well.

Issue #941, opened 2020-03-18
Thalia Pay: also enable late-cancellation fines to happen with Thalia Pay


thaliawww-service:
In GitLab by @JobDoesburg on Oct 23, 2019, 21:40

### One-sentence description

Also enable late-cancellation fines to happen with Thalia Pay

### Motivation

Because it is weird to do let this happen via invoices if people normally can pay with Thalia Pay

### Desired functionality

Currently if somebody wants has said to pay with Thalia Pay, but cancels, we delete the payment. 
After the event, the treasurer should be able to determine the real amount of the fine and have some action available to add TPAY payments for everyone that was too late and deserves a fine (now we will probably need to reopen #821 heheheheh)

### Suggested implementation

thaliawww-service:
In GitLab by @se-bastiaan on Oct 25, 2019, 11:56

The treasurer is able to manually create payments in the admin so I believe this would already be possible.
And they can always add the payment to the batch manually.

JobDoesburg:
With moneybird sync this becomes more important

JobDoesburg:
Also relevant for #3453

PR #3476, opened 2023-11-09
Payment Requests

### LINKED ISSUE #3453: ThaliaPleasePay

mark-boute:
Closes #3453 and #941.

### Implementation

- We will extend the payment models with a PaymentRequest class, this model will store information about the request and the payment itself, it is also a Payable object. Each payed request will become immutable.

- We will add an AdminView which shows all requests, but the default filter only shows 'pending' (unpaid) requests. 

- We will add an AdminView for editing pending requests.

- There will be an AdminForm for adding new requests, here you can fill in details, and select any number of PaymentUsers (must be at least 1!) to whom the request apply. 

- !TODO think of how we would like to handle late-cancellation fines to happen with the payment-request system

- After creating a new request the user should be notified, we will do this by e-mail and by pushnotification. An admin can send-on-save or do it manually later

### Tasklist

- [x]  setup model
- [ ]  write tests for model
- [ ]  admin views
- [ ]  admin view tests
- [ ]  admin form
- [ ]  admin form tests
- [ ]  notification time 
- [ ]  email tests
- [ ]  a connection between fines and service
- [ ]  connection tests
- [ ]   (optional) reminder date

### Notes
- How can we ensure that Users with an expired Membership can still pay (both before the PaymentRequest is created and after if they didn't pay yet but their membership did expire)?


DeD1rk:
> Upon creating a new request the user should be notified, we will do this by e-mail and by pushnotification

This should probably be configurable or a separate manual action. I think it's useful if you can also silenty create a PaymentRequest, send it later, or resend it. 
Also, we might want to wait with adding pushnotifications until the app supports paying with a deeplink (very similar to how the app supports paying sales orders).

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
 
     class Meta:
         ordering = ("created_at",)
+
+
+class PaymentRequest(Payable):
+    requester = models.OneToOneField(Member, on_delete=models.CASCADE)
```
DeD1rk, line 487:
```suggestion
    created_by = models.OneToOneField(Member, on_delete=models.PROTECT)
```

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
...
+
+    request_timestamp = models.DateTimeField(
```
DeD1rk, line 489:
```suggestion
    created_at = models.DateTimeField(
```

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
...
+        verbose_name=_("request timestamp"),
+        blank=False,
```
DeD1rk, line 491:
```suggestion
        auto_now_add=True,
        editable=False,
        blank=False,
```

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
...
+        null=False,
+    )
+
+    payer = models.OneToOneField(PaymentUser, on_delete=models.CASCADE)
```
DeD1rk, line 495:
```suggestion
    payer = models.ForeignKey(PaymentUser, on_delete=models.PROTECT)
```

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
...
+
+    payment_timestamp = models.DateTimeField(
```
DeD1rk, line 497: I think you can just use `payment.created_at`

DeD1rk: But a `payment = OneToOneField(Payment...` is needed.

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
...
+        null=False,
+    )
+
+    notes = models.TextField(
+        max_length=256,
+        verbose_name=_("payment notes"),
+        blank=True,
+        null=True,
+    )
+
```
DeD1rk, line 522: There should probably be some field to decide (for bookkeeping purposes) whether the payable 'has to be paid sometime'? I don't know enough about that though.

website/payments/models.py
```diff
@@ -480,3 +481,78 @@ def __str__(self):
...
+    @property
+    def payed(self):
```
DeD1rk, line 524: Payable interface implementation has to be in a separate class (in a payables.py) to match how all other payables work.

website/payments/models.py
T8902: Should we confirm with roel/nick/someone else that knows about bookkeeping what requirements this field should uphold?

mark-boute:
> > Upon creating a new request the user should be notified, we will do this by e-mail and by pushnotification
> 
> This should probably be configurable or a separate manual action. I think it's useful if you can also silenty create a PaymentRequest, send it later, or resend it. Also, we might want to wait with adding pushnotifications until the app supports paying with a deeplink (very similar to how the app supports paying sales orders).

Changed to: `After creating a new request the user should be notified, we will do this by e-mail and by pushnotification. An admin can send-on-save or do it manually later`

PR #3482, opened 2023-11-13
Fix alumni mailinglist missing people

### LINKED ISSUE #3435: oldmembers mailinglist incomplete
### Describe the bug
The oldmembers mailinglist is incomplete, and some past members are not on the list.

Specifically, the member in question who sent us an email seems to have his old membership listed in the database as a benefactorship. We should probably investigate why that is and how to fix these memberships, or at least work around the problem.

DeD1rk:
Closes #3435.

### Summary
This changes the way the oldmembers mailinglist is determined to include alumni from before concrexit, of whom we often have no (accurate) membership history.

I branched off of the branch for #3420, as both will probably be in the same release, and it would need a harder rebase otherwise.

### How to test

1. Make a new member
2. Check that they can set, unset and reset the receive_oldmembers setting.
3. Make a new benefactor
4. Check that they cannot set the receive_oldmembers setting to True.


T8902 (approved):
Looks good

PR #3484, opened 2023-11-15
Update merchandise pages layout

### LINKED ISSUE #2057: Merchandise page redesign

The current merchandise page is a long list with lot's of text. Part of this problem is the board not hiding the merchandise items that are sold out, but I think some other layout than a list would help to improve the look.

View more images side by side, and make the images larger. The text is not very important for the merch at all, so it can be pushed to the background.

### Motivation

The board is getting new merch and would like the page to look better to get people interested in the merch.

Keep the page as is, because people will buy merch anyway through whatsapp promotion etc.

thijmengg:
Closes #2057.
feature: new front-end page for merchandise

### Summary
New front-end page for merchandise. There are more next to eachother. Also works on smaller devices, when clicking on merch you get more information about it and get redirected to a 404 when link to merch is wrong.

### How to test

1. go to the merchandise page
2. click on merch picture
3. if changing the link to, for example, id = 999, an 404 page occurs


website/merchandise/static/merchandise/css/product.css
DeD1rk, line 1: The CSS seems a bit overcomplicated to mee. Could you look into using bootstrap (grid layout) where possible? 

AlsoI think it would be a bit cleaner to remove the border around the photos, and the space between the header bar and the photo, if you get what I mean.

PR #3485, opened 2023-11-15
Increase profile data minimisation retention period

### LINKED ISSUE #3478: Increase profile retention period
### What?
Currently, Profiles are dataminimised 1 month after the member's last membership ends.
We should increase that period to 3 months. I have discussed this with @JeeVee11.

### Why?
It happens quite often that people forget to renew their membership, and decide to do it later than the first of october. In those cases, currently, their profile will already have been wiped. To make it easier for those people to still renew, we should extend the retention period a bit longer.

### How?
Change a constant somewhere in `members/services.py`, and update the privacy policy.

tvanonna:
Closes #3478.

### Summary

Changed the time after which the profiles of ex-members are data-minimised from 1 to 3 months (90 days). This change is made in the file 'website/members/service.py' as well as in 'website/singlepages/templates/singlepages/privacy_policy.html'.

### DIFF
--- a/website/members/services.py
+++ b/website/members/services.py
@@ -218,7 +218,7 @@ def execute_data_minimisation(dry_run=False, members=None) -> list[Member]:
         .distinct()
         .prefetch_related("membership_set", "profile")
     )
-    deletion_period = timezone.now().date() - timezone.timedelta(days=31)
+    deletion_period = timezone.now().date() - timezone.timedelta(days=93)
     processed_members = []
     for member in members:
         if (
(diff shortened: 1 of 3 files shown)


website/members/services.py
```diff
@@ -218,7 +218,7 @@ def execute_data_minimisation(dry_run=False, members=None) -> list[Member]:
         .distinct()
         .prefetch_related("membership_set", "profile")
     )
-    deletion_period = timezone.now().date() - timezone.timedelta(days=31)
+    deletion_period = timezone.now().date() - timezone.timedelta(days=93)
```
DeD1rk, line 221:
```suggestion
    deletion_period = timezone.now().date() - timezone.timedelta(days=90)
```

DeD1rk (approved):

DeD1rk:
@tvanonna actually the privacy policy change is currently at the section about registrations, not profiles.

PR #3486, opened 2023-11-15
Document mechanisms that reveal user pictures

### LINKED ISSUE #3409: Disclaimer that profile pictures might be visible on the internet
### Profile picture visible on google images
Right now some committees' members are visible on thalia.nu to logged out people (depending on the 'display members' field on active member groups). This means that their profile picture will be visible on google when searching their name which may influence what people want to put in their profile picture

### Possible solution
When people upload a picture get a small pop-up with a warning that their picture could become visible to people when joining a committee.

We should also add a note to the privacy policy.

ColonelPhantom:
Closes #3409.

### Summary

This documents that profile pictures may become public, and also documents the confusingly named `display_members` property that causes this to happen.

I thought it might be nice if a user could somehow check if their profile picture is shown via any committee, but wasn't sure where/how to show this.

The changes to the privacy policy probably require a version bump, but we will need this anyway for #3478 so I left that out here.

### How to test

1. Read the changes
2. Check that the help texts appear.


website/members/migrations/0048_alter_profile_photo.py
```diff
@@ -0,0 +1,30 @@
...
+        ("members", "0047_remove_profile_receive_magazine"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="profile",
+            name="photo",
+            field=thumbnails.fields.ImageField(
+                blank=True,
+                help_text="Note that your photo may be publically viewable and indexable by search engines in some cases. This happens when you are in a commitee with publically viewable members.",
```
DeD1rk, line 20:
```suggestion
                help_text="Note that your photo may be publicly visible and indexable by search engines in some cases. This happens when you are in a committee with publicly visible members.",
```

website/members/models/profile.py
```diff
@@ -206,6 +206,9 @@ class Profile(models.Model):
 
     photo = ImageField(
         verbose_name=_("Photo"),
+        help_text=_(
+            "Note that your photo may be publically viewable and indexable by search engines in some cases. This happens when you are in a commitee with publically viewable members."
```
DeD1rk, line 210:
```suggestion
            "Note that your photo may be publicly visible and indexable by search engines in some cases. This happens when you are in a committee with publicly visible members."
```

website/activemembers/models.py
```diff
@@ -104,6 +104,7 @@ class MemberGroup(models.Model):
 
     display_members = models.BooleanField(
         default=False,
+        help_text="With this enabled, the members of this committee will be viewable to anyone without logging in. Logged-in users can always view the list of members.",
```
DeD1rk, line 107:
```suggestion
        help_text="With this enabled, the members of this committee will be visible to anyone without logging in. Logged-in users can always view the list of members.",
```

DeD1rk (approved):

PR #3492, opened 2023-11-17
email changed

### LINKED ISSUE #3490: Different email for trust contact person Lisa
### What?

The email of Lisa is not her preferred email and she wants this changed .

### Why?

The personal email is not the preferred email right for things like this to be send too.

### How?

Change the email.

T8902:
Closes #3490.

### Summary

Changed the email of TCP lisa

### DIFF
--- a/website/singlepages/templates/singlepages/student_well-being.html
+++ b/website/singlepages/templates/singlepages/student_well-being.html
@@ -5,60 +5,71 @@
 {% block opengraph_title %}{% trans "Student well-being" %} — {{ block.super }}{% endblock %}
 
 {% block body %}
-    <section class="page-section">
-        <div class="container">
-            <h1 class="section-title">{% trans "Services from the university" %}</h1>
-            {% blocktrans %}
-                The university offers various services to help students. Below are the most important points of contact.
-            {% endblocktrans %}
-            <div class="row mt-4">
-                <div class="col-6 col-md-3 my-3">
-                    {% static "img/contact_persons/hallway-huygens.jpeg" as photo %}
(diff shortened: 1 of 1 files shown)


DeD1rk (changes requested):
All of the indentation changed while it's just a single line that actually has to change.

T8902:
> All of the indentation changed while it's just a single line that actually has to change.
Oh kut black formatted it weird didn't see that

DeD1rk:
But also, IIRC it was intentional that TCPs don't use their `@ru.nl` email: the conversations are confidential, so shouldn't be in a university-managed mailbox (and more importantly, definitely not in an `@members.thalia.nu` mailbox). I'm not sure if this has been noted anywhere, but did lisa consider it?

T8902:
> But also, IIRC it was intentional that TCPs don't use their `@ru.nl` email: the conversations are confidential, so shouldn't be in a university-managed mailbox (and more importantly, definitely not in an `@members.thalia.nu` mailbox). I'm not sure if this has been noted anywhere, but did lisa consider it?



Doubt she considered it, I'll discuss the necessity again with her.

DeD1rk:
I can't find anything about it in any minutes or reports, and it probably doesn't matter (the uni's professional confidential advisors also use @ru.nl) but for the sake of being accessible and not sounding official (as technically TCPs are supposed to mainly refer to the right trained professional), I think having personal addresses would still be preferable.

T8902:
After further discussion, the email will be changed later to an email that isn't @ru.nl however not right now.

PR #3517, opened 2023-12-06
Remove image rotation support

### LINKED ISSUE #3489: Drop photo.rotation field
### What?
Since #3324 newly uploaded photos don't get a rotation set. And I've also unrotated all existing photos. And reaxit 3.5.1 doesn't require the API to return a rotation either. So now we can drop the field from the model and API, and clean up any code that uses the rotation.

### Why?
It was messy, and is no longer needed, so we can make our code cleaner :)

### How?
Remove the field and all references to it. Remove it from serializers, and remove `photos/management/commands/unrotateimages.py` entirely.

ColonelPhantom:
Closes #3489.

### Summary

Remove photo.rotation. Also check in the migration that there are no rotated photos to prevent data loss.

### How to test

1. Check that it fails when there are rotated images
2. Check that it works otherwise
3. Grep for remaining references?
4. ???


DeD1rk (approved):
Nice :)

PR #3528, opened 2023-12-13
Fixed Album Pagination Erasing Search Terms

### LINKED ISSUE #3509: Albums pagination doesn't maintain search terms
### Describe the bug
If you open https://thalia.nu/members/photos/?keywords=borrel#photos-albums, then go to the second page using the pagination buttons, the search term is dropped.

### Expected behaviour

The search term remains

### Additional context

Could be since we introduced the shared paginated view template? So it's quite likely this occurs for other paginated filterable/searchable views as well.

tvanonna:
Closes #3509 .

### Summary

When changing pages on the list of albums, any search terms you inserted will no longer be deleted.

### How to test
1. Go to the list of albums on the Thalia Website
2. Choose a search word such that the results take up more than one page
3. Switch between pages and check whether the search word you pack has been erased

### DIFF
--- a/website/thaliawebsite/views.py
+++ b/website/thaliawebsite/views.py
@@ -49,19 +49,17 @@ class PagedView(ListView):
 
         page_range = range(page_range_start, page_range_stop)
 
-        # generates the string of keywords, and applies these to the base_url
-        keyword_string = ""
+        querydict = self.request.GET.copy()
 
-        for keyword in self.keywords:
-            if keyword_string == "":
-                keyword_string += f"{keyword}"
-            else:
-                keyword_string += f"+{keyword}"
+        if "page" in querydict:
+            del querydict["page"]
 
         context.update(
             {
                 "page_range": page_range,
-                "base_url": self.request.path + f"?keywords={keyword_string}",
+                "base_url": f"{self.request.path}?{querydict.urlencode()}&"
+                if querydict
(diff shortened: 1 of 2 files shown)


website/thaliawebsite/views.py
```diff
@@ -48,9 +49,19 @@ def get_context_data(self, **kwargs) -> dict:
...
+        for keyword in self.keywords:
+            if keyword_string == "":
+                keyword_string += f"{keyword}"
+            else:
+                keyword_string += f"+{keyword}"
+
         context.update(
             {
                 "page_range": page_range,
+                "base_url": self.request.path + f"?keywords={keyword_string}",
```
DeD1rk, line 64: Sorry I didnt explain properly. This needs to work with _any_ query parameters, not specifically the ones used in albums (because there are other PagedViews with other parameters. 
It should probably be something like this:
- Make a copy of request.GET, using QueryDict.copy() (see https://docs.djangoproject.com/en/5.0/ref/request-response/ for how it all works).
- Then in that copy, remove the 'page' key if it exists. 
- Then make the base_url by using request.path + QueryDict.urlencode().

(this urlencode part is important for security. As it is right now, query parameters are not escaped, so you can abuse it to e.g. include JS)

DeD1rk (approved):

PR #3540, opened 2023-12-20
Use alpine base image

### LINKED ISSUE #3026: Use alpine base image
### Describe the change
Right now the concrexit docker image is based on the standard python base image. It would be nice to use the alpine base image (`python:3.11-alpine`) instead. See https://medium.com/swlh/alpine-slim-stretch-buster-jessie-bullseye-bookworm-what-are-the-differences-in-docker-62171ed4531d.

### Motivation
Size, speed, stability.

### Current implementation
Bullseye.

### Suggested implementation
Alpine.

### Additional context
See `infra/concrexit/Dockerfile`, and perhaps the entrypoint.

T8902:
Closes #3026.

### Summary

Added alpine base image to switch from deb to alpine because it's smaller. Saves about 100MiB in the base image.

### DIFF
--- a/infra/concrexit/Dockerfile
+++ b/infra/concrexit/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11-slim-bullseye AS base
+FROM python:3.11-alpine AS base
 
 ENV PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1 \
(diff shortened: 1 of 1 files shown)


infra/concrexit/Dockerfile
```diff
@@ -52,9 +74,9 @@ ENV PATH=/venv/bin:$PATH \
     SENDFILE_ROOT="/volumes/media"
 
 # Install runtime dependencies.
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-    postgresql-client libjpeg-dev zlib1g-dev libwebp-dev libmagic1 && \
+RUN apk update && \
+    apk add \
+    postgresql-client jpeg-dev zlib-dev libwebp-dev libmagic runuser && \
     rm -rf /var/lib/apt/lists/*
```
DeD1rk, line 80: the apk cache isn't at `/var/lib/apt/lists` any more. For apk you can use --no-cache, or maybe buildkit's `RUN --mount=type=cache` (https://docs.docker.com/engine/reference/builder/#run---mounttypecache) to get better performance.

DeD1rk (approved):

PR #3544, opened 2023-12-23
Add local docker compose configuration

JAicewizard:
### Summary
This makes it easy to run concrexit locally with postgresql database instead of sqlite for the most realistic testing, with postgres, redis, nginx with tls, and a celery worker. This also runs on a single core to match performance characteristics.

You can simply `make rundocker` (`stopdocker` and `removedocker` to shut down / wipe all data), although you'll also want to know how to make a shell into the container or view the logs. For anything you might need, use `ENV_FILE_SUFFIX=.local docker compose -f infra/docker-compose.yml -f infra/docker-compose.local.yml <SUBCOMMAND>`.

Once running, you can open `https://localhost`. Note that this uses a self-signed certificate that your browser will complain about. On firefox you can ignore that, or set up an exception for the 'local CA' certificate to be trusted.

### How to test
1. run `make rundocker`


DeD1rk:
Oh wow I had no idea that setitng defaulted to disabling

DeD1rk:
Nice!

DeD1rk:
Oh wrong PR im commenting on but still

DeD1rk:
It's actually possible to make the nginx image create self-signed certs, so we can have a local setup that's very complete with minimal duplication. I'll set it up tomorrow.

JAicewizard:
Ah yeah I didn't feel like looking into that, but it was indeed annoying to have to duplicate that

infra/concrexit/Dockerfile
```diff
@@ -26,13 +27,15 @@ ENV PATH=/venv/bin:$PATH \
 
 COPY pyproject.toml poetry.lock ./
 
-RUN poetry install --only main --only postgres
+RUN echo $DJANGO_ENV
+RUN if [ $DJANGO_ENV="development" ]; then echo "_$DJANGO_ENV"; else echo "success!!"; fi
```
JAicewizard, line 31: Remind me to remove this

pyproject.toml
```diff
@@ -53,6 +53,18 @@ optional = true
...
 psycopg2 = "2.9.9"
 
+[tool.poetry.group.postgreslocal]
+optional = true
+
+[tool.poetry.group.postgreslocal.dependencies]
+psycopg2 = "2.9.9"
+django-template-check = "0.4.0"
+django-extensions = "^3.2.1"
+libsass = "0.22.0"
```
JAicewizard, line 63: Using this setup it also does not use the static css files generated during build for some reason, which is why I added libsass as a dep, but it should be doable without libsass like in prod

DeD1rk: I think the problem was in DJANGO_ENV=development. That tends to change static files and templates discovery

JAicewizard:
The settings changes are necessary for login and debug panel. (and other things that check csrf tokens)

JAicewizard:
looks good for the rest, but havn't tested but I assume it works

DeD1rk:
> The settings changes are necessary for login and debug panel. (and other things that check csrf tokens)

Hmmm I have no problem with the CSRF. That's probably fixed by using https and not running in debug mode.

JAicewizard:
Ah I see that it is indeed also running in staging instead of development. I do think there is value in running debug mode with postgresql, so I would keep the internal IP changes with a comment to make it easy for people to switch

PR #3565, opened 2024-01-22
"Live" registration

### LINKED ISSUE #3522: Enable registration button when it opens using a timer
### What?
Currently, the registration button only appears on a page refresh. When registrations open this causes a big storm of users who all refresh this page to get the button.

Using client side scripting to make the button appear we prevent these refreshes.

### Why?
Reduces server loads and prevents API-based gaming of registrations 

### How?
We would need some conditional javascript that can make the button appear on a timer. We should also add the required information for this to the API so that the app can do the same live view update.

ColonelPhantom:
Closes #3522.

### Summary

Adds a permission `create_registration_when_open`, that indicates a user will be able to register for an event as soon as it opens. Additionally, when this permission is true, the event page will contain a disabled register button and some JavaScript that will automatically enable it at the right time.

For the app, we can use this new permission in combination with `registration_start`.

### How to test

1. Create an event that opens in the future
2. Open the event page
3. See that there is a countdown
4. Ensure the button activates when the countdown is done.

### RFC
Should we maybe limit the countdown to e.g. something in the same day (e.g. what if a user has a stale tab?)

Is the styling OK?

### DIFF
--- a/website/events/services.py
+++ b/website/events/services.py
@@ -4,6 +4,7 @@ from django.db.models import Q
 from django.utils import timezone
 from django.utils.datetime_safe import date
 from django.utils.formats import localize
+from django.utils.timezone import timedelta
 from django.utils.translation import gettext_lazy as _
 
 from events import emails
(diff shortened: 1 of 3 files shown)


DeD1rk (approved):
I tweaked a few things (indeed, limited it to 2h before the event). I like how simple it turned out!

PR #3571, opened 2024-01-24
Make users re-enter minimized data on renewal

### LINKED ISSUE #3292: Require data-minimized users to fill in address info again when they re-register.
### Describe the bug
There are a number of users whose address info has been minimized, but who do have an active membership. I think that shouldn't be allowed. For these users, an old membership has existed, and a new membership was created some time later.

For example, see https://thalia.nu/admin/auth/user/1668/change/, whose membership had been expired for +- 2 months before creating a new one. 

In these cases, I think somehow we need to make people fill in their info again. 

### How to reproduce
1. Have your membership expire.
2. Wait long enough for data minimization to delete your address info.
3. Create a new membership.

### Expected behaviour
When creating the new membership, you're asked to fill in the minimized fields again.

T8902:
Closes #3292

### Summary

Making it so that people have to fill in their deleted information before being able to request a renewal.

### How to test

1. Make a user of which their data is minimized
2. Go to renewal page
3. Go to edit profile and fill in missing info
4. Go to renewal page and renew


ColonelPhantom (dismissed):
Looks reasonable, and the failing test seems expected (but please fix it ofc).

One major issue is that when there is a pending renewal, this does not prevent re-minimization. This is also mentioned in the issue:
> Profile dataminimization should not minimize profiles of users who have a renewal open (not rejected and not completed).

More minor, there is no 'vetting' of the user's birthday (although if you were a member, had it expire, and minimized away, you're probably old enough for there to be little incentive to lie).

There's also a UX issue in entering the birthday: there is no date-picker, nor is there any indication of the required format. This could be fixed by e.g. adding a datepicker, or changing the placeholder to "yyyy-mm-dd", so that the required format is clear.

DeD1rk:
For future reference, we had decided that both the 'race condition' where profiles might be re-minimized (between completing the profile and creating a renewal), and the lack of birthday verification are not a problem. 

I've fixed several small bugs, including the birthday UI

DeD1rk (approved):

Noah-marc (approved):
Looks good!

PR #3576, opened 2024-01-25
Improve fixtures and make them validate

### LINKED ISSUE #3573: Improve partner fixtures
### What?
Partner fixtures suck. There is no text, the addresses are invalid, and there are no local partners generated.

### Why?
Fixtures should be representative of realistic data, and not cause validation errors.

### Related
Other fixtures are also quite bad (for example, event fixtures also cause validation errors). We should probably open issues for this.

ColonelPhantom:
Closes #3573.

### Summary
Improves partner fixtures (local partners are now generated, and there is a description)

Some other missing or renamed fields are now also generated, so that fixtures pass `full_clean` now.

This is also verified, which means that editing and saving a fixture should now always work (instead of giving errors related to missing data).

### How to test

1. Run `make fixtures`
2. Check that it works

I think the fixtures are still slightly flaky (group membership sometimes gave an error about impossible random number generation), and are also not resistant to generating duplicates (two events with the same slug for example).

### DIFF
--- a/website/utils/management/commands/createfixtures.py
+++ b/website/utils/management/commands/createfixtures.py
@@ -4,6 +4,7 @@ import math
 import random
 import string
 from datetime import date, datetime, timedelta
+from decimal import Decimal
 from secrets import token_hex
 
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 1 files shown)


ColonelPhantom:
By the way, me and @DeD1rk also both had issues with disappearing image files for fixture-generated things? I still have no idea what causes that, to be honest.

ColonelPhantom:
>be me
>make PR changing fixtures
>fixtures check fails

It's because of the rather odd address validation for partners:  https://github.com/svthalia/concrexit/blob/acc14e5de2ba5ca745a8971af901bffbb6f64438/website/partners/models.py#L55-L62

DeD1rk (approved):
Awesome! Love that you found the missing caption for events too

PR #3577, opened 2024-01-25
Fix color scheme toggle in admin

### LINKED ISSUE #2984: Fix Django color scheme toggle
### Describe the bug
Since v4.2, Django Admin now has a toggle for switching between light and dark mode (and high contrast?). We don't properly support that, though.

### How to reproduce
1. Use the toggle

### Expected behaviour
Proper color schemes. Maybe even applied in the frontend too (not only admin)

DeD1rk:
Also solves an incorrect button color introduced in Django 5.0.

Closes #2984.

### Summary
Makes our color overrides get higher specificity than django's.

### DIFF
--- a/website/thaliawebsite/templates/admin/base_site.html
+++ b/website/thaliawebsite/templates/admin/base_site.html
@@ -70,7 +70,7 @@
             font-weight: normal;
             font-style: normal;
         }
-        :root {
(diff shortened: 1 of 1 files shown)


ColonelPhantom (approved):
Seems to work!

PR #3580, opened 2024-01-31
Fix Partner layout

### LINKED ISSUE #3513: Partner page has a wonky layout
### Describe the bug

The local partner info blocks are not nicely aligned, the logo's are different sizes too.

### How to reproduce

1. Go to the partner page
2. Scroll down
3. See a wonky layout

### Expected behaviour

The partner page has a nice consistent layout

### Screenshots

CleanShot 2023-12-06 at 14 50 03

rodepanda7:
Closes #3513.

### Summary
Images of local partners are padded on the top and bottom to make the image including padding always 250px. Images are 600px wide.

### How to test
1. (optional) Add images for the local partners to see the results better
2. Go to the partners' page
3. Better layout
partners

### DIFF
--- a/website/utils/media/processors.py
+++ b/website/utils/media/processors.py
@@ -42,8 +42,14 @@ def process_upload(image: DaVinciImage, **kwargs):
 def thumbnail(image: DaVinciImage, **kwargs):
     """Thumbnail an image to at most the given size.
 
-    If `kwargs["cover"]`, the image is cropped to the aspect ratio of the given size.
+    If `mode="contain"` (the default), the image is resized to fit within the given size.
+    The original aspect ratio is preserved, so the image may be smaller than specified.
+
+    If `mode="cover"`, the image is cropped to the aspect ratio of the given size.
     The thumbnails are saved to WebP format with lossy compression.
+
+    If `mode="pad"`, the image is resized to fit within the given size.
+    Padding is added to the edges to make the image the exact size specified.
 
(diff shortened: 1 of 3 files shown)


website/utils/media/processors.py
```diff
@@ -42,22 +42,32 @@ def process_upload(image: DaVinciImage, **kwargs):
...
-
-    if size[0] != pil_image.size[0] and size[1] != pil_image.size[1]:
-        pil_image = ImageOps.fit(pil_image, size, Image.Resampling.LANCZOS)
+    mode = kwargs.get("mode", "contain")
+
+    match mode:
+        case "contain":
+            pil_image = ImageOps.contain(pil_image, size, Image.Resampling.LANCZOS)
+        case "cover":
+            pil_image = ImageOps.fit(pil_image, size, Image.Resampling.LANCZOS)
```
DeD1rk, line 67: This changes the behaviour of the cover mode a bit, at least by scaling up if the original is smaller, but maybe also in other ways. So I'd prefer to keep that case unchanged. 
But love the match case stuff!

ColonelPhantom: What do you mean? I think that `cover` just calls `ImageOps.fit`, same as this PR.

For `contain` aka `cover=False` you are right that `ImageOps.contain` replaces our custom logic. But I don't see how our old version prevents scaling up; won't that just give a ratio greater than 1 and also scale up?

DeD1rk (approved):

PR #3607, opened 2024-02-22
fix alumni acounts not getting a contract because there not a member

### LINKED ISSUE #3603: Moneybird contacts are not made for alumni people
### Describe the bug

People who are alumni don't have a membership anymore even if they still have an account and thus don't get a moneybird contact which is bad.

### Expected behaviour

If someone has an account and makes a payment at a thalia event, they should get a moneybird contact.

Luukosaurus:
Closes #3603.

### Summary
Added a check for MoneybirdContact.DoesNotExist exception so that it does not crash when accounts which are not members try to make a contract. This fixes the issue where alumni people don't get a contract.


Noah-marc (approved):
Looks good!

PR #3616, opened 2024-02-27
Disable sentry cron monitoring

### LINKED ISSUE #3605: Disable Sentry cron monitoring
### What?
We need to disable the sentry 'cron' monitoring of periodic tasks.

### Why?
Sentry is making cron monitors paid after the beta.

### How?
I think it's a single line in settings.py, and maybe some cleanup on sentry to remove the existing monitors.

DeD1rk:
Closes #3605.


Noah-marc (approved):
Looks good

PR #3617, opened 2024-02-28
Re-enable HTML in registration status message

### LINKED ISSUE #3515: No registration message no longer accepts links
### Describe the bug
Since #2458 the event registration status (specifically no_registration_message) doesn't support URLs any more on the website (app is unaltered I think). I aslo think the formatting that happens now might be a bit risky.

### How to reproduce
Make an event with URL in the no_registration message and view it.

### Expected behaviour
A nice link.

### Screenshots
afbeelding

ColonelPhantom:
Closes #3515.

### Summary
Fairly trivial, bleach the registration status message so that e.g. links here work again

### How to test

1. Create an event without registrations
2. Set the message so that it contains a link
3. Observe that the link appears in the browser


ColonelPhantom:
One note, the only message that can be modified is the one for when there are no registrations possible. Optional registrations for example do not show this message, because it seems #2458 did not actually add the new fields to the admin :facepalm:

T8902 (commented):
Screenshot from 2024-03-06 18-24-09

ColonelPhantom:
I fixed the above by just getting rid of HTML in the message on the calendar (to prevent links appearing inside which seems annoying).

In addition, I added the other fields to the admin, since I did not do this in the past.

T8902 (approved):
Looks nice! :)

PR #3619, opened 2024-02-29
Update tcp info

Scarletto:
### Summary
Replaced Tom with myself as TCP.
Updated Lisa's picture.

### How to test
Go to the "student well-being" page (under Education).
See if the picture, the name, the hover-over email address and the on-click email creation work as intended.


Noah-marc (approved):
Code is all good, the test failing is because of the fact that it was a leap year and the 29th. If run again, the test won't fail

PR #3620, opened 2024-02-29
Fix leap year problem not being able to view mentor profiles

T8902:
### Summary

Profiles broke due to the leap year. This should now be fixed

### DIFF
--- a/website/members/services.py
+++ b/website/members/services.py
@@ -79,10 +79,12 @@ def member_achievements(member) -> list:
         name = f"Mentor in {mentor_year.year}"
         # Ensure mentorships appear last but are sorted
         earliest = date.today()
-        earliest = earliest.replace(year=earliest.year + mentor_year.year)
         # Making sure it does not crash in leap years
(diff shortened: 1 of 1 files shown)


Noah-marc (commented):
Hupsi, I did not see that tests were failing. I don't really understand why the tests are failing suddenly tho.

@T8902 & @DeD1rk do you have an idea ?

DeD1rk:
@Noah-marc test failures are a leap year problem

DeD1rk:
Test failure mught be different leap yer problems that occur only in test code

Noah-marc:
Yeah the tests fail because the year in the tests is not a leap year and thus querying the 29th fails. 
So it is actually a test that could only fail once ever 4 years. Lol 

I think it is fixed by simply making sure the test uses a leap year as well

DeD1rk:
Yup, tests can be updated to not use timezone.now().replace(year=...) or similar in a separate PR

DeD1rk (approved):

PR #3641, opened 2024-03-13
Update membership upgrade discount conditions to match HR

### LINKED ISSUE #3612: Implement HR changes to discounted membership upgrades/renewal
### What?
The huishoudelijk reglement has changed wrt the time in which (ex-)
members can use the discount when renewing/upgrading their membership to be study-long.

Check the HR precisely for the exact requirements.

### Why?
We need to match the official rules. It has to be released at the latest towards the end of this lecture year, to cover the wave of renewals in august/september and after. 

### How?
I think there's a nice place where the discount is applied either in the RenewalForm or in the Renewal model.

DeD1rk:
Closes #3612.

### Summary

A user now gets the discount also in the year after last having been a member.

### How to test

1. Make a user with a membership that has ended recently.
2. Make upgrade renewal.
3. Get discount.

### DIFF
--- a/website/registrations/forms.py
+++ b/website/registrations/forms.py
@@ -1,3 +1,5 @@
+from datetime import timedelta
+
 from django import forms
 from django.conf import settings
 from django.core.exceptions import NON_FIELD_ERRORS, ValidationError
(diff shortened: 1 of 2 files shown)


Noah-marc (approved):
Looks good

PR #3650, opened 2024-03-20
Import events into newsletter

### LINKED ISSUE #2162: Automatically add newsletter events for events in the upcoming week
It takes quite a while to manually select all events for next week in the newsletter. This can be automized

Either javascript or an admin action that automatically adds newsletter event objects.

### LINKED ISSUE #3317: Automatically add events that open to newsletter
In the newsletter below the introduction, there is usually a list of events which registrations open that week. This has to be done manually, which is quite a lot of work.

Ideally the website would generate the list on its own and add it too the newsletter.

### Motivation
I'm lazy :)

ColonelPhantom:
Closes #2162
Closes #3317

### Summary

- The board can now easily bulk-add events to the newsletter. This can probably use some nice defaults for the corresponding forms but I don't know what, @JeeVee11 probably does?
- The newsletter send page is now actually in the admin namespace
- A slight change in wording for the fine field in the admin
- Removed the `event` field for newsletter events, since this field was not used and kinda pointless (it was never used as far as I could see).

As you can see, apart from the headline feature there's a handful of refactors here as well.

I considered rewriting the newsletter app to use actual event references, but this is inflexible due to things like external events and would also require us to keep the current system alongside it to prevent breaking changes in our schema.

### How to test

1. Create a newsletter.
2. Click the nice pink `Import events` button.
3. Fill in the options.
4. Check that all the events you wanted appeared.


website/newsletters/forms.py
```diff
@@ -43,3 +33,125 @@ class Media:
...
+    def clean(self):
+        cleaned_data = super().clean()
+        if (
+            cleaned_data.get("event_start") is not None
+            and cleaned_data.get("event_end") is None
+        ):
+            self.add_error(
+                "event_end", _("This field is required if you specify a start date.")
+            )
+        if (
```
ColonelPhantom, line 116: I feel like there's a way to make this logic less ugly? Especially since Black formatted it in a horrible way.

ColonelPhantom:
By the way, do we really want the duplicate action buttons on the page? IMO it looks ugly, especially with the bottom one being below the save buttons and stuff. I'd vote for removing that bar and just having the actions on top only. Opinions?

JeeVee11:
I think that for defaults its the events of next week are the best.

Does this also include adding automatically adding events for which the registration opens in the next week, not just events that happen?

ColonelPhantom:
> Does this also include adding automatically adding events for which the registration opens in the next week, not just events that happen?

The idea is that you can fill in date ranges for both, and any event that satisfies either of them will be added.

I did just realize that what currently happens is that only events that satisfy both ranges will be added. Probably requiring only one of them would be much better.

ColonelPhantom:
I fixed the logic. Additionally, I restored the 'event' field on the form to make sure it is still easy to add/edit single events.

DeD1rk (approved):
dt is best wel fucking vet

PR #3652, opened 2024-03-27
Add and use CaseInsensitiveFilter for promotion admin assignment filter

### LINKED ISSUE #3085: The same names with different capitalisation are seen as different
### Describe the bug
When claiming promo requests in the admin site, if the claimant name is entered twice, first without capital and then with one. It is counted as two different persons.

### Expected behaviour
The same name with different capitalisation should still count as the same name.

ColonelPhantom:
Closes #3085.

### Summary

Adds a `CaseInsensitiveFilter` inside the promo admin, and uses it for `assigned_to`.

Maybe this should be in some more central directory, so we can reuse it more widely? I don't know if we want this in other places too. The code is also a bit messy, but it seems pretty much all of it is needed and Django offers no easier way, sadly.

I also used in-Python code for managing the lower-casing, since I think doing it directly would require Postgres or something and we don't want to break SQLite.

### How to test

1. Go to promo admin
2. Assign some requests to different capitalizations of the same name (e.g. one to `quinten` and the other to `Quinten`)
3. Observe that these are merged together in one filter option

### DIFF
--- a/website/promotion/admin.py
+++ b/website/promotion/admin.py
@@ -1,11 +1,83 @@
 """Registers admin interfaces for the models defined in this module."""
+
 from django.contrib import admin
 from django.contrib.admin import ModelAdmin
+from django.db import models
+from django.db.models.functions import Lower
 
 from events.services import is_organiser
 from promotion.forms import PromotionRequestForm
 
 from .models import PromotionChannel, PromotionRequest
+
+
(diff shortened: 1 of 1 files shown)


DeD1rk (approved):

DeD1rk:
Nice job getting the facets in there as well :)

PR #3629, opened 2024-03-06
Thabloid check for register

Luukosaurus:
Closes #3613.

### Summary

Added a check box to the registration form for receiving paper thabloids and removed redundant code for user language

PR #3654, opened 2024-04-03
Remove Language_code attribute since it is no longer used

T8902:
Closes NaN.

### Summary

Removed Lanuage code since it crashes the site and is not used anymore.

### How to test

1. Go to registration/ benefactor registration page
2. Try to register
3. See if it breaks


T8902:
Alread fixed in pr #3629 so will close it now

PR #3666, opened 2024-04-17
Removed question mark and changed message if committee is inactive

### LINKED ISSUE #3635: committees with no members are not handled well.
### Describe the bug
Committees with no members have a weird piece of text instead of members list.
Currently it says "This group doesn't have any members?"

### How to reproduce

1. Go to https://thalia.nu/association/committees/46/
2. This group doesn't have any members?

### Expected behaviour
This should not have a members section, or have a text indicating there are no members. This should not be a question, but a statement.

### Additional context
Ideally there should be a type of committee "Former committee", under which the bit committee would fall, with a text saying this committee no longer exists. However this is probably really annoying, and changing the "?" into a "." is a good enough fix.

I heard people wanted to solve issues with as little character changes as possible, so here you go, it almost cant get any smaller.

Luukosaurus:
Closes #3635 .

### Summary

Added other message is committee is inactive and not just empty
### How to test

Make a committee and check and uncheck the active box in the admin panel


website/activemembers/templates/activemembers/membergroup_detail.html
```diff
@@ -54,8 +54,13 @@ <h1 class="section-title">
 							{% membergroup_member_card member %}
 						</div>
 					{% empty %}
-						{% trans "This group doesn't have any members?" as empty %}
-						{% alert 'danger' empty extra_classes='col-10 offset-1' %}
+                        {% if membergroup.active%}
```
DeD1rk, line 57:
```suggestion
                        {% if membergroup.active %}
```

website/activemembers/templates/activemembers/membergroup_detail.html
```diff
@@ -54,8 +54,13 @@ <h1 class="section-title">
...
+                            {% trans "This group doesn't have any members." as empty %}
+                            {% alert 'danger' empty extra_classes='col-10 offset-1' %}
+                        {% else %}
+                            {% trans "This group is inactive." as empty %}
+                            {% alert 'danger' empty extra_classes='col-10 offset-1' %}
```
DeD1rk, line 62:
```suggestion
                            {% alert 'warning' empty extra_classes='col-10 offset-1' %}
```

DeD1rk (approved):

PR #3667, opened 2024-04-17
Removing apiv1 #3227

### LINKED ISSUE #3227: [Proposal] remove API v1
### Describe the change
We still have API v1 in the codebase. As far as I know there are no remaining consumers for this API, and as such maintaining it bloats up the codebase with a lot of unnecessary code. Additionally, a lot of tests currently work against API v1, and it would be good to comb through those to move them to v2 if there is no equivalent API v2 test.

### Motivation
Reduce maintenance burden.

### Current implementation
Both API v1 and v2

### Suggested implementation
Only API v2

thijmengg:
Closes #3227 

### Summary
API V1 is now removed, but tests still have to be checked such that it can be merged. I will do this later. Therefore its a draft PR


T8902:
closing since @thijmengg is not going to work on this anymore

PR #3671, opened 2024-04-24
Disable opening hidden/processing albums

DeD1rk:
### Summary
Fixes that you can open an album from the direct url despite it being hidden and/or processing.


T8902 (approved):
Tried it with both the slugs of hidden as well as processing albums and it seems to work so nice :)

PR #3682, opened 2024-05-15
Redirect to next or LOGIN_REDIRECT_URL after loggin in with django otp

### LINKED ISSUE #3680: Fix redirects to login not redirecting back afterwards
### Describe the bug
Sometimes redirects to login don't redirect back afterwards. For example during an OAuth authorization flow, but possibly also in other normal cases. We need to find out why and fix it.

### Expected behaviour

If for some page you need to log in, you should be redirected back instead of to `/`.

vincevd1:
Closes #3680.

### Summary
Fixes the issue that users always get redirected to '/' after login

### How to test
1. Go to 'Photos'
2. You will get redirected to login with a next parameter
3. Login with your credentials
4. You are redirected to /members/photos/

### DIFF
--- a/website/thaliawebsite/views.py
+++ b/website/thaliawebsite/views.py
@@ -1,6 +1,5 @@
 """General views for the website."""
 
-from django.conf import settings
 from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth.views import LogoutView as BaseLogoutView
 from django.contrib.auth.views import PasswordResetView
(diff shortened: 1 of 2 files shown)


website/thaliawebsite/views.py
```diff
@@ -77,7 +78,12 @@ class RateLimitedLoginView(LoginView):
     @method_decorator(ratelimit(key="ip", rate="30/h"))
     @method_decorator(ratelimit(key="post:username", rate="30/h"))
     def post(self, request, *args, **kwargs):
-        return super().post(request, *args, **kwargs)
+        response = super().post(request, *args, **kwargs)
+
+        if request.member:
+            return redirect(request.GET.get("next", settings.LOGIN_REDIRECT_URL))
```
DeD1rk, line 84:
```suggestion
            return redirect(self.get_success_url())
```

DeD1rk (approved):
🍻

PR #3704, opened 2024-05-29
add pushnotifications for thabloid

### LINKED ISSUE #3615: Send pushnotification when Thabloid is uploaded
### What?
Send a pushnotification when the thabloid is uploaded, like we do with albums/newsletters. Opening it should open the PDF download. (make sure the S3 link doesn't expire quickly)

### Why?
This should make it easier to read the thabloid digitally, reducing the number we need to print and increasing the number of readers.

### How?
Take inspiration from `pushnotifications/signals/newsletters.py`.

Noah-marc:
Closes #3615 .

### Summary

Added the feature that a pushnotification is send when a new thabloid is uploaded. 
### How to test

1. Go to admin and upload a thabloid
2. See in the admin that under messages (for pushnotifications) a message for that thabloid has been added


DeD1rk (approved):
:)

PR #3716, opened 2024-06-05
Handle 400 when archiving an already archived moneybird contact

### LINKED ISSUE #3713: Handle moneybird exception 400: contact cannot be archived
### Describe the bug
[CONCREXIT-18Z](https://thalia.sentry.io/issues/4723527093/)

When trying to delete/archive a dataminimized use on moneybird, we sometimes get a 400. I checked with Roel and it appears that these users have already been archived. So I'm thinking we either need to just clean up these 6 cases on production once, or we can ignore the 400 and proceed normally by just deleting the moneybirdcontact from out database.

### How to reproduce

1. In moneybird, look up one of the contacts that we tried to archive at night and see that they're archived already.

DeD1rk:
Closes #3713.

### Summary
This assumes that '400: Contact can not be archived' implies that the contact has already been archived.

### DIFF
--- a/website/moneybirdsynchronization/administration.py
+++ b/website/moneybirdsynchronization/administration.py
@@ -56,6 +56,9 @@ class Administration(ABC):
             if description:
                 msg += f": {description}"
 
+            self.status_code = status_code
+            self.description = description
+
             super().__init__(msg)
 
     class Unauthorized(Error):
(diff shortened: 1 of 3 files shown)


T8902 (approved):
:beers:

PR #3720, opened 2024-06-07
Fix check if you are allowed to order food

### LINKED ISSUE #3714: Ordering food for an event you are not registered for
### Describe the bug

You can order food for an event you are not registered for. I was not registered for the 'Spring Beach Party' but I could order food.

### Expected behaviour

I expect you can only order food for an event you are registered for or an event where registration is not necessary.

DeD1rk:
Closes #3714.

Supersedes #3718 which had a wrong target branch.

### Summary
Made it so that you are not able to order food when the ordering has not started or the event has ended. Also fixed that you were able to order food to registration required events if you are not registered or are in the queue of the event you are ordering food at.

### How to test

Make a new Food event and set the time to expire in like a minute. Try to order food when the food event has ended if you are back to the index page of pizza, then this is correct. You should not be able to order food when the event has expired. Make a new food event and do not register for the event that requires registration, then try to order food if you are redirected to the index page of pizza, this is correct. You should again not be able to order food.

### DIFF
--- a/website/events/templates/events/event.html
+++ b/website/events/templates/events/event.html
@@ -256,9 +256,17 @@
                                     </th>
                                     <td>
                                     {% if event.food_event == event.food_event.current or event.food_event.just_ended %}
-                                        <a href="{% url "pizzas:index" %}" class="btn btn-primary">
-                                            {% trans "Order" context "pizzas" %}
-                                        </a>
+                                        {% if event.registration_required and not registration.is_invited %}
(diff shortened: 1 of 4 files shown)


DeD1rk:
@BPC2004 I forgot Wednesday but while this fixed the actual problem, it needs some frontend changes as well. We need to make it clear to users in advance that they can't order and why, and not just allow them to press the button and _then_ see an error.

DeD1rk (commented):
🐛

T8902 (approved):
:crab:

PR #3723, opened 2024-06-09
Fix bookstore link

### LINKED ISSUE #3599: Change bookstore links
### What?

From the 4th quarter on we will switch to another bookstore, intertaal (https://intertaal.nl/). Can you change this on the site? I am probably missing some information, but that is a work in progress. Once I have more information for on the website, I will let you know. 

~ soppie

T8902:
Closes #3599.

### Summary

Put the new book link instead of the old one at @Sophie2104 her request. 
### How to test

1. Go to book page
2. Click on the hyperlink

### DIFF
--- a/website/education/templates/education/books.html
+++ b/website/education/templates/education/books.html
@@ -16,7 +16,6 @@
             <a href="https://wo4you.nl/"
                target="_blank" rel="noopener">here</a>.
             As a member of Thalia you receive a 20% discount on study
-            <!-- To be announced by board -->
             books ordered at this webshop.
         {% endblocktrans %}
     </p>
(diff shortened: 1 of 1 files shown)


DeD1rk (changes requested):
I'm pretty sure the text arpund it also needs to change to reflect the company's actual name, and whatever discount deal there might be.

T8902:
> I'm pretty sure the text arpund it also needs to change to reflect the company's actual name, and whatever discount deal there might be.

Ah yes I might have been a bit lazy good point

T8902:
After discussion with board: There will be discussion to make this page "hidden" not sure how that would look exactly but the contract with the book store holds until April 2025 so we can't remove this page yet unless they are able to get rid of the contract sooner.

PR #2126, opened 2021-12-23
Add (optional) foreign key from payment to payable

JobDoesburg:
Closes #2034 and closes #1957

### Summary
A clear and concise description of the changes that you made. What bug did you solve? Or what feature did you add?

### How to test
Steps to test the changes you made:

PR #3737, opened 2024-06-21
Payment.payable_object property to get payable

### LINKED ISSUE #1957: Provide a way to link payments to payables
Sometimes thing still break with payments. Currently, there is only little that we can solve in the backend

- Introduce a new permission for 'linking payments to payables'
- change the admin widget for payments to also display a SelectWidget, to link a new/other payment to a payable

### Motivation
To be able to manually fix things for payables.

Not doing this

### LINKED ISSUE #2034: Filter for payments without payable linked
Currently, it is almost impossible to search for payments that are not linked to a payable object. Such payments, however, are interesting to investigate. There should be an easy way to filter for payments without a payable, in the admin

A filter in the admin

### Motivation
Have an easy way to check if payments are payments of a payable, or that the are "independent"

Not doing this

DeD1rk:
Supersedes #2126.
Closes #1957.
Closes #2034.

### Summary
This makes use of prefetching the <payable model>.payment OneToOneFields' reverse relations to efficiently get the payable for payments.

As this does not introduce a duplicate relation in the database, it should be more robust.
In the long term, I do think moving to *only* a GenericForeignKey from Payment to payable would be better, that seems like a change that is very likely to break something. So I'd rather just get this working properly in a way that is certain not to break stuff.

Replacing the 1to1's with a reverse generic FK would still be nice to do some day, but I don't think we should have both as in #2126.

### How to test

1. Try out the new filter in the admin
2. See the hyperlink from a payment changeview to the payable object's change view


T8902 (approved):
Tested looks good :)

PR #3764, opened 2024-08-01
Fix incorrect template location

DeD1rk:
Fixes an incorrect template path that crashed uploading an album.

Closes [CONCREXIT-1F4](https://thalia.sentry.io/issues/5628520484/).


T8902 (approved):
Top

PR #3775, opened 2024-08-14
Upgrade to Django 5.1

### LINKED ISSUE #3774: Upgrade to Django 5.1
We should upgrade to Django 5.1, the newest version which was just released.

### What?

Should be mostly plug-and-play. A clear exception is that we still use `get_storage_class` which has been deprecated.

### Why?

We should keep up-to-date with current Django versions. Additionally, this upgrade unblocks multiple other issues: #3436 and I think it was #3622 but @mark-boute knows which one it is I think.

### How?

We should upgrade the dependency, and fix the `get_storage_class` call.

ColonelPhantom:
Closes #3774 

### Summary
Upgrades to Django 5.1. Also uses the new `{% querystring %}` functionality. Additionally, it removes a use of `get_storage_class` because it was removed in this version. I looked at its implementation which just uses `import_string` so I decided to replace it with that.

### How to test

1. Check that the site still runs.
2. Check that private media still loads.
3. Check that pagination still works correctly.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -119,13 +119,13 @@ files = [
 
 [[package]]
 name = "babel"
-version = "2.15.0"
+version = "2.16.0"
 description = "Internationalization utilities"
 optional = false
 python-versions = ">=3.8"
 files = [
-    {file = "Babel-2.15.0-py3-none-any.whl", hash = "sha256:08706bdad8d0a3413266ab61bd6c34d0c28d6e1e7badf40a2cebe67644e2e1fb"},
-    {file = "babel-2.15.0.tar.gz", hash = "sha256:8daf0e265d05768bc6c7a314cf1321e9a123afc328cc635c18622a2f30a04413"},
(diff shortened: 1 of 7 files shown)


ColonelPhantom:
One problem seems to be remaining:

> [FileField](https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.FileField) now raises a [FieldError](https://docs.djangoproject.com/en/5.1/ref/exceptions/#django.core.exceptions.FieldError) when saving a file without a name.

However, I'm not sure what name to give to these file fields, so I'm not sure how to work around this.

ColonelPhantom:
Apparently we have a `newsletter_filename` function that overrides the filename field already...

Why the hell does Django mandate that we set it if it overrides it later anyway??

Anyway, I think using that function to get the name would be neater?

ColonelPhantom:
Actually, it seems like upload_to is intended to override the given filename? So I think it is fine as is. Unfortunately I cannot review as I made the PR :(

DeD1rk (approved):

PR #3777, opened 2024-08-16
Fix current membership is inactive future membership

### LINKED ISSUE #3776: "Renew membership" notification even after renewing/extending membership and still being member
### Describe the bug
After renewal from normal membership (which is until 1 september 2024) to benefactor (from 1 sept. 2024 until 1 sept. 2025) successfully I know get the banner that I'm not a member anymore and need to renew.

Almost all (member specific) parts of the website seem to work, except viewing other peoples profiles which result in a '403: Premission denied' error.

### How to reproduce
1. Renew your Thalia membership while still Thalia member for the next year from member to benefactor
2.  Go through the registration process just fine and get a 'Renewal successful" mail
3. See the banner "You're currently not a member of Thalia. [Renew your membership](https://thalia.nu/user/membership/) to get access to all parts of the website." appear on top of the website
4. Get 403 errors when clicking on a members profile

### Expected behaviour
To be seen as a member of Thalia until 1 sept. 2024 and as benefactor the year after.

### Screenshots

image

image

DeD1rk:
Closes #3776.

### Summary
Before this, two things were done in an ugly/inaccurate way:

1. `Member.current_membership` returned the _latest_ membership, even if it hadn't begun yet.
2. Memberships created in august started on the 1st of september (so technically, sjaars were no members during intro).

But before 0f6488a2ed2973ce3ae8d2446f48359a88d3c7cf, we did not check for the `membership.since` date when checking if a membership is active. This meant that the above 2 things together gave the expected behaviour. So adding that check broke the expected but hacky behaviour.

In this PR, I got rid of both weird things. Memberships normally start at the date they're completed (typically i.e. paid). For memberships completed in august, that means that the last few days/weeks of august are given 'for free' as they were before, but just also stored that way. And the `current_membership` property and friends now do what one would expect: return only _current_ memberships.

### How to test

1. At least simulate introduction registrations as they will happen in a few days.

### DIFF
--- a/website/members/models/member.py
+++ b/website/members/models/member.py
@@ -184,7 +184,7 @@ class Member(User):
 
         return (
             self.profile.event_permissions in ("all", "no_drinks")
-            and self.current_membership is not None
+            and self.has_active_membership()
         )
 
     @property
--- a/website/members/models/member.py
+++ b/website/members/models/member.py
@@ -99,39 +99,47 @@ class Member(User):
         return f"{self.get_full_name()} ({self.username})"
 
     def refresh_from_db(self, **kwargs):
-        # Clear the cached latest_membership
-        if hasattr(self, "_latest_membership"):
-            del self._latest_membership
+        # Clear the cached current- and latest_membership
+        if hasattr(self, "_current_membership"):
+            del self._current_membership
+        if hasattr(self, "current_membership"):
+            del self.current_membership
         if hasattr(self, "latest_membership"):
             del self.latest_membership
 
         return super().refresh_from_db(**kwargs)
 
-    @property
+    @cached_property
     def current_membership(self) -> Membership | None:
         """Return the currently active membership of the user, None if not active.
 
-        Warning: this property uses the *cached* `latest_membership`.
-        You can use `refresh_from_db` to clear it.
-        """
-        membership = self.latest_membership
-        if membership and not membership.is_active():
-            return None
-        return membership
-
-    @cached_property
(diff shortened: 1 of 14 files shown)


website/members/models/member.py
```diff
@@ -99,39 +99,47 @@ def __str__(self):
...
+            del self._current_membership
+        if hasattr(self, "current_membership"):
+            del self.current_membership
         if hasattr(self, "latest_membership"):
             del self.latest_membership
 
         return super().refresh_from_db(**kwargs)
 
-    @property
+    @cached_property
```
T8902, line 112: Whats the benefit of changing this to a cached property?

DeD1rk: It doesn't do the query multiple times if called multiple times and not prefetched

T8902:
image


when trying to pay for a registration using cash

T8902:
> image
> 
> when trying to pay for a registration using cash

correction, any payment

website/members/models/member.py
T8902: Ah fair enough

DeD1rk:
> > when trying to pay for a registration using cash
> 
> correction, any payment

Looks like you're running on a database from your study-long branch

T8902:
> > > when trying to pay for a registration using cash
> > 
> > 
> > correction, any payment
> 
> Looks like you're running on a database from your study-long branch

Yes I am just stupid

T8902 (approved):

PR #3783, opened 2024-08-28
Renewal reminder emails

### LINKED ISSUE #3753: Reminders for renewals
### What?
We delete old registration and renewal requests, and we also send notifications to the board for (somewhat less old) registration requests. We should also add reminders for renewal requests.

### Why?

Symmetry between the two, and ensures that renewals are also handled (although it is more likely for a renewal that the person in question would ask).

### How?

We should just add this in the same place the other reminders are sent. This is in `website/registrations/tasks.py`.

ColonelPhantom:
Closes #3753.

### Summary

Send renewal reminder emails, just like for new registrations.


Noah-marc (approved):
Looks good

PR #3792, opened 2024-09-10
Fix album processed email warnings

DeD1rk:
This prevents warning about directories existing in the uploaded archive, and fixes  crash because of a bug in the email template.

Closes [CONCREXIT-1GZ](https://thalia.sentry.io/issues/5810557399/).

### Summary

- Ignores directories inside the upload (these are zipinfo/tarinfo objects that only indicate directories existing, regardless of their content).
- Fixes the template to loop over warnings correctly (which makes it possible to send warnings now instead of crash)
- Reorganizes the email and uploaded signal to outside of the transaction, so crashing during the email or while doing facedetection does not cancel the saving of the album.

### How to test

1. Upload an album with some non-photos in it as well as a directory.
2. See an email being sent (to your console if you do it locally) with warnings about the non-photos, and not about the directory.

### DIFF
--- a/website/photos/tasks.py
+++ b/website/photos/tasks.py
@@ -41,24 +41,24 @@ def process_album_upload(
             album.is_processing = False
             album.save()
 
-            # Send signal to notify that an album has been uploaded. This is used
-            # by facedetection, and possibly in the future to notify the uploader.
-            album_uploaded.send(sender=None, album=album)
+        # Send signal to notify that an album has been uploaded. This is used
+        # by facedetection, and possibly in the future to notify the uploader.
+        album_uploaded.send(sender=None, album=album)
 
-            if uploader is not None:
-                # Notify uploader of the upload result.
-                send_email(
-                    to=get_member_email_addresses(uploader),
-                    subject=("Album upload processed completed."),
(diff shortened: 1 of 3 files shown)


website/photos/tasks.py
```diff
@@ -22,46 +22,49 @@ def process_album_upload(
     archive_upload_id: str, album_id: int, uploader_id: int | None = None
 ):
     upload = TemporaryUpload.objects.get(upload_id=archive_upload_id)
-    archive = upload.file
 
     try:
         album = Album.objects.get(id=album_id)
     except Album.DoesNotExist:
         logger.exception("Album %s does not exist", album_id)
-        archive.delete()
```
T8902, line 31: This shouldnt make a difference right?

T8902: since archive = upload.file will just point to the correct object? Or am i overlooking something

DeD1rk: Should be equivalent yes, I just got confused by the mixed use and want to avoid any risk that it might not be equivalent if somehow the tmp upload object gets modified

T8902: Fair enough

T8902 (approved):
Good :)

PR #3794, opened 2024-09-10
Send all data to moneybird as JSON

DeD1rk:
Closes [CONCREXIT-1FN](https://thalia.sentry.io/issues/5732597812/).

### Summary
According to the engineers at moneybird, this may be a workaround for the 500 they've been returning. Their parsing seems to be failing on deep nested JSON in the formdata.

If it works, we'll know it soon. If not, it's still good practice to use the right contenttype.

Update: seems to work. I hotfixed it in the worker and reran the synch.


DeD1rk:
This already ran in prod so I'll merge it as is.

PR #3795, opened 2024-09-11
Remove libmagic from dependencies

T8902:
Closes @mark-boute .

### Summary

Remove libmagic as a needed dependicies by removing a library that uses it and crashes if it's not present


mark-boute (approved):

mark-boute:
This closes me?

PR #3796, opened 2024-09-11
Add 2FA help page

### LINKED ISSUE #3790: Add more information about 2FA
### What?
Add more information about 2FA, most notably what the two options entail, and a bit more explanation about how to use an authenticator app. 

### Why?
Hello, it is boomer Nienke checking in. I had to set up 2FA today, and as someone that does not have a lot of experience with it (except for the forced 2FA uni uses), I found it a confusing experience. I was unfamiliar with the two options presented, so I think a bit more explanation of what they are would be nice. I tried both. I didn't understand the first at all when I tried it, because I am an idiot and when I read Google authenticator I thought it had to do something with the Thalia google suite account or whatever, and I do not have the Google authenticator app. Then I tried the second option, but my browser got confused (apparently, as Ties explained to me, this is if you have a password manager, but the Firefox password manager is not good enough for some reason), and did not explain to me why it got confused. Then I got frustrated, went to Studievertraging, where Stijn helped me through it. Apparently the Microsoft Authenticator app also works, which is still a bit of black magic to me (as in, it is not at all logical why Microsoft Authenticator works with the Thalia website; did the security courses fail me? Probably. It might also be that I took them ages ago when authenticators were not so popular), but hey it works. 

All this is to say: the whole process is a bit confusing if you are a boomer/stupid/not a frequent 2FA user like me. A bit more info would be appreciated

### How?
A link to a separate page with explanation on the black magic? A bit more information on the different pages with the different steps? A quick explanation of what the two options are for noobs? Mentioning multiple authenticator app options would also already be nice. And maybe mention that you should install those on your phone. Esp mentioning Microsoft Authenticator might be nice since everyone should have it because of uni. 

Boomer Nienke checking out

ColonelPhantom:
Closes #3790.

### Summary

Adds a help page for the 2FA wizard. @NienkeWessel do you think this is clear?

### DIFF
--- a/website/templates/two_factor/help.html
+++ b/website/templates/two_factor/help.html
@@ -1,0 +1,19 @@
+{% extends 'simple_page.html' %}
+
+{% block page_content %}
+<h1>{% block title %}2FA instructions{% endblock %}</h1>
+
+<p>We require 2FA before you are able to access the administration section of the website. This way, we ensure that users who can see and modify data (including personal data) have secure accounts.</p>
+
+<p>In order to set up 2FA, our website offers a step-by-step wizard. However, the wizard can be unclear if one is not familiar with 2FA terminology, so we hope this page will provide some more clarity.</p>
+
+<h2> Method </h2>
+<p>The wizard asks what method you want to use for 2-factor authentication. If you're unsure, you should pick the default of "Token generator" here. This works via an authenticator app (such as Google Authenticator, Microsoft Authenticator, Authy, or any other TOTP app).</p>
+
+<h2> Token generator setup </h2>
+<p>The easiest way to proceed here is to use the QR code. To use this, open your 2FA app, select the option to add a new (generic) account. This should enable you to scan the QR code. You will then get an initial authentication code that you should enter into the webpage. Alternatively, you can enter the displayed secret code.</p>
+
+<h2> Backup codes </h2>
+<p>After setting up 2FA, you will also get a list of backup codes. These are useful to ensure you can still get into your account if you lose access to your 2FA app.</p>
+
+{% endblock %}
--- a/website/templates/two_factor/help.html
+++ b/website/templates/two_factor/help.html
@@ -8,12 +8,12 @@
 <p>In order to set up 2FA, our website offers a step-by-step wizard. However, the wizard can be unclear if one is not familiar with 2FA terminology, so we hope this page will provide some more clarity.</p>
 
 <h2> Method </h2>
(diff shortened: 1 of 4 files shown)


website/templates/two_factor/help.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block page_content %}
+<h1>{% block title %}2FA instructions{% endblock %}</h1>
+
+<p>We require 2FA before you are able to access the administration section of the website. This way, we ensure that users who can see and modify data (including personal data) have secure accounts.</p>
+
+<p>In order to set up 2FA, our website offers a step-by-step wizard. However, the wizard can be unclear if one is not familiar with 2FA terminology, so we hope this page will provide some more clarity.</p>
+
+<h2> Method </h2>
+<p>The wizard asks what method you want to use for 2-factor authentication. If you're unsure, you should pick the default of "Token generator" here. This works via an authenticator app (such as Google Authenticator, Microsoft Authenticator, Authy, or any other TOTP app).</p>
```
DeD1rk, line 11:
```suggestion
<p>The wizard asks what method you want to use for 2-factor authentication. If you're unsure, you should pick the default of "Token generator" here. This works via an authenticator app (such as Google Authenticator, Microsoft Authenticator, Authy, and other TOTP apps, as well as some password managers such as Bitwarden and 1Password).</p>
```

website/templates/two_factor/help.html
```diff
@@ -0,0 +1,19 @@
...
+
+<h2> Token generator setup </h2>
+<p>The easiest way to proceed here is to use the QR code. To use this, open your 2FA app, select the option to add a new (generic) account. This should enable you to scan the QR code. You will then get an initial authentication code that you should enter into the webpage. Alternatively, you can enter the displayed secret code.</p>
+
+<h2> Backup codes </h2>
+<p>After setting up 2FA, you will also get a list of backup codes. These are useful to ensure you can still get into your account if you lose access to your 2FA app.</p>
```
DeD1rk, line 17:
```suggestion
<p>After setting up 2FA, you will also get a list of backup codes. These are useful to ensure you can still get into your account if you lose access to your 2FA app. Of course, you can always ask the Technicie for help to reset your 2FA.</p>
```

DeD1rk (commented):
Some textual remarks, and maybe you could link to the help page from inside the wizard as well? Just with a line like 'not sure what to do? read the _help page_. ' somewhere

website/templates/two_factor/help.html
T8902:
```suggestion
<p>After setting up 2FA, you will also get a list of backup codes. These are useful to ensure you can still get into your account if you lose access to your 2FA app. Of course, you can also always ask the Technicie for help to reset your 2FA.</p>
```

DeD1rk (approved):

PR #3813, opened 2024-10-10
Add info page for iCal feed

### LINKED ISSUE #3812: Make iCal feed more discoverable
### What?
The calendar feeds should be easier to find for new users, and could maybe also use some documentation on how to use them.

### Why?
Yesterday I revealed this feature to someone who wanted it but didn't know it existed, leading me to think that there are more people who miss its existence.

### How?
I was thinking we should at least add text to the buttons leading to the feed. We could also replace them by a single button leading to a help page with the iCal feed links.

### Alternatives you have considered 
We have the event confirmation emails. I'm not sure if we do anything like this, but are those able to easily be added to one's calendar by attaching an invite? That would also make it easier for people to keep their personal calendars up to date. (We could of course also do both!)

ColonelPhantom:
Closes #3812.

### Summary
Adds an information page for the iCal feed, instead of the obscure buttons on the main events page.

Note: styling is currently still horrendous!

### DIFF
--- a/website/events/templates/events/index.html
+++ b/website/events/templates/events/index.html
@@ -1,5 +1,5 @@
 {% extends "simple_page.html" %}
-{% load i18n static personal_feed %}}
+{% load i18n static %}
 {% get_current_language as LANGUAGE_CODE %}
 
 {% block title %}{% trans "Calendar"|capfirst %} —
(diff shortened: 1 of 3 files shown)


DeD1rk:
Oh I totally missed that you did this already, and opened #3816. Your text is much more explanatory, so let's take that (mine was made fully by copilot, so it's not the best but also didn't take any time).

PR #3816, opened 2024-10-21
Improve ical discoverability

### LINKED ISSUE #3812: Make iCal feed more discoverable

DeD1rk:
Closes #3812.

### Summary
Moves the iCal feed urls to a separate page with explanations.
Also adds some action points to the 'welcome' email.

# TODO:
- [x] Update welcome email txt variant.
- [x] Try out welcome email.

### DIFF
--- a/website/events/templates/events/ical_help.html
+++ b/website/events/templates/events/ical_help.html
@@ -1,0 +1,31 @@
+{% extends "simple_page.html" %}
+{% block page_title %}iCal feed instructions{% endblock %}
+
+{% block page_content %}
+    <p>This page provides information on how to use an iCal feed to stay updated with Thalia events.</p>
+    <h3>What is an iCal feed?</h3>
+    <p>
+        An iCal feed is a calendar format that allows you to subscribe to events and have them automatically appear in your calendar application.
+        We offer two different feeds: one with all Thalia events, and one that is personalized for you, with only the events that you've registered for.
+    </p>
+    <h3>How to subscribe to an iCal feed</h3>
+    <p>To subscribe to an iCal feed, follow these steps:</p>
+    <ol>
+        <li>Copy the iCal feed URL from below.</li>
+        <li>Open your calendar application (e.g., Google Calendar, Apple Calendar, Outlook).</li>
+        <li>Find the option to add a new calendar subscription by URL.</li>
+        <li>Paste the iCal feed URL into the provided field.</li>
+        <li>Save the new calendar.</li>
+    </ol>
+    <p>In some cases, simply opening the URL may navigate to your calendar application too.
+    <h3>iCal feed URLs</h3>
+    <p>There are two iCal feed URLs available:</p>
+    <ul>
+        <li>All events: <a href="{{ all_events_feed }}">{{ all_events_feed }}</a></li>
+        {% if user.is_authenticated %}
+            <li>Personal feed: <a href="{{ personal_feed }}">{{ personal_feed }}</a></li>
+        {% else %}
+            <li>Personal feed: <a href="{% url 'login' %}?next={% url 'events:ical-help' %}">Log in to access your personal iCal feed</a></li>
+        {% endif %}
+    </ul>
+{% endblock %}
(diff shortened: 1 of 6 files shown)


T8902 (approved):
Toppie :beer:

website/events/templates/events/ical_help.html
```diff
@@ -0,0 +1,34 @@
...
+    </p>
+    <h3>How to subscribe to an iCal feed</h3>
+    <p>To subscribe to an iCal feed, follow these steps:</p>
+    <ol>
+        <li>Copy the iCal feed URL from below.</li>
+        <li>Open your calendar application (e.g., Google Calendar, Apple Calendar, Outlook).</li>
+        <li>
+            In your calendar application find the button to add an (iCal) calendar via its URL.
+            For example, in Google Calendar, this is "Other calendars" and then "Via URL".
+            In the macOS Calendar the option is under the "File" menu and is called "New calendar subscription".
```
ColonelPhantom, line 19: I think the double "is" is worse grammatically than single "is". The option is ((located here) and (called that)), rather than The option (is located here) and (is called that).

website/events/templates/events/index.html
```diff
@@ -16,26 +16,15 @@
...
        target="_blank"
-       class="btn btn-primary first-right"
+       class="btn btn-primary first-right d-flex align-items-center"
        data-bs-toggle="tooltip"
        data-bs-placement="top"
-       title="{% trans 'iCal feed (all events)' %}"
+       title="Configure iCal feed"
     >
-        <i class="fas fa-calendar-alt"></i>
+        <i class="fas fa-calendar-alt me-2"></i>iCal feed
```
ColonelPhantom, line 26: "iCal feed" is quite deep into the realm of jargon. Calling the button something like "import calendar" or "synchronize calendar" would probably be more user-friendly.

Even though Thalia's audience is CS people, we should not assume familiarity with this kind of jargon; consider 2FA setup for an example where this led to issues.

website/members/templates/members/email/welcome.txt
```diff
@@ -13,6 +13,12 @@ Now that you're a member, you can join the official Thalia Promotion WhatsApp gr
 Through this group you will be kept up to date with Thalia events and other matters
 in the association. You can join the group via the following link: {{ base_url }}/whatsapp/.
 
+And finally, there are some other optional next steps for you:
+- Subscribe to a personalized calendar feed to see events in your calendar app:
+  {{ base_url }}{% url "events:ical-help" %}
+- Upload a reference picture for face detection, to you get an overview of all the photos you're on:
```
ColonelPhantom, line 19: "to you" -> "so you" or "to be able to" (also in HTML version)

website/events/views.py
```diff
@@ -248,3 +249,20 @@ def get(self, request, *args, **kwargs):
...
+
+    template_name = "events/ical_help.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context["all_events_feed"] = settings.BASE_URL + reverse("events:ical-en")
+        if self.request.member:
+            token = FeedToken.objects.get_or_create(member=self.request.member)[0].token
+            context[
+                "personal_feed"
```
ColonelPhantom, line 265: `personal_feed` was previously a template tag; we should either delete the code for the tag, or use that instead of manually setting context here.

ColonelPhantom (commented):
Bit late since this was already merged, but had some remarks (mostly textual, and some dead code).

Issue #3728, opened 2024-06-12
Improve statistics


DeD1rk:
### What?

Add some new statistics, and change some of the current ones.

- The food orders pie chart is a mess. It could be filtered on the top-10 products, or on products that have been sold in the past year.
- I would like a graph with the number of active members over time (committee memberships in total, and distinct number of active members).
- Maybe include the number of society members (over time) as well.
- I would like some kind of statistics on event attendance. For example, a histogram of (non-cancelled, non-queued) event registrations, both for events with required registration and optional registration, over time. That could be by year in a histogram, or maybe by quarter. Or a line plot with the months on x axis, and with a line for each year or academic year (which allows for easy comparison between the same month on current and previous years.
- Maybe statistics on `sales` product sales, similar to the pizza piechart, or something over time.

### Why?

Such statistics could be a valuable tool for making policy. In yesterday's general meeting (2024-06-11) we saw some policy that was based hunches about statistics, but not on real numbers.

### How?
Here: https://github.com/svthalia/concrexit/blob/2d870c3eae327f6fb5547cd79df87000ab0e0a53/website/members/views.py#L192

JAicewizard:
color scheme of the statistics is sometimes too dark, and sometimes too light (on light mode). This should be made less dark and less light

T8902:
This one seems closed by PR #3820? Or is there still more changes needed for this to be closed @DeD1rk

annalorenz123:
Not closed, statistics can still be improved in other ways. My PR was just a start.

T8902:
Ok are you going to work on that or should i assign someone else?

annalorenz123:
yes me and luuk

PR #3820, opened 2024-10-23
Small improvements on statistics page

annalorenz123:
#3728.

### Summary

- Added current study year to number of members per year chart and it now uses October instead of September to get statistic.
- Added pie chart for top 10 borrel orders.
- Change pizza pie chart to only show top 10.

### How to test

1. Go to statistics


website/members/views.py
```diff
@@ -19,6 +19,8 @@
 import activemembers.services as activemembers_services
 import events.services as event_services
 import pizzas.services
+import sales
```
DeD1rk, line 22:
```suggestion
```

DeD1rk (approved):

PR #3823, opened 2024-11-03
Introduce uv and ruff formatting

### LINKED ISSUE #3817: Switch to ruff for formatting
### What?
[ruff](https://docs.astral.sh/ruff/) has a formatter and import sorter that is much faster than black and isort. I think it would be nice to switch to that.

### Why?
Nicer dev experience.

### How?
- Change the depenencies.
- Update configurations in pyproject.toml and .pre-commit-config.yaml.
- Update the Makefile
- Update the README
- Update vscode settings to enable formatting with ruff onsave, and disable isort and black.

### LINKED ISSUE #3818: Switch from poetry to uv
### What?
[uv](https://docs.astral.sh/uv/#getting-started) is like poetry. We can replace poetry with it.

### Why?
`uv` is much faster, and has some additional features. Mainly, uv can (and will automatically) also manage your python installation. That is, when doing `uv sync` it will e.g. install python 3.13 if that's the version configured in pyproject.toml, regardless of your OS. So I think it can simplify setup by quite a bit.

One (temporary) downside is that `uv.lock` is not yet supported by Dependabot. There is WIP on that front, but it seems to be taking some time.

### How?
- Update the pyproject.toml for the new format.
- Update the README.
- Update the Dockerfiles.
- Update CI and the pre-commit config.
- Also do the same in the miniconcrexit-tutorial repo, and see what steps are needed to get uv installed in github codespaces and apply those there and optionally here.

DeD1rk:
Closes #3818.
Closes #3817.

Note that dependabot doesn't support uv yet. So either we wait for that, or accept that it's not working yet and update deps manually in the mean time.


T8902 (approved):
Joeppie :beers:

PR #3827, opened 2024-11-06
Disabled js i18n API

### LINKED ISSUE #3627: Disable js i18n API
### What?
We should turn off the i18n API, and make the frontend not use it any more. 

### Why?
Currently, despite us not using translations, JS is still calling `/i18n/js/`. This is not necessary and makes the site slower.

### How?
Probably read the Django docs. We might need to tweak some settings and JS, and remove the urls.

LucAngevare:
Closes #3627 .

### Summary
cool shit


DeD1rk (approved):
🍻

PR #3840, opened 2024-11-13
Add extra dropdowns for every year in payment history

### LINKED ISSUE #3839: Enable viewing entire payment history
### What?
Make it so you can view your entire payment history on the view payments in your account. You can currently only see 1 year back.

### Why?
It is available to change the URL and go further back, then 1 year would be nice to be able to do this via the GUI.

### How?
Add the extra buttons to view payments at your account.

BPC2004:
Closes #3839.

### Summary
Made it so you can look through your payment history with the GUI. 

### How to test
1. Go to account and then view payments.
2. Then open all the dropdown menus and see if the links work.


website/payments/views.py
```diff
@@ -145,14 +146,21 @@ def get_queryset(self) -> QuerySet:
 
     def get_context_data(self, *args, **kwargs):
         filters = []
-        for i in range(13):
+
+        for i in range(85):
```
DeD1rk, line 150: 85 is kind of a weird number?... Is it based on the legal storage period for payment info? For the user it might be nice if years in which there were no payments (or at least years before their first payment) are not shown.

website/payments/views.py
```diff
@@ -144,15 +145,27 @@ def get_queryset(self) -> QuerySet:
         )
 
     def get_context_data(self, *args, **kwargs):
+        current_time = timezone.now().replace(tzinfo=None)
+        nr_months = 0
+        earliest_membership = self.request.member.earliest_membership
+        difference = relativedelta(current_time, earliest_membership.since)
```
ColonelPhantom, line 151: This crashes if a user with no memberships tries to open it:
image

ColonelPhantom: Suggestion that I think does not break code coverage: add `if earliest_membership else None` to the end of this line, and do the same but with `else 0` for `nr_months`

website/partners/static/partners/css/style.scss
```diff
@@ -116,3 +116,7 @@
         }
     }
 }
+
+.wrapped-link {
+    word-wrap:break-word;
+}
```
T8902, line 122: This is not the right pr for these changes

website/partners/templates/partners/partner.html
```diff
@@ -38,7 +38,7 @@ <h4 class="mt-4">{% trans "Address" %}</h4>
                     {% if partner.link %}
                         <h4>{% trans "Website" %}</h4>
                         <p>
-                            <a href="{{ partner.link }}" rel="noopener" target="_blank">{{ partner.link }}</a>
+                            <a href="{{ partner.link }}" rel="noopener" target="_blank" class="wrapped-link">{{ partner.link }}</a>
                         </p>
                     {% endif %}
```
T8902, line 44: Same here

T8902 (changes requested):

website/payments/views.py
```diff
@@ -144,15 +145,33 @@ def get_queryset(self) -> QuerySet:
...
+            if earliest_membership
+            else None
+        )
+        nr_months = (
+            difference.years * 12 + difference.months + 1 if earliest_membership else 0
+        )
+
         filters = []
-        for i in range(13):
+        for i in range(min(nr_months, 85)):
```
T8902, line 161: Why 7 years and a month?

ColonelPhantom: The old value is also a year and a month. That's necessary to show results from exactly 1 (or now 7) years ago. Also keep in mind that Python ranges are exclusive.

If 0 is August 2025, then 12 is August 2024, and 84 is August 2018.

T8902: Makes sense

PR #3843, opened 2024-11-13
Add announcements list API and refactor persistent announcements

### LINKED ISSUE #3830: Add announcements to API
### What?
Add an announcements endpoint to the API.

### Why?
We would like to be able to show announcements in the app.

### How?
The website currently collects and displays announcements. We should collect these with plain-text as well as rich-text, and add an endpoint to fetch the plain-text version (for example).

ColonelPhantom:
Closes #3830.

### Summary
Adds an announcement API and refactors persistent announcements to fit into it more easily by being more like normal announcements.

### How to test

1. Go to `/api/v2/announcements/announcements/`
2. Check that all announcements are in there

### TODO
We should maybe add an API endpoint to close an announcement (in addition to `/announcements/close-announcement`).

### DIFF
--- a/website/announcements/api/v2/serializers.py
+++ b/website/announcements/api/v2/serializers.py
@@ -1,10 +1,15 @@
-from rest_framework import serializers
-
 from announcements.models import Slide
+from rest_framework import serializers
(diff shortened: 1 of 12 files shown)


T8902 (approved):
Toppie

PR #3848, opened 2024-11-18
Feature/study long to benefactor

### LINKED ISSUE #3844: Easier to become benefactorer with until the end of studies membership
### What?

Right now and with the coming implementation of being a member until the end of your studies, a member will need someone with admin rights to actually grant them a benefactorers membership and they cannot register for it themselves without help. This should be made easier.

### Why?

Because people with a membership until the end of their studies should still be able to become benefactorer once they are done studying without help.

### How?

Probably allow only either extension of the current membership or a new membership as benefactorer.

T8902:
Closes #3844

### Summary

Make it possible for members with a membership until the end of their studies to extend their membership as a benefactor once they stopped studying

### How to test

1. make member with membership until end of studies
2. end their membership
3. try extend as benefactor

### DIFF
--- a/website/registrations/views.py
+++ b/website/registrations/views.py
@@ -12,7 +12,9 @@ from django.utils.decorators import method_decorator
 from django.views import View
 from django.views.generic import CreateView, FormView
 from django.views.generic.base import TemplateResponseMixin, TemplateView
+
 from django_ratelimit.decorators import ratelimit
+
 from members.decorators import membership_required
 from members.models import Membership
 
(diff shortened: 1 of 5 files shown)


website/registrations/views.py
```diff
@@ -289,7 +291,10 @@ def get_form(self, form_class=None):
 
     def post(self, request, *args, **kwargs):
         request.POST = request.POST.dict()
-        if request.member.latest_membership.type == Membership.BENEFACTOR:
+        if (
+            request.member.latest_membership.type == Membership.BENEFACTOR
+            or request.member.latest_membership.study_long
```
DeD1rk, line 296: If I'm not mistaken, the frontend will stull allow selecting 'member' instead of benefactor? You can prevent that by customizing the field's 'choices' as is done a few lines above this

T8902: It will not

PR #3853, opened 2024-11-22
remove make_random_password

T8902:
Closes [CONCREXIT-1K0](https://thalia.sentry.io/issues/6082646906/)

### Summary

See https://docs.djangoproject.com/en/5.1/releases/5.1/ and https://docs.djangoproject.com/en/5.1/releases/4.2/

### DIFF
--- a/website/activemembers/gsuite.py
+++ b/website/activemembers/gsuite.py
@@ -25,7 +25,7 @@ class GSuiteUserService:
         return get_directory_api()
 
     def _generate_password(self, member: Member):
-        alphabet = string.ascii_letters + string.digits + string.punctuation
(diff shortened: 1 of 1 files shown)


website/activemembers/gsuite.py
```diff
@@ -22,13 +24,17 @@ def directory_api(self):
             return self._directory_api
         return get_directory_api()
 
+    def _generate_password(self, member: Member):
+        alphabet = string.ascii_letters + string.digits + string.punctuation
```
DeD1rk, line 28: Let's not use all of the punctuation options. Some of them are pretty weird for in a password.

```suggestion
        alphabet = string.ascii_letters + string.digits + "!@#$%^&*-_=+?"
```

DeD1rk (approved):

PR #3859, opened 2024-11-27
Add count/next/previous to announcement list

### LINKED ISSUE #3830: Add announcements to API

ColonelPhantom:
Closes #3830.

### Summary
Adds count/next/previous to announcement list api


DeD1rk (changes requested):
Looks like a mess to me

ColonelPhantom:
We're going to do this in the app instead.

PR #3860, opened 2024-11-27
Feature/reimbursements

### LINKED ISSUE #3712: Add reimbursements
**This is very similar to the example issue we made for the Technicie workshop. It seems like an actual good idea for the real concrexit as well.**

### What?
Within Thalia, members often need to request reimbursements from the treasurer:

When a committee spends money to organize something that will be funded by Thalia, the committee fills in a paper form, attaches the receipt from a payment they made, and hands the filled in form to the treasurer. The treasurer then checks whether the reimbursement form is correct, and approves it. Then, Thalia transfers money back to the committee(member) that spent some money.

Here is the form that is currently used: https://thalia.nu/association/documents/document/58/.

I think it would be great if people can submit their reimbursement requests online.

### Why?
Printing out the form, and filling it in on paper is tedious. You need to print, which is a pain on campus, and most students don't have printers in their room. It's also not practical for the treasurer to keep the books on paper.
So if we can do it on Thalia's website, that would make many peoples' lives a little bit easier.

The next thing that makes this more valuable, is that we may be able to automate processing them on moneybird a little bit more. The current process is:
1. Treasurer receives a form on paper.
2. On moneybird, they make a _purchase invoice_, with:
  - The person who paid for it as contact.
  - The date from the receipt
  - The amounts on one or multiple invoice rules. Each rule gets an amount, VAT rate and ledger account. 
3. They make a payment to reimburse the requester.
4. Once the payment shows up in moneybird, they link the purchase invoice document to the payment.  

Assuming that we already get digital versions of the reimbursement request, we can automate this partially as follows:
1. The treasurer approves the reimbursement requests (perhaps in batch) on the website.
2. Then, the website creates _typeless documents_ on moneybird.
  - The reference contains a unique id of the reimbursement request on concrexit, and a short topic.
  - The date is taken from the request (has to be provided by the requester).
  - The requester is set as contact.
  - The website uploads the attachment pdf or picture (maybe be want to support multiple?).
3. The typeless document with attachments then shows up on moneybird in the 'incoming' tab, where the treasurer can fill in the document type. Moneybird automatically tries to determine the invoice rules, but this is for the treasurer to check and correct.

Now, the attachments are visible on moneybird which means the treasurer doesn't need to look in the paper book to check things. 

### How?
Deviating slightly from the tutorial variant of this issue, I think for our real implementation it would be better to stay entirely within the admin. People without any permissions have no business making reimbursement requests, so keeping active member stuff in the admin is a slightly better separation of concerns IMO. 

So:
- We add a `reimbursements` app.
- Add a `ReimbursementRequest` model with the necessary fields. We add either a `status` (awaiting review, appproved, processed) or only `is_approved` field. One (pdf or image) document can and must be attached. Maybe we also want to support multiple attachments.
- Allow only people with 'create reimbursementrequest' permission to create a reimbursement request.
- Allow people with something like a `review reimbursement` permission to approve the request or maybe to otherwise change its status.
- Keep track of who created and approved a reimbursement.
- Notify the creator by email when the status changes.
- Maybe email the treasurer periodically or when a request is created too?
- In the admin, normal users see only the requests made by themselves. Users with the `review reimbursement` permission see all requests.
- The admin of course should have filters by status, search by name of the creator, etc.
- The treasurer can use an action to bulk approve requests.
- Once approved:
  - We push a typeless document to moneybird as described above.
  - Maybe we also monitor it from moneybird every night to check whether it's been processed there and use that to edit the status on concrexit? This is very optional though. 

@nvoers @JobDoesburg what do you guys think about this? And just to be sure, do you agree this would be good enough to be able to get rid of the paper form entirely?

T8902:
Closes #3712.

### How to test

1. Go to 'admin'
2. Click on 'reimbursement'
3. Make a 'reimbursement'
4. With or without iban

### TODO:

- [x] Maybe email the treasurer periodically or when a request is created too? (Probably later addition)
- [x] Decide on bank accounts: should people have an active bank account (but not necessarily with a mandate) if they want to create a reimbursement, or can they fill in an IBAN?
- [x] Test moneybird
- [x] Immediate moneybird push on acceptance


DeD1rk:
https://developer.moneybird.com/api/documents_receipts/

Issue #2958, opened 2023-04-05
Education Exam/Summary count incorrect.


DeD1rk:
### Describe the bug
For some courses it says there are >500 documents. That's clearly not correct. Something seems to be wrong here: https://github.com/svthalia/concrexit/blob/ce784be158c2e26afa9d389d67065db1cb1a716c/website/education/views.py#L48-L58

### How to reproduce
Go to https://thalia.nu/education/courses/

ColonelPhantom:
Okay, this does actually go wrong, but I haven't managed to reproduce locally even with #3890.

ColonelPhantom:
Found the problem: https://docs.djangoproject.com/en/5.1/topics/db/aggregation/#combining-multiple-aggregations

PR #3890, opened 2025-01-15
Optimize summaries page with queryable properties

ColonelPhantom:
Closes #ISSUE.

### Summary
Adds queryable properties for summary and exam counts. Then, these are prefetched immediately, which should greatly speed up the summaries page.

### How to test

1. Open summaries page
2. See that results are the same
3. Observe that there are fewer queries to the database and that hopefully the page loads faster

### DIFF
--- a/website/education/models.py
+++ b/website/education/models.py
@@ -1,8 +1,12 @@
 import datetime
 
 from django.db import models
+from django.db.models import Count, Q
 from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
+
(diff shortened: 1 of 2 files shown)


DeD1rk (approved):

DeD1rk:
😍

ColonelPhantom:
I actually just noticed I also tried this earlier but it gave problems: #2958 

However there I did not check for acceptedness, which was quite likely to be the problem.

PR #3908, opened 2025-02-25
Fix deep links on android

JAicewizard:
Closes nothing.

### Summary
This should fix the app links not working on android, and allow the debug app to also open debug links

### How to test

1. Go to an events page
2. Click on open in app


DeD1rk (approved):
Very nice!

PR #3909, opened 2025-02-26
Add extra bookstore

### LINKED ISSUE #3907: Ebook Store Link
### What?
Please add this link: <https://thalia.vitalsource.com/> to the 'Book Sale' section of the website.

### Why?
We have entered a contract with another bookseller that will provide us with ebooks, Vital Source!! 

### How?
I think it would be really cool to make both of them buttons, rather than just a link and some description, but just adding another link another description will suffice.

T8902:
Closes #3907 

### Summary

Add extra bookstore link because we needed that

### DIFF
--- a/website/education/templates/education/books.html
+++ b/website/education/templates/education/books.html
@@ -11,10 +11,13 @@
 {% block page_content %}
     <p>
         {% blocktrans trimmed %}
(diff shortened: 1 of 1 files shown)


ana-cullen (approved):
Good job :-)

PR #3948, opened 2025-06-18
Fixed ordering of the member list page.

### LINKED ISSUE #3916: Pagination of users seems broken
### Describe the bug

one Member pages 15 and 16 is duplicate.

### How to reproduce

1. Go to member list
2. Click on page 15
3. Click on page 16
4. See double member

### Expected behaviour

all members should only show up once in the member list

### Screenshots

No screenshots due to privacy concerns
### Additional context

The user does have the same url thus they are not a duplicate in the database it seems to be a pagination bug might not be our bug.

BPC2004:
Closes #3916.
bug

### Summary
For the ordering of the member list page if the first name is the same the last name and pk are also now considered for the ordering.

### How to test
Create 30 users with the same first name but different last names so that the entire page is filled up. And then check if a user is shown twice.


website/partners/static/partners/css/style.scss
```diff
@@ -116,3 +116,7 @@
         }
     }
 }
+
+.wrapped-link {
```
ColonelPhantom, line 120: These changes are unrelated to this PR and should not be in here.

website/partners/templates/partners/partner.html
```diff
@@ -38,7 +38,7 @@ <h4 class="mt-4">{% trans "Address" %}</h4>
                     {% if partner.link %}
                         <h4>{% trans "Website" %}</h4>
                         <p>
-                            <a href="{{ partner.link }}" rel="noopener" target="_blank" class="text-break">{{ partner.link }}</a>
+                          <a href="{{ partner.link }}" rel="noopener" target="_blank" class="text-break">{{ partner.link }}</a>
```
ColonelPhantom, line 41: seems like a pointless (and arguably wrong) indentation change

ColonelPhantom (changes requested):

website/partners/templates/partners/partner.html
BPC2004: very true but there was a merge conflict and thought i changed it so it was correct, but clearly not

PR #3960, opened 2025-08-13
Update django version to 5.2.5

T8902:
..


DeD1rk (approved):
Based on changelogs this should be alright

PR #3968, opened 2025-10-22
Add the position of photo to make picking an album cover easier

### LINKED ISSUE #3579: Simplify selecting a cover image
### What?
We should make it easy to select the cover image of an album.

### Why?
Since we changed filenames to be randomized, paparazcie has a harder time selecting a cover image. They used to look up the number and then select it, but that's no longer possible (they can now only look up the filename and search for it).

### How?
One option might be to somehow render the position of a photo in the cover image dropdown on the album admin, which would restore the previous behavior. But other options can also work, such as allowing the user to upload a file to use as cover image instead.

T8902:
Closes #3579 .

### Summary

the position of the picture will be saved with the random hex name.

This is a bit of a shitty solution, however it works @DeD1rk do you have any ideas on how to make this more clean?
### How to test

1. make fixtures

### DIFF
--- a/website/photos/models.py
+++ b/website/photos/models.py
@@ -27,7 +27,8 @@ logger = logging.getLogger(__name__)
 
 def photo_uploadto(instance, filename):
     ext = os.path.splitext(filename)[1]
-    return f"photos/{instance.album.dirname}/{token_hex(8)}{ext}"
+    position = str(instance).split("-")[0]
+    return f"photos/{instance.album.dirname}/{position}-{token_hex(8)}{ext}"
 
 
 class DuplicatePhotoException(Exception):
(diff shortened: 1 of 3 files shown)


website/photos/models.py
```diff
@@ -27,7 +27,8 @@
 
 def photo_uploadto(instance, filename):
     ext = os.path.splitext(filename)[1]
-    return f"photos/{instance.album.dirname}/{token_hex(8)}{ext}"
+    position = str(instance).split("-")[0]
```
DeD1rk: This is weird:

```
    def __str__(self):
        """Return the filename of a Photo object."""
        return os.path.basename(self.file.name)
```

That wouldn't have a - at all, and seems like a circular dependency to check the file name from uploadto

DeD1rk: Ah, I see, you're adding it in the filename in _try_save_photo Can you just get the index in this function directly instead of expecting the filename to have a specific format?

PR #3975, opened 2025-10-24
Update some dependencies

T8902:
I read through the update logs of these dependencies and they don't seem to have breaking changes. However, it's probably a good idea for someone to double check


Scarletto (approved):
jottem

T8902:
Waarom deed je niet mergen nu is conflicts :(

PR #4003, opened 2025-12-10
Add email to treasurer when late queue entrant should not get fine.

### LINKED ISSUE #3995: Entered event via queue
### What?

If someone enters an event that first waited in the queue, there should be some sort of checkbox showing that they entered the event via the queue and when they entered the event via the queue.
### Why?

This way it is easier to check if someone gets a fine or not for not showing up. Some people cancel too late. The first place in the queue joins the event, even though maybe forgot they were in the queue and do not show up, which prompts the treasurer to send a fine, which they don't necessarily deserve.
### How?

There is a remark box, maybe just add the letter q when someone joins via the queue.

ana-cullen:
Closes #3995 

### Summary
- Added emails to notify treasurer that a queue member can now attend the event but should not be fined if they don't attend because the registration deadline has passed.
- Added check to trigger this email to the cancel_registration service

### How to test
1. Create an event with registration and max participants
2. Add max participants+1
3. Cancel registration of 1 participant (via the user's own 'Cancel registration' button)
4. Wow check out that email!

### DIFF
--- a/website/events/templates/events/email/late_queue_entry_email.html
+++ b/website/events/templates/events/email/late_queue_entry_email.html
@@ -1,0 +1,12 @@
+{% extends "email/board_email.html" %}
+
+{% block content %}
+  Hi,<br>
+
+  <p>A member that was registered for the event '{{ event.title }}' has cancelled their registration after the deadline.</p>
+  <p>The first waiting queue member can now attend the event, but they should not receive a fine for non-attendance as the cancellation deadline has passed.</p>
+
+  <p>Name: {{ registration.member.profile.display_name }}</p>
+
+  <p>~ With love, from Techie</p>
+{% endblock %}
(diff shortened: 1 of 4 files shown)


website/events/templates/events/email/late_queue_entry_email.txt
```diff
@@ -0,0 +1,9 @@
+Hi,
```
T8902, line 1: Same here

website/events/templates/events/email/late_queue_entry_email.html
```diff
@@ -0,0 +1,12 @@
+{% extends "email/board_email.html" %}
+
+{% block content %}
+  Hi,<br>
```
T8902, line 4: This is to board i see, lets start it with dear board

T8902 (changes requested):

T8902:
Apologies for the initial comments, I confused myself

PR #4023, opened 2026-05-20
added items,payment and payer to api query

### LINKED ISSUE #4014: Extend `admin/shift/orders` api
### What?
Add payment info, payer, and price to the `/api/v2/admin/sales/shifts/{id}/orders/` endpoint.

### Why?
The original pizza API returns the the full order when requesting the orders via the admin API. Currently the API returns just the PK creation date and total number of items in `/api/v2/admin/sales/shifts/{id}/orders/`, which is not enough for the app.

The current pizza interface allows the user to filter/sort by certain things, like username or payment type. This is very useful when looking for members who haven't payed, or sorting by username. The new shift API limits this, as we do not have this information (without doing N+1 queries).

Currently this is kind of a blocker for releasing the shifts interface to the Thaliapp

### How?
No idea

Luukosaurus:
Closes #4014 .

### How to test

0. Make database entries for sales event.
1. Go to 'http://127.0.0.1:8000/api/v2/admin/sales/shifts/1/orders/'
2. See that there is now more useful data being fetched

### DIFF
--- a/website/sales/api/v2/admin/views.py
+++ b/website/sales/api/v2/admin/views.py
@@ -158,4 +158,5 @@ class OrderDetailView(AdminRetrieveAPIView, AdminUpdateAPIView, AdminDestroyAPIV
             "order_items", "order_items__product", "order_items__product__product"
         )
         queryset = queryset.prefetch_related("payment")
+        queryset = queryset.prefetch_related("payer")
         return queryset
--- a/website/sales/api/v2/admin/views.py
+++ b/website/sales/api/v2/admin/views.py
@@ -116,7 +116,7 @@ class OrderListView(AdminListAPIView, AdminCreateAPIView):
         queryset = queryset.prefetch_related(
             "order_items", "order_items__product", "order_items__product__product"
         )
-        queryset = queryset.prefetch_related("payment")
+        queryset = queryset.select_related("payment")
         return queryset
 
     def get_serializer_context(self):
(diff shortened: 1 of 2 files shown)


website/sales/api/v2/admin/views.py
```diff
@@ -158,4 +158,5 @@ def get_queryset(self):
             "order_items", "order_items__product", "order_items__product__product"
         )
         queryset = queryset.prefetch_related("payment")
+        queryset = queryset.prefetch_related("payer")
```
ColonelPhantom: I'm a bit rusty on the details of prefetch_related vs select_related, but if possible you should prefer using select_related by adding "payer" to its parameters (that does only 1 query instead of 2). Payer is a foreignkey so I think that should work.

ColonelPhantom (commented):
Did not test, but code looks good at a glance other than the prefetch thingy.

website/sales/api/v2/admin/views.py
ColonelPhantom: (While you're at it, the same may be possible for payment, which is OneToOne. Again, a bit rusty here.)

Luukosaurus: I changed it to this, but it's also a change on line 119. This is "old" code before my pull request, but it seems logical there as well. If you think this is wrong, just revert it.

JAicewizard:
I propose we merge this

ColonelPhantom (approved):

